PR #53, opened 2014-03-19
add cmake build system

auroraeosrose:
This is part one of making folly "windows friendly" - adding a cmake driven build system
because autotools on windows is pain

It works for the most part as is for the platforms cmake supports (which is a lot) - checks should be identical to those done by the automake system, a folly-config.h file is generated

checks are divided into compiler related, configure related, dependencies and options for the build

However this is not entirely complete

there are several cmake modules bundled including one that detects c++11 features (note that noexcept is missing from detection and should be added) - also the features absolutely required by folly should be added as well

What is yet todo - integrate gtest and building test binaries. dealing with python generated files, dealing with experimental files, doing the current libfolly partial libraries, and writing docs


sgolemon:
FYI - I'm going to use most of this, but it's going to be a big refactor as well.
I'll leave you with credit though, since I'm using much of this as inspiration.

PR #56, opened 2014-03-23
noexcept changes

auroraeosrose:
noexcept gets two macros, one for noexcept(value) and one for regular
noexcept - both are replaced with throw() which isn't an exact
replacement but should be "good enough"

### DIFF
--- a/folly/AtomicStruct.h
+++ b/folly/AtomicStruct.h
@@ -46,7 +46,7 @@ class AtomicStruct {
 
   Atom<Raw> data;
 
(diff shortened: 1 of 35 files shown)


facebook-github-bot-5:

Gownta:
Apologies for the delay in dealing with this pull request. I'm in the middle of going through all of our still-open pull requests.

The philosophy of folly is to be on the cutting edge, so we keep it relatively up-to-date with the current language spec. So unfortunately we can't accept this pull request.

PR #65, opened 2014-06-16
Make the STL functions in FunctionalExcept.cpp conditionally compiled

apurvam:
Make the STL functions in FunctionalExcept.cpp conditionally compiled 
based on the config.

Without this patch, programs which statically link libfolly and newer
versions of libstdc++ will cretae linker errors because of multiple
definitions of the __throw_* functions: one from libfolly and one from
libstdc++.

### DIFF
--- a/folly/detail/FunctionalExcept.cpp
+++ b/folly/detail/FunctionalExcept.cpp
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#ifndef FOLLY_HAVE_BITS_FUNCTEXCEPT_H
+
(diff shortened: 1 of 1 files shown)


facebook-github-bot-5:

Gownta:
Sorry for taking so long to get to this. The Makefile already conditionally adds FunctionalExcept.cpp for compilation; what process are you using that adds this to libfolly when the macro is false?

sgolemon:
Closing this as (already stated) this is being handled in `folly/Makefile.am` already (which is where it should be happening.  Also, that define is probably not the one you wanted anyway.

PR #81, opened 2014-08-24
Fix Traits.h std::* declarations conflict with libc++

mcuelenaere:
When compiling folly with clang and libc++, the declaration in folly/Traits.h
clash with the ones from the system library. Fix this by including the original
ones, instead of trying to do that ourselves.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -235,6 +235,8 @@ template <class T> struct IsZeroInitializable
  * although that is not guaranteed by the standard.
  */
 
+#ifndef _LIBCPP_VERSION
+
 FOLLY_NAMESPACE_STD_BEGIN
 
 template <class T, class U>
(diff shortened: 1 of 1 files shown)


facebook-github-bot-5:

Gownta:
Sorry for taking so long to get to this. I've been going through our github pull request over the last two weeks and dealing with all the languishing requests.

Thanks for making a patch to make folly better for different platforms. I agree that it would be appropriate to correct the definitions on libc++. However, I don't think that the appropriate solution is to include the full headers - that's what the forward declarations are trying to avoid in the first place. We probably need a fix that correctly forward-declares the types.

We're currently working on getting folly (and hhvm) continuously tested on OSX environments (and other platforms). As soon as we do, we'll be able to properly evaluate and maintain this patch. In the interim, we'll keep this patch listed on github so that it stays on our radar.

sgolemon:
So this appears to have been resolved by a fix applied to `Portability.h`  If you're still experiencing issues connected to it, please reopen this issue with additional information like: clang version, error outputs, etc...

PR #82, opened 2014-08-27
std::out_of_range underflow exception

huzhiguang:
std::out_of_range underflow exception is not easy to find the problem(I don't know  which of the statements when throw out_of_range ),so I add prefix can make better analysis.


facebook-github-bot-5:

Gownta:
For encapsulation, we prefer to keep implementation details out of error messages. This patch would best be applied by interested parties instead of being landed in trunk.

Thank you for your interest in making folly better!

PR #109, opened 2014-12-18
Add python binary check to build configuration

micahsaint:
#### Overview:
Currently when a user attempts to build folly without a python binary on their machine, the build will fail with a long and cryptic error message from the Makefile.

This commit will add the autoconf macro `m4/ax_python.m4` which will look for a python executable as well as the associated header files and libraries. If found, the macro will then set the binary path to `PYTHON_BIN`. That variable will then be used by the Makefile to execute the `build/generate*.py` files.

#### This commit rectifies the following use cases: 

**Problem**
The user's python binary is followed by the python version number (i.e., `python2.7`). This happens in Linux distributions such as Arch Linux and possibly others.
**Solution**
This patch will recurse through several python versions looking for a possible executable followed by its version number and then execute python scripts using the found executable.

**Problem**
The user does not have a python version on their machine and is confused as to why make fails, presenting a cryptic and verbose error message.
**Solution**
This patch will cause the configuration stage of the build to fail while subsequently presenting a clear error message explaining that python is required for the build to complete.

### DIFF
--- a/folly/Makefile.am
+++ b/folly/Makefile.am
@@ -250,15 +250,15 @@ nobase_follyinclude_HEADERS = \
 	wangle/detail/FSM.h
 
 FormatTables.cpp: build/generate_format_tables.py
-	build/generate_format_tables.py
+	$(PYTHON_BIN) build/generate_format_tables.py
 CLEANFILES += FormatTables.cpp
 
 EscapeTables.cpp: build/generate_escape_tables.py
-	build/generate_escape_tables.py
+	$(PYTHON_BIN) build/generate_escape_tables.py
 CLEANFILES += EscapeTables.cpp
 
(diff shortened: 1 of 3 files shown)


facebook-github-bot-9:

Gownta:
I don't think we can accept this patch - the license inside the code is GPL v2, and the special exemption only exempts files "that are the output of Autoconf", which the actual autoconf script itself is not.

sgolemon:
As @Gownta said, we can't merge this diff due to it being GPL, but since autotools comes with AM_PATH_PYTHON, we should be able to get the same effect with a much smaller patch.

I'll put up a new diff using the builtin macro and close this once that's merged.

sgolemon:
Internal diff: D2147977

PR #114, opened 2014-12-30
Include type_traits in ForEach.h

nsuke:
Without this, `make check` doesn't compile for me.

Maybe for some boost versions, `type_traits` is includeded from the `remove_cv.hpp`.

I use boost 1.57.0 and gcc 4.9.2


mpwl:
the fix has been fixed in fe41434ed66e679162423eb608df857d844013b8. Thanks for taking time to submit the bug and the fix!

PR #125, opened 2015-02-04
fix configure script to check pthread_atfork(3) in the correct way

0mok:
In Ubuntu 14.04, configure script can't find `pthread_atfork`.

`AC_CHECK_FUNCS()` doesn't include `pthread.h`. Therefore it can't find the function.


marcinpe:
I verified the fix on latest arch linux with gcc 4.9.2. Big thanks for help!

PR #127, opened 2015-02-06
Initial patch to build Folly in PPC64 machines

lbianc:
This is an initial patch to build Folly in PPC64 machines. With these changes it's possible to start porting HHVM to PPC64 as well.
Reviews are welcome.


lbianc:
@fugalh @djwatson @jsedgwick 

Could you help reviewing this PR?
Thanks!!

lbrandy:
I've raised this issue internally, but my initial thoughts are that I'm not sure we can accept this pull request because it's very unlikely we can and would be willing to maintain or support this going forward. Realistically, we're going to keep breaking this support. 

The only way something like this would work is if you, or someone like you, is willing to put in the time to keep it up to date. I think that would be better as a separate branch that you maintain. If you are long-term committed to keeping such a branch maintained, I'd even add a link to your branch in our docs to help others find it.

Gownta:
It is my understanding, from reading the comments on this patch, that this pull request should now be closed. There is apparently a release tagged with this revision.

We're working on getting folly continuously tested on different platforms. As soon as we do, we'll be able to properly maintain other platforms, which should make everyone's life easier.

fabioperez:
@Gownta 

Sorry, but I couldn't find the release that was tagged with this revision. Isn't https://github.com/facebook/folly/pull/127#commitcomment-10027494 referring to https://github.com/facebook/folly/commit/75278c6326af4e6a81b8761acb5e24f6442dfa8f and not to this pull request?

PR #130, opened 2015-02-14
EmulatedFutexAtomic doesn't copy or move

cppfool:
EmulatedFutexAtomic concisely says here that it doesn't copy construct,
move construct, copy assign, or move assign.


lbrandy:
I've committed your patch to our internal repository. Next public push should auto-close this PR.

PR #139, opened 2015-02-27
Add const to the variable: needle

loverszhaokai:
Make it clear that the variable is const


lbrandy:
I've committed this internally and will go out in the next public sync (which I'll hopefully do later today). This should autoclose then.

loverszhaokai:
So, Should I close this pull ?

PR #140, opened 2015-02-27
Update the comment of Benchmark.h

loverszhaokai:
Remove one of the "one" in the comment of BENCHMARK


lbrandy:
I've committed this internally and will go out in the next public sync (which I'll hopefully do later today). This should autoclose then.

loverszhaokai:
So, Should I close this pull ?

PR #144, opened 2015-03-03
Additional header check for bits/funcexcept.h

zauguin:
Added an ad-hoc existance check for bits/functexcept.h on clang. This is required if folly is configured for the wrong standard library implementation.

### DIFF
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -199,9 +199,11 @@ struct MaxAlign { char c; } __attribute__((__aligned__));
 
 #if defined(__has_include)
 #if __has_include(<bits/functexcept.h>)
-#define FOLLY_HAVE_BITS_FUNCTEXCEPT_H
+#undef FOLLY_HAVE_BITS_FUNCTEXCEPT_H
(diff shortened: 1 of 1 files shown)


Gownta:
This feels like we're making code to fix a symptom of the problem instead of the source. Isn't the correct solution to configure folly correctly?

If folly is configured incorrectly, then this will not be the only source of error; we would actually prefer to fail hard and fail fast instead of silently dealing with the issue.

zauguin:
Actually you're right, the nice way would be to configure folly for the right library, but IB the Gase of multiple standard library implementations there is no "right" one. So you need one folly Installation for every library. It's not a huge problem, but for most header only usage, it feels unnecessarry. I agree that this may break in other circumstances, but than IMHO there should be a Static_assert or similar. That would fail hard and fail fast and give good diagnostics.

Gownta:
I think this patch would be best applied by people who have multiple libraries, and not in folly trunk. The use of __has_include isn't ready for gcc < 5.0, and folly is still supposed to work on 4.8.

PR #149, opened 2015-03-13
AcceptCallback disallow undeclared copy and move, but allow default ctor

cppfool:
AcceptCallback disallow undeclared copy construction, copy assignment,

move construction, move assignment. But it allow implicit default

construction.

Test Plan: all folly/tests, make check for 37 tests, passed.

### DIFF
--- a/folly/io/async/AsyncServerSocket.h
+++ b/folly/io/async/AsyncServerSocket.h
@@ -62,6 +62,9 @@ class AsyncServerSocket : public DelayedDestruction {
 
   class AcceptCallback {
    public:
+    // AcceptCallback disallow undeclared copy and move.
+    // But allow implicit default construction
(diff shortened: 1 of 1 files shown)


chipturner:
Hello, thank you for your patch (and the others, which I am in the process of merging).  Can you offer a little more context on this diff?  Why only partially disabling value semantics?

cppfool:
@chipturner, Classes which inherits from AcceptCallback need
default constructor but others.
At least not yet.

cppfool:
@chipturner, We're =deleteing move assign here. But we're not explicitely =deleteing other
value semantics enablers. By asking "only partially disabling value semantics" are
you targeting previous two sentences?

Gownta:
Unless it offers a concrete benefit, we should not be marking functions as deleted.

PR #152, opened 2015-03-29
SmoothLoopTime has a "constexpr constructor" and a "constexpr get" member

cppfool:
The constructor "explicit SmoothLoopTime::SmoothLoopTime(uint64_t);" 

may construct during compilation, and same with

"double SmoothLoopTime::get() const;" member.

Test Plan:

all folly/tests, make check for 37 tests, passed.

### DIFF
--- a/folly/io/async/EventBase.h
+++ b/folly/io/async/EventBase.h
@@ -485,10 +485,10 @@ class EventBase : private boost::noncopyable,
 
   class SmoothLoopTime {
    public:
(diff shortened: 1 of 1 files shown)


cppfool:
This pull request has been closed in favour

of pull request #180 and #181.

This commit compiles well with "clang++ -std=c++11",

but fails with "g++ -std=c++11". The reason for this

failure is:

"clang++ -std=c++11" allows c++14-extended constexpr

while g++ doesn't do so. The pull request #180 and #181

address this failure.

PR #153, opened 2015-03-29
EventBaseObserver disallows copy, move, and default ctor

cppfool:
EventBaseObserver disallows copy construction, copy assignment,

construction.

for ~EventBaseObserver.

Test Plan:


Gownta:
We don't want to delete the move constructor - it unnecessarily restricts the inheritance contract of EventBaseObserver.

PR #154, opened 2015-03-29
RequestEventBase disallows copy, move, and default ctor

cppfool:
RequestEventBase disallows copy construction, copy assignment,

construction.

Test Plan:


Gownta:
We don't want to delete the move constructor for this class.

PR #155, opened 2015-03-29
EventBase disallows copy, move, and default ctor

cppfool:
EventBase doesn't inherit from boost::noncopyable

anymore, thus, maybe, a little more readable.

EventBase disallows copy construction, copy assignment,

construction.

We may compute kDEFAULT_IDLE_WAIT_USEC during compilation.

Test Plan:


Gownta:
We're going to keep boost::noncopyable for the while.

PR #156, opened 2015-03-29
EventBase::EventBase(event_base*, bool) explicitly states narrowing

cppfool:
EventBase::EventBase(event_base*, bool) initializer list uses

"initialization syntax" and thus explicitly states where it narrows,

e.g., in "nextLoopCnt_{static_cast<uint64_t>(-40)}" .

Test Plan:


Gownta:
This is mostly a stylistic change - it doesn't fix anything.

PR #157, opened 2015-03-29
LoopCallback disallows copy and move, but allows default ctor

cppfool:
EventBase::LoopCallback disallows copy construction, copy assignment,

move construction, and move assignment. But it allows default

construction for classes which inherit from it, to construct

base subobject for derived classes.

for EventBase::~LoopCallback.

Test Plan:

### DIFF
--- a/folly/io/async/EventBase.h
+++ b/folly/io/async/EventBase.h
@@ -113,7 +113,10 @@ class EventBase : private boost::noncopyable,
    */
   class LoopCallback {
    public:
-    virtual ~LoopCallback() {}
+    // It disallows copy and move, but allows default ctor
+    LoopCallback& operator=(LoopCallback&&) = delete;
+    
+    virtual ~LoopCallback() = default;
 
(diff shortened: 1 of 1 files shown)


Gownta:
We don't want to delete this class' move constructor.

@cppfool: if you'd like to make a diff that changes ALL of folly's default-implemented (i.e. {}) functions into =default, that would be an impactful diff.

folly/io/async/EventBase.h
```diff
@@ -113,7 +113,10 @@ class EventBase : private boost::noncopyable,
    */
   class LoopCallback {
    public:
-    virtual ~LoopCallback() {}
+    // It disallows copy and move, but allows default ctor
+    LoopCallback& operator=(LoopCallback&&) = delete;
+    
+    virtual ~LoopCallback() = default;
```
Gownta: If you change ALL such functions in folly, that would be an impactful change.

PR #160, opened 2015-03-29
folly::kNoFD can be constexpr

cppfool:
We may compute folly::kNoFD during compilation.

Test Plan:


Gownta:
It's in a .cpp, so the constexpr doesn't empower other people. If we need to make it constexpr as part of another change, we'll do it there.

PR #161, opened 2015-03-29
FunctionLoopCallback disallows copy, move, and default ctor

cppfool:
FunctionLoopCallback disallows copy construction,

construction.

Test Plan:


Gownta:
There isn't a need to delete the move constructor.

PR #162, opened 2015-03-29
FunctionRunner disallows copy and move, but allows default ctor

cppfool:
EventBase::FunctionRunner disallows copy construction,

copy assignment, move construction, and move assignment.

But it allows default construction for classes which inherit

from it, to construct base subobject for derived classes.

Test Plan:


Gownta:
EventBase::FunctionRunner is in the cpp - it is unreferenceable from external code. Further, it has no constructors or implictly-defined methods, since it inherits from NotificationQueue. I do not think that this is appropriate here.

PR #163, opened 2015-03-29
EventBase::EventBase(bool) explicitly states narrowing

cppfool:
EventBase::EventBase(bool) initializer list uses "initialization syntax"

and thus explicitly states where it narrows, e.g.,

in "nextLoopCnt_{static_cast<uint64_t>(-40)}" .

Test Plan:


Gownta:
What does this patch accomplish?

Gownta:
Per an email with @cppfool, this is a stylistic change, and is not immediately impactful.

PR #164, opened 2015-03-31
DelayedDestruction disallows copy, move, and default ctor

cppfool:
DelayedDestruction disallows copy construction,

construction.

DelayedDestruction needn't inherit from boost::noncopyable

anymore, thus, maybe, it's little more readable and simple.

Test Plan:

### DIFF
--- a/folly/io/async/DelayedDestruction.h
+++ b/folly/io/async/DelayedDestruction.h
@@ -16,7 +16,6 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


Gownta:
I think it would be best to leave the noncopyable as is. Its intent is unambiguous. Having =delete statements in the code, while easier to grep for, suggests that individual pieces can be enabled.

PR #165, opened 2015-03-31
DelayedDestruction() may get computed during compilation

cppfool:
We may default construct DelayedDestruction during compilation,

by marking default constructor constexpr.

for ~DelayedDestruction().


Gownta:
We need to wait a bit on this one, to give people the opportunity to upgrade boost (since, e.g., boost 1.51 doesn't have noexcept constructors). Boost 1.57 has defaulted (and noexcepted) constructors.

PR #166, opened 2015-03-31
DelayedDestruction() can go noexcept

cppfool:
Initializations, guardCount_{0} and destroyPending_{false},

are trivial, and far from throwing anything.

For esoteric cases, we might not like to handle any exception(s)

out of DelayedDestruction().


Gownta:
What is the practical rationale behind this diff? Is there a concrete case where we need noexcept constructors? The general consensus on noexcept seems to be that they are useful for move-type operations, and specific use-cases.

Gownta:
Per email conversation with @cppfool, this is a stylistic change, and not an immediately-impactful change.

PR #167, opened 2015-03-31
DestructorGuard disallows copy, move, and default ctor

cppfool:
DestructorGuard disallows copy construction,

construction.

Test Plan:


Gownta:
Since we have a copy constructor, move construction will call the copy constructor, which results in the appropriate behavior. Disabling move construction would unduly restrict this class' API.

PR #168, opened 2015-03-31
DestructorGuard(DelayedDestruction*) and copy ctor can go constexpr

cppfool:
We may construct DestructorGuard(DelayedDestruction*) and

DestructorGuard(const DestructorGuard&) during compilation,

by marking them constexpr.


cppfool:
This pull request has been closed in favour

of pull request #179.

This commit compiles well with "clang++ -std=c++11",

but fails with "g++ -std=c++11". The reason for this

failure is:

"clang++ -std=c++11" allows c++14-extended constexpr

while g++ doesn't do so. The pull request #179

address this failure.

PR #169, opened 2015-03-31
DestructorGuard(DelayedDestruction*) and copy ctor can go noexcept

cppfool:
Constructions, DestructorGuard(DelayedDestruction*) and

DestructorGuard(const DestructorGuard&), are trivial.

We might not like to handle any exception(s)

out of the constructions.


Gownta:
There isn't a general advantage to add noexcept to non-move/non-destruction functions.

PR #171, opened 2015-04-05
void dampen(double) can be computed during compilation

cppfool:
If we're compiling for c++14, void is a literal in c++14,

void SmoothLoopTime::dampen(double) can be computed during

compilation.

Test Plan:


Gownta:
We try to keep folly on one version, without a bunch of ifdefs. When we formally upgrade folly to c++14, we can add the constexpr then. Until that time, this type of code becomes a maintenance burden.

PR #172, opened 2015-04-05
ThreadCachedInt(IntT, uint32_t) can be marked constexpr and noexcept

cppfool:
ThreadCachedInt(IntT, uint32_t) can be marked constexpr and noexcept.

Because initializations target_{initialVal} and cacheSize_{cacheSize}

are constexpr and noexcept, ThreadCachedInt(IntT, uint32_t) doesn't have to do

anything else. Therefore, there seems no reason why

ThreadCachedInt(IntT, uint32_t) shouldn't be constexpr and noexcept.

Constructions std::atomic<IntT>(IntT) and std::atomic<uint32_t>(uint32_t)

are constexpr and noexcept.

Test Plan:

### DIFF
--- a/folly/ThreadCachedInt.h
+++ b/folly/ThreadCachedInt.h
@@ -42,8 +42,9 @@ class ThreadCachedInt : boost::noncopyable {
   struct IntCache;
 
  public:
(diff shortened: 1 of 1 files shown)


Gownta:
If we someday need a constexpr ThreadCachedInt, we can add this then. If not, then this adds a restriction to the API for future development.

PR #173, opened 2015-04-05
ThreadCachedInt disallows copy, move, and default ctor

cppfool:
ThreadCachedInt disallows copy construction,

construction.

ThreadCachedInt needn't inherit from boost::noncopyable

anymore, thus, maybe, it's little more readable and simple.

Test Plan:


Gownta:
We're going to keep boost::noncopyable for the while.

PR #174, opened 2015-04-06
Funtions can be noexcept

cppfool:
The following funtions don't call anything which isn't noexcept.

So we can safely mark them noexcept, which compiler might like.

IntT ThreadCachedInt::readFast() const;

IntT ThreadCachedInt::readFastAndReset();

void ThreadCachedInt::setCacheSize(uint32_t);

uint32_t ThreadCachedInt::getCacheSize() const; 

Test Plan:


Gownta:
noexcept isn't all that useful except for destructors and move operations.

PR #175, opened 2015-04-06
Funtions can be considered to be noexcept

cppfool:
Almost all calls inside these functions are noexcept.

So we can safely mark them noexcept, which compiler might like.

Consider the function

void ThreadCachedInt::IntCache::increment(IntT inc);.

We might not like to handle unlikely exception out of 

one trivial integral addition and one integer increment.

Other than that, it doesn't call anything which isn't noexcept.

Consider the function

void ThreadCachedInt::IntCache::flush() const;.

We might not like to handle unlikely exception out of 

one trivial integer assignment.

Other than that, it doesn't call anything which isn't noexcept.

Test Plan:


Gownta:
Adding noexcept to non-move/destruction functions isn't impactful.

PR #176, opened 2015-04-06

cppfool:

time computation.

IntT ThreadCachedInt::readFast() const;

IntT ThreadCachedInt::readFastAndReset();

void ThreadCachedInt::setCacheSize(uint32_t);

uint32_t ThreadCachedInt::getCacheSize() const;

void ThreadCachedInt::IntCache::flush() const;

Test Plan:


Gownta:
When we move to c++14, we can consider adding these constexprs. Until then, the ifdefs are a maintenance burden.

PR #177, opened 2015-04-06
IntCache disallows copy, move, and default ctor

cppfool:
ThreadCachedInt::IntCache disallows copy construction,

construction.

Test Plan:


Gownta:
This is a private class - there is no benefit to deleting the move constructor.

PR #178, opened 2015-04-06
IntCache(ThreadCachedInt&) can be marked constexpr and noexcept

cppfool:
ThreadCachedInt::IntCache::IntCache(ThreadCachedInt&) can be marked

constexpr and noexcept. Because initializations val_{0} and reset_{false}

are constexpr and noexcept. And initializations parent_{&parent} and

numUpdates_{0} are trivial.

Test Plan:


Gownta:
If there becomes a need to make these constructors constexpr, then we'll add them then.

PR #179, opened 2015-04-08
DestructorGuard(DelayedDestruction*) and copy ctor can go constexpr

cppfool:
We may construct DestructorGuard(DelayedDestruction*) and

DestructorGuard(const DestructorGuard&) during compilation,

by marking them constexpr, for c++14.


Gownta:
We're going to avoid all the ifdefs in folly. When we upgrade to c++14, we can put the code in without the preprocessor guards.

PR #180, opened 2015-04-08
SmoothLoopTime has a "constexpr constructor"

cppfool:
The constructor "explicit SmoothLoopTime::SmoothLoopTime(uint64_t);" 

may construct during compilation, for c++14.

Test Plan:

all folly/tests, make check for 37 tests, passed.


Gownta:
We're going to leave out the ifdefs, and just move to c++14 when the time is right.

PR #181, opened 2015-04-08
SmoothLoopTime has a "constexpr get" member

cppfool:
The member function "double SmoothLoopTime::get() const;" 

may compute during compilation.

Test Plan:


Gownta:
We're going to leave constexpr off of functions unless it is necessary.

PR #184, opened 2015-04-09
We might avoid some temporaries in putMessagesImpl

cppfool:
It seems we might avoid some temporaries in putMessagesImpl

function. To do so we pass arguments directly to constructors

with the help of emplace_back member.

Test Plan:

### DIFF
--- a/folly/io/async/NotificationQueue.h
+++ b/folly/io/async/NotificationQueue.h
@@ -514,9 +514,7 @@ class NotificationQueue {
       if (numActiveConsumers_ < numConsumers_) {
         signal = true;
       }
-      queue_.push_back(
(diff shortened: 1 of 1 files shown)


Gownta:
@cppfool: I like this diff! I'll put it up for review internally.

Gownta:
I've landed this internally - next time folly is synced, this will appear on github.

Gownta:
@cppfool, our internal landing system clobbered your name in our git history and replaced it with mine. :/ I've commented on the commit, attributing it to you, but I'd also like to cross-reference it here:

https://github.com/facebook/folly/commit/0011d8a0631c54b05cd40f35f47b181109165e52

PR #186, opened 2015-04-09
"Consumer();" might be constexpr and noexcept

cppfool:
We shouldn't come in the way of compiler when it can

decide whether Consumer(); is constexpr and noexcept.

We don't declare and define Consumer();, but we ask for

that from compiler using "Consumer() = default" and

brace-initialize the data members appropriately.

Test Plan:


Gownta:
@cppfool: this diff is interesting, but I think we'll maintain the old style of constructors. If a change like this is to be pulled, we'd need to do it to all of folly.

PR #187, opened 2015-04-09
ARM64 assembler fixes for Folly.

ajasty-cavium:
Requesting a tracking branch. These fixes are required to compile and run hhvm on aarch64.

### DIFF
--- a/folly/MPMCQueue.h
+++ b/folly/MPMCQueue.h
@@ -646,7 +646,7 @@ struct TurnSequencer {
       // the first effectSpinCutoff tries are spins, after that we will
       // record ourself as a waiter and block with futexWait
       if (tries < effectiveSpinCutoff) {
-        asm volatile ("pause");
+	PAUSE();
         continue;
       }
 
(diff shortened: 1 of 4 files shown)


paulbiss:
cc @edwinsmith

edwinsmith:
and cc @jdelong in case of folly questions. We will need the CLA before this can be committed, but a few initial comments:

* (minor) there's one place where PAUSE needs to be indented one more space
* PAUSE should probably be called FOLLY_PAUSE since macros aren't in any namespace.  I'm not very familliar with folly coding conventions, @jdelong - should it be an inline function instead? and is there a better place than Portability.h?
* is FOLLY_X64 intended to be x86-64 specific or just 64-bit specific? i'm assuming it implies arch and bits, so I think this diff should also introduce FOLLY_A64, and if there's code that's really using FOLLY_X64 as a 64-vs-32 bit flag, we might just want a separate FOLLY_64 flag for that. (maybe separate diff).

sgolemon:
Combined with #190

PR #188, opened 2015-04-09

cppfool:

time computation, and also become noexcept.

NotificationQueue* NotificationQueue::Consumer::getCurrentQueue() const;

void NotificationQueue::Consumer::setMaxReadAtOnce(uint32_t);

uint32_t NotificationQueue::Consumer::getMaxReadAtOnce() const;

EventBase* NotificationQueue::Consumer::getEventBase();

Test Plan:


Gownta:
We'll wait for c++14 before adding this code, without the preprocessor guards.

PR #189, opened 2015-04-09

cppfool:

time computation.

void NotificationQueue::setMaxQueueSize(uint32_t);

bool NotificationQueue::checkQueueSize(size_t, bool) const;

bool NotificationQueue::checkDraining(bool);

bool NotificationQueue::tryConsumeEvent();

And the following function can also be noexcept.

void NotificationQueue::setMaxQueueSize(uint32_t);

Test Plan:


Gownta:
We'll wait for c++14 before adding this code, without the preprocessor guards.

PR #190, opened 2015-04-10
Arm64yk

ykarlsbrun:
Cosmetic changes for ajusty-cavium pull request
Changed test HashTest.cpp: for arm64 signed char should be mentioned explicitly
Test CpuId for arm64 is stub for now

### DIFF
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -273,4 +273,10 @@ inline size_t malloc_usable_size(void* ptr) {
 }
 #endif
 
+#if defined(__x86_64__)
+#define PAUSE() asm volatile ("pause")
+#elif defined(__aarch64__)
+#define PAUSE() asm volatile ("wfe")
+#endif
+
 #endif // FOLLY_PORTABILITY_H_
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -274,9 +274,9 @@ inline size_t malloc_usable_size(void* ptr) {
 #endif
 
 #if defined(__x86_64__)
-#define PAUSE() asm volatile ("pause")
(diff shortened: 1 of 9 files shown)


paulbiss:
cc @edwinsmith

Gownta:
Thank you very much for your effort to make folly available on more platforms! Sorry for taking so long to comment on this patch. I've been going through the github pulls over the last few weeks and commenting on languishing requests.

We're working on getting folly continuously tested on different platforms. As soon as we do, we'll be able to properly evaluate and maintain this patch. In the interim, we'll keep this patch listed on github so that it stays on our radar

paulbiss:
I'm assuming you meant to leave this open.

Gownta:
Yes! Good catch, thank you. I must have pressed 'close and comment' instead of 'comment'

folly/test/CpuIdTest.cpp
```diff
@@ -24,5 +24,9 @@ using namespace folly;
 TEST(CpuId, Simple) {
   // All CPUs should support MMX
   CpuId id;
+#ifdef __aarm64__
```
Orvid: This doesn't quite seem right. Shouldn't it be the other way around?

folly/Portability.h
```diff
@@ -273,4 +273,10 @@ inline size_t malloc_usable_size(void* ptr) {
 }
 #endif
 
+#if defined(__x86_64__)
```
Orvid: This should be `#if defined(FOLLY_X64)`, which is what normally wraps the pause calls.

sgolemon:
Rebased and applied the notes so far while pulling: D2152868

PR #193, opened 2015-04-22
We might accidentally move an lvalue

cppfool:
Do we really need to move them?

We might accidentally move an lvalue handler.

Which we can avoid by forwarding it rather than moving it.

Test Plan:


Gownta:
@cppfool: instead of many small diffs changing moves to forwards (where appropriate), could you please make one diff that fixes this problem in all of folly.

PR #194, opened 2015-04-22
avoid temporary

cppfool:

Test Plan:


Gownta:
@cppfool: instead of many small diffs that change inserts-of-temporaries to emplaces, could you please make one diff that fixes this in all of folly. It will be more impactful, and significantly easier to test (since we'll only need to do one internal review and test pass instead of many).

PR #195, opened 2015-04-22
We might avoid temporary

cppfool:

We need constructor which moves folly::SocketAddress

The constructor NetworkAddress(const folly::SocketAddress&&, unsigned)

is a help when we move folly::SocketAddress.

For example when we pass folly::SocketAddress using

emplace or emplace_back members of containers.

Test Plan:


Gownta:
@cppfool: see comment on https://github.com/facebook/folly/pull/194

PR #196, opened 2015-04-22
temporary is not needed

cppfool:

Test Plan:


Gownta:
@cppfool: see comment on #194

PR #200, opened 2015-05-02
Some fixes to build on Mac OS X

andycheg:
When i tried to build hhvm on Mac OS X, build of folly failed with some mistakes. These fixes help me to build folly.

### DIFF
--- a/folly/Singleton.cpp
+++ b/folly/Singleton.cpp
@@ -44,8 +44,12 @@ struct FatalHelper {
 
   std::vector<detail::TypeDescriptor> leakedSingletons_;
 };
-
+#ifdef __APPLE__
+// OS X doesn't support __init_priority__. Just pray it works, I guess.
+FatalHelper fatalHelper;
+#else
 FatalHelper __attribute__ ((__init_priority__ (101))) fatalHelper;
+#endif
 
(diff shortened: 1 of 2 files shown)


Gownta:
Thank you for your interest in folly!

It is important that folly's singleton class be initialized at the specified priority (101 means "first!"), since singletons have peculiar behavior during static initialization and destruction (your "Just pray it works" comment is more apt than first appears!).

We're working on getting folly continuously tested on different platforms, including OSX. Until that time, I think that this patch would be best kept on your local branch. Alternatively, if you know of a way to force the fatalHelper to be initialized first on OSX, that would be a very appropriate fix.

Gownta:
Update: we have someone looking at this internally.

andycheg:
I've made this fix looking at similar one:

https://github.com/facebook/folly/commit/be419a8edc10bc8a6b21bac0efc7ac9a0d88107b

I guess this link may be usefull.

jwatzman:
I think my diffs (which haven't been pushed out to github yet) fix both of these issues. Sorry I didn't see this PR since it was closed at the time :(

andycheg:
Issues fixed by other commits.

PR #201, opened 2015-05-05
Keep Makefile.am from double-installing detail/Clock.h

fish2000:
On non-Linux systems, the autotools-based build system conditionally appends `detail/Clock.h` to `libfollybase_la_SOURCES` – even though that file is in there statically no matter what (as per [line 46](https://github.com/facebook/folly/blob/master/folly/Makefile.am#L46)), which causes the `install` command to error out trying to install the file twice, during `make install`… this patch takes out the conditional append.

### DIFF
--- a/folly/Makefile.am
+++ b/folly/Makefile.am
@@ -423,7 +423,6 @@ libfolly_la_SOURCES += \
 endif
 
 if !HAVE_LINUX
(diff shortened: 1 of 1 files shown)


sgolemon:
Nice catch.

Pulled for review: D2148004

PR #206, opened 2015-05-14
Fix for clang

191919:
Roughly described in https://github.com/facebook/hhvm/issues/5344.

And this is for compilation fix, so without  https://github.com/191919/mcrouter/commit/6d6731456894e5ee288bbdf146108f9f45b3750e, clang could compile mcrouter.

### DIFF
--- a/folly/DynamicConverter.h
+++ b/folly/DynamicConverter.h
@@ -55,7 +55,7 @@ BOOST_MPL_HAS_XXX_TRAIT_DEF(iterator);
 BOOST_MPL_HAS_XXX_TRAIT_DEF(mapped_type);
 
 template <typename T> struct class_is_container {
-  typedef std::reverse_iterator<T*> some_iterator;
+  typedef std::reverse_iterator<typename T::iterator> some_iterator;
   enum { value = has_value_type<T>::value &&
                  has_iterator<T>::value &&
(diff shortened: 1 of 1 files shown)


folly/DynamicConverter.h
```diff
@@ -55,7 +55,7 @@ BOOST_MPL_HAS_XXX_TRAIT_DEF(iterator);
 BOOST_MPL_HAS_XXX_TRAIT_DEF(mapped_type);
 
 template <typename T> struct class_is_container {
-  typedef std::reverse_iterator<T*> some_iterator;
+  typedef std::reverse_iterator<typename T::iterator> some_iterator;
```
sgolemon: This broke folly builds on our internal CI, but I think I came up with something suitable for both clang and us

```
template <typename T> struct iterator_class_is_container {
  typedef std::reverse_iterator<typename T::iterator> some_iterator;
  enum { value = has_value_type<T>::value &&
              std::is_constructible<T, some_iterator, some_iterator>::value };
};

template <typename T> struct class_is_container
  : std::conditional<
      has_iterator<T>::value,
      iterator_class_is_container<T>,
      std::false_type
    >::type {};
```

A diff containing this is currently up for review as D2170803

PR #207, opened 2015-05-14
Add EOS checking to 'parseLeadingNumber' function.

andrianyablonskyy:
parseLeadingNumber does not contain EOS validation and raise an exception when '/sys/devices/system/cpu/cpu0/index0/shared_cpu_list' contains only one number like '0'.


sgolemon:
Pulled for review: D2148015

Issue #205, opened 2015-05-14
Build failure with Boost 1.58 on Linux


fifilyu:
when I try to use Boost 1.58, I get the following error:
```
$ make check

Making check in .
make[1]: Entering directory '/home/fifilyu/workspace/folly/folly'
make[1]: Leaving directory '/home/fifilyu/workspace/folly/folly'
Making check in test
make[1]: Entering directory '/home/fifilyu/workspace/folly/folly/test'
Making check in .
make[2]: Entering directory '/home/fifilyu/workspace/folly/folly/test'
make  libgtestmain.la libgtest.la benchmark_test concurrent_skiplist_benchmark sorted_vector_types_test foreach_test hash_test timeout_queue_test conv_test range_test bits_test bit_iterator_test spin_lock_test small_locks_test packed_sync_ptr_test small_vector_test discriminated_ptr_test cpuid_test fbstring_test_using_jemalloc thread_cached_int_test thread_local_test fbvector_test dynamic_test json_test scope_guard_test endian_test rw_spinlock_test synchronized_test concurrent_skiplist_test histogram_test group_varint_test map_util_test string_test producer_consumer_queue_test atomic_hash_array_test atomic_hash_map_test format_test fingerprint_test portability_test spooky_hash_v1_test spooky_hash_v2_test
make[3]: Entering directory '/home/fifilyu/workspace/folly/folly/test'
make[3]: 'libgtestmain.la' is up to date.
make[3]: 'libgtest.la' is up to date.
/bin/sh ../libtool  --tag=CXX   --mode=link g++  -std=gnu++0x -g -O2 -lboost_thread -lboost_filesystem -lboost_system -lboost_regex -lpthread  -o benchmark_test BenchmarkTest.o libgtestmain.la ../libfollybenchmark.la ../libfolly.la -llzma -lz -lsnappy -llz4 -liberty -ljemalloc -levent -ldouble-conversion -lssl -lgflags -lglog 
[...]

PR #210, opened 2015-05-25
fix for Boost >= 1.56

fifilyu:
Fix for [issues #205](http://example.net/)
Folly still continue to build and work on earlier boost versions(boost < 1.56).

**boost::context::make_fcontext() change history:**
* __1.55:__    fcontext_t * make_fcontext( void * sp, std::size_t size, void(* fn)(intptr_t) );
* __1.56:__    fcontext_t make_fcontext(void* sp,std::size_t size,void(*fn)(intptr_t));

**boost doc:**
http://www.boost.org/doc/libs/1_55_0/libs/context/doc/html/context/context/boost_fcontext.html
http://www.boost.org/doc/libs/1_56_0/libs/context/doc/html/context/context/boost_fcontext.html


sgolemon:
Pulled for review: D2148069

PR #211, opened 2015-05-27
Fix capitalization error in README.md

grewal:


lbrandy:
That particular capitalization is intentional, since it's ostensibly where the 'y' in 'folly' comes from.

sgolemon:
Pulled for review: D2144702

Issue #213, opened 2015-05-27
Variadic Macros seem to be borked


xnox:
I have applied below, to make folly aware of the dual c++ abi... as otherwise there is std::string 98 and 11 variants, whilst everything defaults to 11 variant... Pretty sure this is not the right way to patch this, but it works for those on the gcc5 & 11 ABI by default.

```
diff --git a/folly/FBString.h b/folly/FBString.h
index e492e8f..896a4dd 100644
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -909,7 +909,7 @@ public:
   }
 
 private:
-  std::basic_string<Char> backend_;
+  std::__cxx11::basic_string<Char> backend_;
 };
 #endif // !_LIBSTDCXX_FBSTRING
 
diff --git a/folly/Traits.h b/folly/Traits.h
index 309157c..acc9cab 100644
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -19,6 +19,9 @@
 #ifndef FOLLY_BASE_TRAITS_H_
 #define FOLLY_BASE_TRAITS_H_
 
+#include <list>
+#include <string>
+
 #include <memory>
 #include <limits>
 #include <type_traits>
```

After doing that I'm still getting preprocessor / gcc5 errors about variadic macros being wrong... I'm not quite sure how to resolve that =( Any thoughts as to why __VA_ARGS__ are getting rejected here?!

```
In file included from ./../folly/FBString.h:59:0,
                 from ./../folly/Conv.h:27,
                 from detail/CacheLocality.cpp:23:
./../folly/Traits.h:155:38: error: template argument 1 is invalid
   struct IsRelocatable<  __VA_ARGS__ > : std::true_type {};
                                      ^
./../folly/Traits.h:221:3: note: in expansion of macro 'FOLLY_ASSUME_RELOCATABLE'
   FOLLY_ASSUME_RELOCATABLE(__VA_ARGS__<T1, T2, T3>) }                   \
   ^
./../folly/Traits.h:427:1: note: in expansion of macro 'FOLLY_ASSUME_FBVECTOR_COMPATIBLE_3'
 FOLLY_ASSUME_FBVECTOR_COMPATIBLE_3(std::basic_string);
 ^
./../folly/Traits.h:171:48: error: template argument 1 is invalid
   struct has_nothrow_constructor<  __VA_ARGS__ > : ::boost::true_type {};
                                                ^
[...]

Issue #208, opened 2015-05-19
Folly not building with gcc5 


VictorRodriguez:
HI I am trying to build fully with gcc5. However I am having the following problem: 

./../folly/Traits.h: In substitution of ‘template<class RHS, RHS rhs, class LHS> bool folly::detail::less_than_impl(typename std::enable_if<((rhs <= std::numeric_limits<_UIntType>::max()) && (rhs > std::numeric_limits<_UIntType>::min())), LHS>::type) [with RHS = long int; RHS rhs = -9223372036854775808l; LHS = bool]’:
./../folly/Traits.h:412:4:   required from ‘bool folly::less_than(LHS) [with RHS = long int; RHS rhs = -9223372036854775808l; LHS = bool]’
./../folly/Conv.h:110:5:   required from ‘typename std::enable_if<((std::is_integral<_Tp2>::value && std::is_integral<_Tp>::value) && (! std::is_same< <template-parameter-1-1>, <template-parameter-1-2> >::value)), Tgt>::type folly::to(const Src&) [with Tgt = long int; Src = bool; typename std::enable_if<((std::is_integral<_Tp2>::value && std::is_integral<_Tp>::value) && (! std::is_same< <template-parameter-1-1>, <template-parameter-1-2> >::value)), Tgt>::type = long int]’
./../folly/dynamic-inl.h:759:30:   required from ‘T folly::dynamic::asImpl() const [with T = long int]’
./../folly/dynamic-inl.h:392:68:   required from here
./../folly/Traits.h:325:10: warning: comparison of constant ‘-9223372036854775808l’ with boolean expression is always true [-Wbool-compare]
     (rhs <= std::numeric_limits<LHS>::max()
          ^
./../folly/Traits.h:326:14: warning: comparison of constant ‘-9223372036854775808l’ with boolean expression is always false [-Wbool-compare]
       && rhs > std::numeric_limits<LHS>::min()),
              ^
‘
Internal compiler error: Error reporting routines re-entered.
0x624411 build_new_op_1
	../../gcc-5.1.0/gcc/cp/call.c:5691
0x624dbe build_new_op(unsigned int, tree_code, int, tree_node*, tree_node*, tree_node*, tree_node**, int)
	../../gcc-5.1.0/gcc/cp/call.c:5750
0x6c058e build_x_binary_op(unsigned int, tree_code, tree_node*, tree_code, tree_node*, tree_code, tree_node**, int)
[...]

PR #214, opened 2015-05-27
Fix forward declaration of basic_string and list, for dual abi libstdc++

xnox:
Fix forward declaration of basic_string and list, for dual abi libstdc++ with inline std::__cxx11 namespace.

Closes facebook/folly#213
Closes facebook/folly#208

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -19,6 +19,9 @@
 #ifndef FOLLY_BASE_TRAITS_H_
 #define FOLLY_BASE_TRAITS_H_
 
+#include <list>
+#include <string>
+
 #include <memory>
 #include <limits>
 #include <type_traits>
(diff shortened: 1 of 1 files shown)


xnox:
This is an example on how to fix this problem and is trivial. There are other ways, e.g. detect that dual abi is available by querying  _GLIBCXX_USE_CXX11_ABI and if set to 1 properly forward declare inlined namespace std::* or skip forward declarations.

Can maintainers comment on this merge proposal? I don't see point to go through employeer cla, if this change is trivial / non-copyrightable / will not be merged / etc.

sgolemon:
So this is trivial enough that yeah, we could probably forego the CLA, *but* it's related to another PR #81 and there's some concern from the main folly team about including the official headers where a handful of forward declarations would do...

I'm not convinced of their argument, but we're gonna fix the underlying problem regardless.

For now, I'm going to close this as a duplicate of #81 and what we land will probably be something different from either.

TL;DR - I wouldn't worry about the CLA :)

xnox:
@sgolemon Thanks a lot. Forward declaration is a compile time optimisation, and indeed they are correct. However, libstdc++ 5.0 does something rather crazy, and the "C++11" abi declarations are wrapped in an inline namespace. Thus "std::string" is actually defined as "std::__cxx11::string" (with the __cxx11 inline namespace). I think the compiler is wrong here, and that a forward declaration should resolve to the inline namespace implementation, instead of resolving as a name conflict. Gcc people cite the holy standard, and pretend it is wrong to forward declare "std::*". If the forward declaration can be "inlined" to match __cxx11, that would work... but i have no idea how to do that, hence deployed the hammer solution of a full include. As I really don't want to get involved into the code-style battle of forward declares (i.e. if one should use them, if they should work with dual-abi libstdc++, and whose behaviour/expectations are correct).

PR #216, opened 2015-05-28
Get *=default*ed default constructors

cppfool:
Defaulted (*=default*ed) default constructors are better
because they can be constexpr and/or noexcept when C++ Standard
thinks it is right to do so. And they remain user-declared rather
than user-provided. Regarding *=default*ed default constructor,
benifit is that the work might be done during compilation and we
might not need to worry about exception paths. And for destructors,
apart from that that =defaulted definition is in sync with defaulted
default constructor we might discover that in some cases "() {}" might
be ill-formed when "= default;" compiltion indicates so. If =defaulted
definition for destructor doesn't do any harm then why not go for it.

Test Plan:

### DIFF
--- a/folly/Arena.h
+++ b/folly/Arena.h
@@ -148,8 +148,8 @@ class Arena {
     }
 
    private:
-    Block() { }
-    ~Block() { }
+    Block() = default;
+    ~Block() = default;
   } __attribute__((__aligned__));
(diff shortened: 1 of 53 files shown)


sgolemon:
Pulled (with merge conflicts resolved) for review: D2145322

PR #218, opened 2015-05-31
Using emplace_back to avoid temporary

cppfool:
Directly pass the arguments to respective constructors.
Instead of first making temporary and then pass that.

Test Plan:

### DIFF
--- a/folly/Padded.h
+++ b/folly/Padded.h
@@ -426,7 +426,7 @@ class Adaptor {
 
   void push_back(value_type x) {
     if (lastCount_ == Node::kElementCount) {
-      c_.push_back(Node());
+      c_.emplace_back();
       lastCount_ = 0;
     }
     c_.back().data()[lastCount_++] = std::move(x);
--- a/folly/io/async/test/AsyncSocketTest2.cpp
+++ b/folly/io/async/test/AsyncSocketTest2.cpp
@@ -1574,28 +1574,28 @@ class TestAcceptCallback : public AsyncServerSocket::AcceptCallback {
 
   void connectionAccepted(int fd, const folly::SocketAddress& clientAddr)
       noexcept {
-    events_.push_back(EventInfo(fd, clientAddr));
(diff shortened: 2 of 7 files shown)


sgolemon:
Pulled for review: D2151582

sgolemon:
This diff is about to be reverted because it broke some stuff, so I'm reopening the task.  I'll get it reapplied next week once we've worked out what broke and why.

cppfool:
@sgolemon 
@fugalh 
The commit de79f37492266a96e fails future_test out of the available tests.
The commit (just before(chronogically) the above one) cc7a5803e4dc passes all available tests.
future_test is a newly added one.

[  FAILED  ] 3 tests, listed below:
[  FAILED  ] Collect.collect
[  FAILED  ] Collect.collectNotDefaultConstructible
[  FAILED  ] Collect.parallel

fugalh:
Hi, these tests are newly enabled in the open source build system, and are failing on osx (maybe other platforms?). Stage 1 was to get the tests building, I will be doing the portability fixes soon.

If it's a problem that these tests are failing in the meantime, we can disable them.

These tests are not related to the original issue.

sgolemon:
fwiw, I'm not sure these are the test failures that precipitated the revert. (There's more detail on the fb-internal task).  I'll get it sorted out whatever it is....

folly/Padded.h
```diff
@@ -426,7 +426,7 @@ class Adaptor {
 
   void push_back(value_type x) {
     if (lastCount_ == Node::kElementCount) {
-      c_.push_back(Node());
+      c_.emplace_back();
```
sgolemon: Comment from internal reviewer:

not every Container supports emplace_back, this change broke unicorn build (Container == folly::TypedIOBuf<>)

code using Container = strings::TypedBuffer is also affected

sgolemon:
Resubmitted as D2156978 without folly/Padded.h change.

PR #220, opened 2015-06-02
Initial work porting Folly to windows.

Orvid:
I've tried limit the changes in this particular PR to fairly simple changes, unlikely to break other platforms.

This is not a complete port to Windows, but rather, the start of my work porting it to Windows. It requires at least MSVC 2015, due to C++11 support.


Orvid:
Due to various structural (include structure, a single include rather than a whole lot of `#ifdef`s) changes done when I implemented the file IO side, and similar changes while I'm implementing the socket functionality, I'm going to close this PR, in favor of a PR that I will open once I finish implementing the last of the socket work.

PR #222, opened 2015-06-07
Needless temporary gone

cppfool:
Summary:
We might be doing:
					1) Create a temporary
					2) Copy/Move out of it
					3) Destroy that temporary.
Which isn't needed in many places.
And copy/move elision doesn't work for a temporary
bound to a reference.

We can forward arguments, directly.

To get the work done three constructors were added.

Test Plan:

### DIFF
--- a/folly/Subprocess.cpp
+++ b/folly/Subprocess.cpp
@@ -819,7 +819,7 @@ void Subprocess::closeParentFd(int childFd) {
 std::vector<Subprocess::ChildPipe> Subprocess::takeOwnershipOfPipes() {
   std::vector<Subprocess::ChildPipe> pipes;
   for (auto& p : pipes_) {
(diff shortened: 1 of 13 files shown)


cppfool:
In my editor, text format for this diff was good for the following constructors:

ChildPipe(int fd, folly::File&& ppe);
NextProtocolsItem(int wt, const std::list<std::string>& ptcls);
CertificateInfo(const std::string& crtPath,
				const std::string& kyPath,
				const std::string& passwdPath);.

But it wasn't good in my terminal for this diff.

Also it isn't good here. Please put some effort to correct the format.
Tests have been done.

sgolemon:
@cppfool It's because you used tabs, but all fbcode projects use space indenting.  No worries, I fixed it up.

Pulled for review: D2151731

PR #223, opened 2015-06-09
Fix libc++ use

cppfool:
Use of std::max from libc++ resulted in error.
std::max found type of its arguments diffrent
whereas it expected them to be same.
This diff fix the error.

### DIFF
--- a/folly/io/async/AsyncServerSocket.cpp
+++ b/folly/io/async/AsyncServerSocket.cpp
@@ -727,7 +727,7 @@ void AsyncServerSocket::handlerReady(
     std::chrono::time_point<std::chrono::steady_clock> nowMs =
       std::chrono::steady_clock::now();
     int64_t timeSinceLastAccept = std::max(
-      int64_t(0),
+      decltype(nowMs.time_since_epoch().count()){0},
       nowMs.time_since_epoch().count() -
       lastAccepTimestamp_.time_since_epoch().count());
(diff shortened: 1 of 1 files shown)


sgolemon:
Pulled for review: D2144639

folly/io/async/AsyncServerSocket.cpp
```diff
@@ -727,7 +727,7 @@ void AsyncServerSocket::handlerReady(
     std::chrono::time_point<std::chrono::steady_clock> nowMs =
       std::chrono::steady_clock::now();
     int64_t timeSinceLastAccept = std::max(
```
jwatzman: Why not

```cpp
    int64_t timeSinceLastAccept = std::max<int64_t>(
       0,
       nowMs.time_since_epoch().count() -
       lastAccepTimestamp_.time_since_epoch().count());
```

Especially since we're already assigning this to an `int64_t` result? It seems a lot shorter and cleaner than this large `decltype`.

cppfool:
@jwatzman 
Using decltype() delays the conversion until assignment this fact
we can leverage if we'll ever want to avoid that conversion. And
that conversion delay is also the narrowing delay.

The type in the play is the decltype() gets us and std::max shouldn't
move away from that type by coerceing std::max to std::max<int64_t>.
std::max<int64_t> takes away the chance of recovery we might ever need.

PR #224, opened 2015-06-10
fix Baton.h typo

cppfool:
Reread it.

Test Plan: checked the sense


sgolemon:
Pulled for review: D2144508

PR #230, opened 2015-06-24
Remove double-conversion as a source dep in OS X script

marpaia:
double-conversion was recently added to homebrew, making it the last
folly dependency to do so. this diff modifies the existing OS X
bootstrap tool to account for that. I also added an rm for the gtest
wget, so there wasn't a leftover zip file in your repo. alternatively,
we could remove that and add it to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ autom4te.cache
 build-aux
 libtool
 folly/test/gtest-1.6.0
+folly/test/gtest-1.6.0.zip
 folly/folly-config.h
 folly/test/*_benchmark
 folly/test/*_test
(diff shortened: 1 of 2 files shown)


sgolemon:
I vote for adding it to .gitignore

What will it take to get gtest into homebrew?

fugalh:
$ brew search gtest
Installing gtest system-wide is not recommended; it should be vendored
in your projects that use it.

Google has the annoying position that gtest (and gmock) Should Not Be Packaged. So it would take ignoring their guidance, and talking whatever gatekeepers there are for homebrew into doing the same.

sgolemon:
Merged as b75ef0a0af48766298ebcc946dd31fe0da5161e3

PR #235, opened 2015-07-06
Port Folly to MSVC 2015 RC

Orvid:
This is the foundation of my work in progress port of HHVM to Windows. Once this is in, and HHVM's third-party repo is updated, I can begin to submit PRs porting pieces of HHVM to windows.

This ignores most of the wangle directory, and anything excluded by the HHVM CMake file.
There are various temporary hacks in place, cordoned off by #ifdef's of #defines that are defined in Portability.h, which will need to be either removed or reviewed once VS2015 RTM is released.


sgolemon:
On the whole, looks awesome.  FYI wangle is now at github.com/facebook/wangle as a separate project and it's about to vanish from folly/ (HHVM will be switching to pulling it in separately soon).

Soon as I finish catching up after what became a long weekend (don't ask), I'll pull this in for review.  Might take a lil' while because of the number of files touched, but should be managable. (In the future, this could have been done as multiple small diffs which tend to get reviewed faster as they're more digestible -- Plus you get a higher commit count :D ).

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+}
+
+inline char* asctime_r(const tm* tm, char* buf) {
+  char tmpBuf[64];
+  if (asctime_s(tmpBuf, sizeof(tmpBuf), tm))
+    return NULL;
+  return strcpy(buf, tmpBuf);
+}
+
+inline void bzero(void* s, size_t n) {
```
ariccio: Since this is for Windows, you might as well annotate `s` with `_Pre_writable_size_(n)`

ariccio: Actually, you might as well go ahead and declare it like this for maximum static analysis effectiveness:
```
_At_buffer_(s, _Iter_, n, _Post_satisfies_(s[_Iter_] == 0))
Inline void bzero(  _Pre_writable_size_(n) _Out_ void* s, size_t n )

```

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+}
+
+inline unsigned int alarm(unsigned int seconds) {
+  SetTimer(NULL, 0, (UINT)(seconds * 1000), __alarm_callback_func);
+  return 0;
+}
+
+inline char* asctime_r(const tm* tm, char* buf) {
+  char tmpBuf[64];
+  if (asctime_s(tmpBuf, sizeof(tmpBuf), tm))
```
ariccio: You shouldn't need to explicitly specify the buffer size,  because [Secure Template Overloads](https://msdn.microsoft.com/en-us/library/ms175759.aspx) allow for [template parameter inference](https://randomascii.wordpress.com/2013/04/03/stop-using-strncpy-already/), which is pretty cool.

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+inline void __stdcall __alarm_callback_func(HWND, UINT, UINT_PTR, DWORD) {
+  raise(SIGALRM);
+}
+
+inline unsigned int alarm(unsigned int seconds) {
+  SetTimer(NULL, 0, (UINT)(seconds * 1000), __alarm_callback_func);
+  return 0;
+}
+
+inline char* asctime_r(const tm* tm, char* buf) {
```
ariccio: `buf` should be `_Out_writes_z_(64)`, and the return value should be `_Ret_writes_maybenull_z_(64)`.

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+inline void bzero(void* s, size_t n) {
+  ZeroMemory(s, n);
+}
+
+#define DT_UNKNOWN 0
+#define DT_DIR     1
+#define DT_REG     2
+#define DT_LNK     3
+struct dirent {
+  unsigned char d_type;
```
ariccio: Should be `_Field_range_(DT_UNKNOWN, DT_LNK)`

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+  char* d_name;
```
ariccio: `_Field_z_`

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+};
+
+struct DIR {
+  dirent dir;
+  HANDLE hand;
+  int cnt;
+  char name[MAX_PATH * 3];
+};
+
+inline int closedir(DIR* dir) {
```
ariccio: Protip : Mark it `_Post_ptr_invalid_` and you'll be able to detect use-after-free and double free bugs.

Marking the function `_At_(dir->hand, _Post_ptr_invalid_)` will do the same for double handle closes.


Is `FindClose` what you wanted to use here, or is it `CloseHandle`?

Also, why does this return an `int`, if it doesn't check any return values?

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+}
+
+inline char* ctime_r(const time_t* t, char* buf) {
+  char tmpBuf[64];
+  if (ctime_s(tmpBuf, sizeof(tmpBuf), t))
+    return NULL;
+  return strcpy(buf, tmpBuf);
+}
+
+inline int dprintf(int fd, const char *fmt, ...) {
```
ariccio: Mark `fmt` as `_Printf_format_string_` for format string error detection!

folly/CPortability.h
```diff
@@ -53,4 +53,725 @@
...
+  rlim_t rlim_cur;
+  rlim_t rlim_max;
+};
+#define RLIMIT_CORE 0
+#define RLIMIT_NOFILE 0
+#define RLIMIT_DATA 0
+#define	RLIMIT_STACK 3
+#define RLIM_INFINITY SIZE_MAX
+
+inline int getrlimit(int type, rlimit* dst) {
```
ariccio: Mark the function as `_Success_( return == 0 )` and `dst` as `_Out_` to detect uninit memory usage.

folly/CPortability.h
Orvid: If this weren't in the C portability header, that would be true, but, as it is in the C portability header, I have kept the file compilable as C. I could do an `#ifdef` for C++ for that though.

folly/CPortability.h
Orvid: `FindClose` is indeed what I want, as this is closing the find handle, not a normal handle.

This returns an int because the standard `closedir` API returns an int. I've added a check for errors on the `FindClose` call, but I won't be marking the function for analysis, as this is intended as a functional portability mechanism, not an extensive API.

Orvid:
I've put the annotations in place in a couple of locations, but I've decided not to add most of them, as they would error the same on Posix platforms as they do on Windows.

folly/CPortability.h
ariccio: Fair enough.

ariccio:
Ok.

Personally, I've grown to be a major believer in annotations, *especially* in library code. I've been slowly annotating SQLite (chosen because it's *so well tested*) and have already come across all sorts of questionable code that no human would've ever noticed.

 It's also pretty cool to get compile-time errors for things that'd otherwise not show up until testing.

OSR has also fallen in love with annotations: ["SAL Annotations: Don’t Hate Me Because I’m Beautiful"](https://www.osr.com/blog/2015/02/23/sal-annotations-dont-hate-im-beautiful/)

ariccio:
(if you're curious about the SQLite annotations):
I plan to eventually have a drop-in compatible amalgamation for any dev who uses SQLite in their program, to detect misuse of the (classically bad) C API. The docs are pretty complex as it is, so having a static analysis capability is extremely valuable.

Orvid:
I can probably split this into 3 separate PRs if that would make this any easier. I might be able to make a few smaller ones as well, but there aren't nearly as many changes that are just MSVC specific that don't require the modifications to the 4 core headers, `Portability.h`, `CPortability.h`, `FilePortability.h`, and `SocketPortability.h` as there are in HHVM.

Orvid:
I am, for now at least, going to close this PR. I'll likely re-open it later split into a few different ones.

PR #238, opened 2015-07-12
Added the (stripped) root headers used by the MSVC port

Orvid:
This includes the headers used by the Windows port, but doesn't include any of the actual content of the headers for the Windows version. This is intended to allow PRs to be made to HHVM for header adjustments while the changes that are part of the core headers are still under review.
This also includes the full changes to `Portability.h`, to allow PRs based on the changes in it to be made as well.


Orvid:
This will be handled as part of my new set of PRs instead.

PR #244, opened 2015-07-22
Port fbstring_core to big-endian architectures.

cole14:
There's 2 ways this could be implemented - either as a series of preprocessor blocks depending on target architecture (as I have implemented it here), or by encapsulating access to MediumLarge::capacity_ within getters/setters as in a similar manner to setSmallSize() and smallSize(). The first option makes the code a bit harder to read, but the second option changes the existing control flow a bit which could slightly alter performance.

I opted for the first so as to keep the existing amd64 flow untouched, but can easily change the pull request to the second option to keep code readability a priority.

### DIFF
--- a/folly/docs/FBString.md
+++ b/folly/docs/FBString.md
@@ -9,8 +9,8 @@ allocator. In particular, `fbstring` is designed to detect use of
 jemalloc and cooperate with it to achieve significant improvements in
 speed and memory usage.
 
-`fbstring` supports x32 and x64 architectures. Porting it to big endian
-architectures would require some changes.
+`fbstring` supports 32- and 64-bit and little- and big-endian
+architectures.
 
 ### Storage strategies
 ***
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -266,8 +266,8 @@ private:
 
 /**
  * This is the core of the string. The code should work on 32- and
- * 64-bit architectures and with any Char size. Porting to big endian
- * architectures would require some changes.
+ * 64-bit and both big- and little-endianan architectures with any
+ * Char size.
  *
  * The storage is selected as follows (assuming we store one-byte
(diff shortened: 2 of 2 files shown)


chrismwendt:
The documentation at folly/docs/FBString.md might need to be updated as well.

cole14:
Updated folly/docs/FBString.md to reflect that both big and little endian is now supported as per @chrismwendt

Orvid:
It does appear that a good portion of the ifdefs could be abstracted out and centralized. That might make the code quite a bit cleaner.

cole14:
Updated to combine most of the ifdefs into a single MediumLarge::setCapacity function, greatly improving code readability.

folly/docs/FBString.md
```diff
@@ -9,8 +9,8 @@ allocator. In particular, `fbstring` is designed to detect use of
 jemalloc and cooperate with it to achieve significant improvements in
 speed and memory usage.
 
-`fbstring` supports x32 and x64 architectures. Porting it to big endian
-architectures would require some changes.
+`fbstring` supports 32- and 64-bit and little- and big-endian
```
Orvid: I don't typically see `32-bit` as `32-`, even when referenced immediately before `64-bit`. Normally it's just `32`. The same case can be made for `little-`.

cole14:
Any update on this PR?

JoelMarcey:
We had an internal review of this and @yfeldblum had a comment...

> Use `constexpr` throughout instead of macros.

> OK to define the `constexpr` once at the top in terms of the macro though if necessary. Clang doesn't allow `reinterpret_cast` in `constexpr` expressions, so tricks like testing byte 0 of a 2-byte word or this may not be possible.

Orvid:
It's worth noting that, except in the root evaluation scope, which they don't document, MSVC doesn't allow `reinterpret_cast` in `constexpr` expressions either.

PR #247, opened 2015-07-24
Adjust Format-inl.h to work with MSVC

Orvid:
This adjusts a format string to be sane, I don't think `'` is a valid format specifier to begin with, and MSVC2015 added a warning for invalid specifiers, which this triggers.
This also adjusts a use of `std::enable_if` to eliminate the `enable_if` entirely, because MSVC doesn't like it when you try to do a sizeof check on a type like this. (I haven't checked, but I wouldn't be surprised if MSVC were just returning `-1` for the size here, either way, the new way it's written should still have the desired effect)

### DIFF
--- a/folly/Format-inl.h
+++ b/folly/Format-inl.h
@@ -471,7 +471,12 @@ class FormatValue<
 
       valBufBegin = valBuf + 3;  // room for sign and base prefix
       int len = snprintf(valBufBegin, (valBuf + valBufSize) - valBufBegin,
-                         "%ju", static_cast<uintmax_t>(uval));
+#ifdef _MSC_VER
+                         "%ju",
+#else
+                         "%'ju",
+#endif
+                         static_cast<uintmax_t>(uval));
       // valBufSize should always be big enough, so this should never
       // happen.
       assert(len < valBuf + valBufSize - valBufBegin);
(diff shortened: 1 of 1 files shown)


folly/Format-inl.h
```diff
@@ -471,7 +471,7 @@ class FormatValue<
 
       valBufBegin = valBuf + 3;  // room for sign and base prefix
       int len = snprintf(valBufBegin, (valBuf + valBufSize) - valBufBegin,
-                         "%'ju", static_cast<uintmax_t>(uval));
```
sgolemon: ' provides thousands separation.  So I would say the very least would be something along the lines of:

```
#ifdef FOLLY_NO_PRINTF_GROUPING
  "%ju",
#else
  "%'ju",
#endif
  static_cast<uintmax_t>(uval));
```

And add a configure check to conditionally set that define.

Better would be to implement a thousands separating output for non-gnu platforms, but that's a bit more work.

Orvid:
As it's gotten by so far like this, I'll avoid the configure check and just special case it for MSVC.

sgolemon:
Eh, fair 'nuff.  Can always add a configure check or something later.

tudor:
This changes behavior, so I won't merge it.

The 'n' format specifier should use a locale-aware thousands separator, per https://docs.python.org/2/library/string.html#format-specification-mini-language (which folly::Format is mimicking). On Windows, you probably want to use GetNumberFormatEx. https://msdn.microsoft.com/en-us/library/dd318113(v=VS.85).aspx

Orvid:
I've updated the branch to use `GetNumberFormat`, this just needs to be re-opened, or I need to create another PR.

PR #250, opened 2015-07-24
Implement Random.cpp for MSVC

Orvid:
This uses `<random>` to implement it, because there is no `/dev/urandom` on Windows.

### DIFF
--- a/folly/Random.cpp
+++ b/folly/Random.cpp
@@ -26,18 +26,26 @@
 #include <folly/File.h>
 #include <folly/FileUtil.h>
 
+#ifdef _MSC_VER
+#include <wincrypt.h>
+#endif
+
 namespace folly {
 
 namespace {
 
 void readRandomDevice(void* data, size_t size) {
 #ifdef _MSC_VER
-  static std::uniform_int_distribution<int> dist(0, 255);
-  static std::mt19937_64 randEng;
-  uint8_t* data2 = (uint8_t*)data;
-  for (size_t i = 0; i < size; i++) {
-    data2[i] = (uint8_t)dist(randEng);
+  static bool initialized = false;
+  static HCRYPTPROV cryptoProv;
+  if (!initialized) {
+    bool bcac = CryptAcquireContext(&cryptoProv, nullptr, nullptr, PROV_RSA_FULL, 0);
+    PCHECK(bcac);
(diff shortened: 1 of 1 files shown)


sgolemon:
From @yfeldblum ...

This looks deterministic. See: http://www.cplusplus.com/reference/random/mersenne_twister_engine/mersenne_twister_engine/.

That is not OK for readRandomDevice, even on Windows. Windows has CryptGenRandom, though - perhaps that's something that can be used here.

folly/Random.cpp
```diff
@@ -26,16 +26,33 @@
...
+#include <wincrypt.h>
+#endif
+
 namespace folly {
 
 namespace {
 
 void readRandomDevice(void* data, size_t size) {
+#ifdef _MSC_VER
+  static bool initialized = false;
```
sgolemon: (From @yfeldblum) This is not threadsafe. Two threads that simultaneously call readRandomDevice can overwrite each other's data. Using std::once_flag should help.

```
#include <mutex>

static std::once_flag flag;
static HCRYPTPROV provider;
std::call_once(flag, [&] }
  PCHECK(CryptAcquireContext(&provider, nullptr, nullptr, PROV_RSA_FULL, 0));
});
CHECK(size <= std::numeric_limits<DWORD>::max()); // not PCHECK
PCHECK(CryptGenRandom(provider, (DWORD)size, (BYTE*)data);
```

The other possibility is to use ThreadLocal.

Orvid: That was actually deliberate, and is a result of my own personal form of insanity. As the impact of a second call to `CryptoAcquireContext` is minimal on resource usage, basically just incrementing a reference count, and as `cryptoProv` will be entirely written from a single instruction, there is no possibility of getting part of one thread's info and part of another's. It's perfectly fine for a second thread to be trying to initialize the provider at the same time, that's why it's setting initialized after the provider has been acquired.

If it's preferred to use a mutex though, I can certainly do that?

yfeldblum: Is this behavior of `CryptAcquireContext` specified or documented somewhere?

Orvid: I believe my thought process on that might have come from the way `CryptReleaseContext` is documented. Either way, I've switch this to a proper mutex instead.

PR #251, opened 2015-07-24
Fix Random-inl.h under MSVC

Orvid:
With MSVC, `seedData.begin()` does not return a `void*`, so get a pointer to the data, and explicitly cast that to a `void*` instead.


sgolemon:
Feedback from @yfeldblum:

Given:

```
class Random {
public:
  static void secureRandom(void* data, size_t len);
}
```

We need an address. The proper way to do that with std::array<T, N> is using array::data().
http://www.cplusplus.com/reference/array/array/data/

This solution is general, and does not need to be specialized to MSVC.

PR #253, opened 2015-07-24
Handle some scoping issues in Conv.h under MSVC

Orvid:
No idea why these are the only two places it complains about, but it does.
This just explicitly scopes them.

### DIFF
--- a/folly/Conv.h
+++ b/folly/Conv.h
@@ -815,7 +815,7 @@ typename std::enable_if<sizeof...(Ts) >= 3
     typename detail::last_element<Ts...>::type
   >::type>::value>::type
 toAppend(const Ts&... vs) {
-  detail::toAppendStrImpl(vs...);
+  ::folly::detail::toAppendStrImpl(vs...);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


sgolemon:
What error, specifically, was this producing?

Orvid:
I believe these particular ones were complaining about `detail` being ambiguous.

sgolemon:
Can you reproduce and tell us which namespace it's confusing it with?  Would like to understand the scope of the issue.

Orvid:
I believe this is simply MSVC being weird about when it resolves types in templates:
```
F:\Projects\hhvm\third-party\folly\src\folly/Conv.h(845): error C2872: 'detail': ambiguous symbol (compiling source file F:\Projects\hhvm\third-party\folly\src\folly\io\RecordIO.cpp)
  F:\Projects\hhvm\third-party\folly\src\folly/String-inl.h(653): note: could be 'folly::detail' (compiling source file F:\Projects\hhvm\third-party\folly\src\folly\io\RecordIO.cpp)
  F:\Projects\hhvm\third-party\folly\src\folly/io/RecordIO-inl.h(62): note: or       'folly::recordio_helpers::detail' (compiling source file F:\Projects\hhvm\third-party\folly\src\folly\io\RecordIO.cpp)
```

PR #263, opened 2015-07-24
Implement a generalized mechansim for pushing/poping warnings

Orvid:
This is based on the design used in libGlog.
This replaces a few of the uses of `#pragma GCC diagnostic push/pop`, but does not replace all of them.


Orvid:
Replaced internally with D2866264.

PR #264, opened 2015-07-24
Add and implement a couple of fallback defines for MSVC

Orvid:
Specifically, this defines `MSVC_NO_NONVOID_ATOMIC_IF`, `MSVC_NO_STATIC_INCLASS_CONSTEXPR_INITIALIZATION`, and `MSVC_NO_CONSTEXPR_EASY_POINTER_CASTS` if the current version of MSVC is the 2015 RTM or below.
This also implements the places that needed these defines.

### DIFF
--- a/folly/AtomicHashArray.h
+++ b/folly/AtomicHashArray.h
@@ -128,28 +128,19 @@ class AtomicHashArray : boost::noncopyable {
     int    entryCountThreadCacheSize;
     size_t capacity; // if positive, overrides maxLoadFactor
 
-#ifdef MSVC_NO_CONSTEXPR_EASY_POINTER_CASTS
   private:
     static constexpr KeyT k_emptyKey = (KeyT)-1;
     static constexpr KeyT k_lockedKey = (KeyT)-2;
     static constexpr KeyT k_erasedKey = (KeyT)-3;
+    
   public:
-#endif
-
-    constexpr Config() :
-#ifdef MSVC_NO_CONSTEXPR_EASY_POINTER_CASTS
-      emptyKey(k_emptyKey),
-      lockedKey(k_lockedKey),
-      erasedKey(k_erasedKey),
-#else
-      emptyKey((KeyT)-1),
-      lockedKey((KeyT)-2),
-      erasedKey((KeyT)-3),
-#endif
-      maxLoadFactor(0.8),
-      growthFactor(-1),
-      entryCountThreadCacheSize(1000),
-      capacity(0) {}
+    constexpr Config() : emptyKey(k_emptyKey),
+                         lockedKey(k_lockedKey),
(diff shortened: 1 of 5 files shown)


folly/AtomicHashArray.h
```diff
@@ -128,13 +128,28 @@ class AtomicHashArray : boost::noncopyable {
...
     size_t capacity; // if positive, overrides maxLoadFactor
 
-    constexpr Config() : emptyKey((KeyT)-1),
-                         lockedKey((KeyT)-2),
-                         erasedKey((KeyT)-3),
-                         maxLoadFactor(0.8),
-                         growthFactor(-1),
-                         entryCountThreadCacheSize(1000),
-                         capacity(0) {}
+#ifdef MSVC_NO_CONSTEXPR_EASY_POINTER_CASTS
```
sgolemon: Why not just always use the constexpr values and avoid the #ifdefs?

folly/AtomicLinkedList.h
```diff
@@ -63,11 +64,19 @@ class AtomicLinkedList {
    * Note: list must be empty on destruction.
    */
   ~AtomicLinkedList() {
+#ifdef MSVC_NO_NONVOID_ATOMIC_IF
```
sgolemon: Maybe reduce this assertion to just:

```
  ~AtomicLinkedList() {
    assert(empty());
  }
```

And avoid duplicating the logic and the ifdefs.

folly/AtomicHashArray.h
Orvid: I can certainly do that. My goal with the port was to make as few changes as possible to the code that effects more than just MSVC, which is why I only enabled it for MSVC at the moment.

folly/Portability.h
```diff
@@ -284,6 +284,17 @@ inline size_t malloc_usable_size(void* ptr) {
 # define FOLLY_HAS_RTTI 1
 #endif
 
+#if defined(_MSC_VER) && _MSC_FULL_VER <= 190023026 // 2015 RTM or below
```
sgolemon: How concerned do we need to be about checking for which pre-relesae we're working with?
Seems like we can just assume "MSVC 2015 Latest" until final comes out.

Can we kill this check and assume the two defines below are just not set?
Makes the code a lot simpler if so.

Orvid: RTM is the release version, it isn't  a prerelease, but there is a good chance that these will be fixed in an update, which is why these are set to go away when the update is released. This is basically future-proofing, as the eventual goal is to remove them completely.

Orvid:
Ekk... When this was merged, it excluded the defines that are present. Those defines were there in order to disable the workarounds on future versions of MSVC, those defines are still needed for the current release version of MSVC 2015.

PR #265, opened 2015-07-24
Don't do memory operations in chunks on MSVC

Orvid:
Because, especially with the way that munmap is implemented in my windows port, this is a very bad idea. Plus, it's completely unnecessary on Windows.


sgolemon:
Feedback from @yfeldblum 

```
This is the minimal change, but having memOpInChunks do a full (non-chunked) mem op on Windows is weird naming.

Given the commit description, the callsites which do mem ops in case of failure of the first mem op need to be fixed as well, specifically MemoryMapping::mlock.
```

Orvid:
After checking, `mlock` is the only place that did anything in the case of a failure, so I've disabled the `munlock` call there for MSVC.

PR #274, opened 2015-07-24
Implement MSVC support for AtomicHashUtils.h

Orvid:
Becuase MSVC doesn't support inline asm in 64-bit mode.


ariccio:
Not [`YieldProcessor`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms687419.aspx)?

Orvid:
I chose to do it this way so that I'm not forcing the inclusion of the full windows headers. The result is the same.

PR #276, opened 2015-07-24
Exclude the FB_DEC_TYPE section in dynamic-inl.h under MSVC

Orvid:
I have no idea what this section is intended to do, MSVC produces a bunch of errors if it's included, and both Folly and HHVM build just fine with it excluded, so exclude it for MSVC.


Orvid:
MSVC handles the new code just fine, this is no longer needed.

PR #277, opened 2015-07-24
Fix a parse error in detail/ThreadLocalDetail.h under MSVC

Orvid:
I have no idea why this specific declaration would error, or if there is a difference in meaning, but this fixes MSVC's absurd parser error.

### DIFF
--- a/folly/detail/ThreadLocalDetail.h
+++ b/folly/detail/ThreadLocalDetail.h
@@ -450,7 +450,7 @@ struct StaticMeta {
 
 #ifdef FOLLY_TLD_USE_FOLLY_TLS
 template <class Tag>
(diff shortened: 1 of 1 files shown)


ariccio:
Parse error? Do you mean an *internal compiler error*, or something else?

Orvid:
No, I really do mean a parse error. For some absurd reason, MSVC ends up complaining about expecting a `;` before the end of the document in anything that includes this without this change. This one had me beating my head against a wall trying to figure out where the error was for 3 full days before I finally figured it out. I would have filed an issue against MSVC about it, but I haven't been able to reproduce it anywhere other than this one place in this one file.

ariccio:
Holy crap, that's insane. Carry on.

PR #278, opened 2015-07-24
Don't warn if pthread_atfork isn't avaliable on MSVC

Orvid:
Because we don't even have `fork` to begin with under MSVC.


folly/detail/ThreadLocalDetail.h
```diff
@@ -212,7 +212,7 @@ struct StaticMeta {
                          /*parent*/ &StaticMeta::onForkParent,
                          /*child*/ &StaticMeta::onForkChild);
     checkPosixError(ret, "pthread_atfork failed");
-#elif !__ANDROID__
+#elif !__ANDROID__ && !_MSC_VER
```
sgolemon: Shouldn't this be `!defined(_MSC_VER)` ?

Orvid: Yes.

PR #279, opened 2015-07-24
Handle MSVC correctly in detail/IPAddress.h

Orvid:
Winsock's structures don't have a member named `s6_addr16`, so use the name for Winsock's structure.

### DIFF
--- a/folly/detail/IPAddress.h
+++ b/folly/detail/IPAddress.h
@@ -283,7 +283,11 @@ inline std::string fastIpv4ToString(
 }
 
 inline std::string fastIpv6ToString(const in6_addr& in6Addr) {
+#ifdef _MSC_VER
+  const uint16_t* bytes = reinterpret_cast<const uint16_6*>(&in6Addr.u.Word);
+#else
   const uint16_t* bytes = reinterpret_cast<const uint16_t*>(&in6Addr.s6_addr16);
+#endif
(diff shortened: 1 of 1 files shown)


ariccio:
While you're at it, remind me: 

- Is folly configured to use `/analyze`? I've seen lots of well tested libraries use networking APIs incorrectly, in ways that can be detected by `/analyze`.
- Do folly tests run under Application Verifier on Windows? Essentially every library (libraries used in `git`, `nmap`, etc...) that I've run under Application Verifier has unnoticed issues with their use of networking APIs.

Orvid:
I don't believe folly is configured to use `/analyze`, as the only Windows support prior to my set of PRs here was in this file and `Bits.h`, both of which had their own set of issues.

I've actually not tried to run the folly tests under it.

folly/detail/IPAddress.h
```diff
@@ -283,7 +283,11 @@ inline std::string fastIpv4ToString(
 }
 
 inline std::string fastIpv6ToString(const in6_addr& in6Addr) {
+#ifdef _MSC_VER
+  const uint16_t* bytes = reinterpret_cast<const uint16_6*>(&in6Addr.u.Word);
```
sgolemon: `uint16_6` looks like a typo, can you confirm that?

Orvid: Yes, it is. Looks like it happened when I was extracting the changes into separate PRs.

PR #280, opened 2015-07-24
Implement clock_getres for Windows

Orvid:
Because the windows version of pthread very definitely does not implement the clock. The Cygwin and MinGW headers might, so let them still fall back to the pthread includes.

### DIFF
--- a/folly/detail/Clock.cpp
+++ b/folly/detail/Clock.cpp
@@ -48,7 +48,125 @@ int clock_getres(clockid_t clk_id, struct timespec* ts) {
 
   return 0;
 }
-#elif _MSC_VER
+#elif defined(_MSC_VER)
+// The MSVC version has been extracted from the pthreads implemenation here:
+// https://github.com/songdongsheng/libpthread
+// Copyright(c) 2011, Dongsheng Song <songdongsheng@live.cn>
+//
+// It is under the Apache License Version 2.0, just as the rest of the file is.
+// It has been mostly stripped down to what we have.
+
+#include <WinSock2.h>
+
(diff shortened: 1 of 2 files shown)


folly/detail/Clock.cpp
```diff
@@ -48,7 +48,125 @@ int clock_getres(clockid_t clk_id, struct timespec* ts) {
 
   return 0;
 }
-#elif _MSC_VER
+#elif defined(_MSC_VER)
+// The MSVC version has been extracted from the pthreads implemenation here:
+// https://github.com/songdongsheng/libpthread
+// Copyright(c) 2011, Dongsheng Song <songdongsheng@live.cn>
+//
+// It is under the Apache License Version 2.0, just as the rest of the file is.
```
sgolemon: Unfortunately, we can't take this.  Could you come up with another approach, or an implementation from a more permissive license?

Orvid: This is the most permissive license of a proper implementation of this that I could find. It should be the same as the rest of the file, unless I missed something? Re-implementing it is possible, as trying to work around not having it will be more of a pain, as it's used in a few spots in HHVM.

sgolemon: I don't want to go into legalese as I'm not a lawyer, but as I understand it, the only Apache licensed code we can use is code which the actual copyright holder gives us explicit rights to, and I believe the code already in this file falls under that umbrella.

This new code comes from another party.  If you can get him to sign our CLA and contribute it under those terms, we can take it.  Otherwise.... sigh... :(

PR #281, opened 2015-07-24
Handle less_than_impl and greater_than_impl under MSVC

Orvid:
MSVC chokes on the template constraints of these functions, so use a single implementation that will be const-folded by the optimizer instead.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -312,6 +312,13 @@ struct is_negative_impl<T, false> {
   constexpr static bool check(T) { return false; }
 };
 
+// folly::to integral specializations can end up generating code
+// inside what are really static ifs (not executed because of the templated
+// types) that violate -Wsign-compare so suppress them in order to not prevent
+// all calling code from using it.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-compare"
+
 template <typename RHS, RHS rhs, typename LHS>
 bool less_than_impl(LHS const lhs) {
   return
@@ -319,6 +326,8 @@ bool less_than_impl(LHS const lhs) {
     rhs <= std::numeric_limits<LHS>::min() ? false :
     lhs < rhs;
 }
+
+#pragma GCC diagnostic pop
 
 template <typename RHS, RHS rhs, typename LHS>
 bool greater_than_impl(LHS const lhs) {
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -312,108 +312,21 @@ struct is_negative_impl<T, false> {
   constexpr static bool check(T) { return false; }
 };
 
-// Early versions of MSVC choke on the template
-// constraints.
-#if _MSC_FULL_VER <= 200022816 // 2015 RC or below
-
-FOLLY_PUSH_WARNING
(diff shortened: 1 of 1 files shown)


folly/Traits.h
```diff
@@ -312,6 +312,36 @@ struct is_negative_impl<T, false> {
   constexpr static bool check(T x) { return false; }
 };
 
+// Early versions of MSVC choke on the template
+// constraints.
+#if _MSC_FULL_VER <= 200022816 // 2015 RC or below
```
sgolemon: AIUI, 2015 final doesn't need this, yes?  If so, we can drop this PR entirely as we're going to declare 2015 final as the minimum version for folly.

Orvid: Unfortunately it does still need it. I believe that is the version number for the RTM release, so the comment must not have gotten updated.

JoelMarcey:
@Orvid - Can you please rebase this pull request? Thanks.

Orvid:
Rebased, although there weren't any conflicts...

JoelMarcey:
Strange. There were when I tried to import it a few minutes ago before your rebase. Now that you rebased we are good. So it helped me at least :)

folly/Traits.h
```diff
@@ -312,6 +312,36 @@ struct is_negative_impl<T, false> {
   constexpr static bool check(T) { return false; }
 };
 
+// Early versions of MSVC choke on the template
+// constraints.
+#if _MSC_FULL_VER <= 200022816 // 2015 RC or below
+
+FOLLY_PUSH_WARNING
```
JoelMarcey: Where is this actually defined?

folly/Traits.h
```diff
@@ -312,6 +312,36 @@ struct is_negative_impl<T, false> {
...
+template <typename RHS, RHS rhs, typename LHS>
+bool greater_than_impl(LHS const lhs) {
+    if (rhs <= std::numeric_limits<LHS>::max() && rhs >= std::numeric_limits<LHS>::min())
+        return lhs > rhs;
+    else if (rhs > std::numeric_limits<LHS>::max())
+        return false;
+    else // if (rhs < std::numeric_limits<LHS>::min())
+        return true;
+}
+FOLLY_POP_WARNING
```
JoelMarcey: ...and this?

JoelMarcey:
There is some discussion about just using your MSVC workaround for everything instead of having the preprocessor.

Comment internally is: 

> The templated version is clever, but I presume the static checks in the non-templated version should be evaluated at compile time too. Therefore, the MSVC version seems like it should be just as optimal as the templated version. Why not just switch to it completely? Make the statically-testable conditions come first, though, instead of last."

What do you think?

folly/Traits.h
Orvid: Ah, woops, hadn't realized that slipped into this PR. It's from #263, which hasn't been merged yet.

folly/Traits.h
Orvid: See reply above.

Orvid:
I'm fine with making this the only implementation, it's far cleaner than the current one.

PR #285, opened 2015-07-24
Handle platforms where pthread_t is a structure

Orvid:
These platforms include MSVC, and I don't know of any others. This implements the changes needed for this to work properly on MSVC.

### DIFF
--- a/folly/Hash.h
+++ b/folly/Hash.h
@@ -23,6 +23,10 @@
 #include <utility>
 #include <tuple>
 
+#ifdef _MSC_VER
+#include <pthread.h>
+#endif
+
 #include <folly/ApplyTuple.h>
 #include <folly/SpookyHashV1.h>
 #include <folly/SpookyHashV2.h>
@@ -89,6 +93,14 @@ size_t hash_combine_generic(const T& t, const Ts&... ts) {
 // strings or pairs
 class StdHasher {
  public:
+#ifdef _MSC_VER
+  // std::hash doesn't support structs, but, on windows,
+  // pthread_t is a struct, so combine the hash of the
+  // two members.
+  static size_t hash(const pthread_t& t) {
+    return hash_combine_generic<StdHasher>(t.p, t.x);
+  }
+#endif
   template <typename T>
   static size_t hash(const T& t) {
     return std::hash<T>()(t);
--- a/folly/Hash.h
+++ b/folly/Hash.h
@@ -23,10 +23,6 @@
 #include <utility>
 #include <tuple>
 
-#ifdef _MSC_VER
-#include <pthread.h>
-#endif
-
 #include <folly/ApplyTuple.h>
 #include <folly/SpookyHashV1.h>
 #include <folly/SpookyHashV2.h>
@@ -93,14 +89,6 @@ size_t hash_combine_generic(const T& t, const Ts&... ts) {
 // strings or pairs
 class StdHasher {
  public:
-#ifdef _MSC_VER
-  // std::hash doesn't support structs, but, on windows,
-  // pthread_t is a struct, so combine the hash of the
-  // two members.
-  static size_t hash(const pthread_t& t) {
-    return hash_combine_generic<StdHasher>(t.p, t.x);
-  }
-#endif
   template <typename T>
   static size_t hash(const T& t) {
     return std::hash<T>()(t);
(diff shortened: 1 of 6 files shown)


folly/io/async/NotificationQueue.h
```diff
@@ -225,7 +225,7 @@ class NotificationQueue {
     : eventfd_(-1),
       pipeFds_{-1, -1},
       advisoryMaxQueueSize_(maxSize),
-      pid_(getpid()),
+      pid_((pid_t)getpid()),
```
ariccio: Huh?

Orvid: With the windows implementation of pthread, it defines `pid_t` as a `void*`, and `getpid()` returns an `int`. MSVC doesn't like this implicit construction of a pointer from a raw integer and produces an error. Explicitly casting it solves the problem.

ariccio: Ahh, ok. Any reason why it's not a C++ cast?

Orvid: Personal habit.

folly/Portability.h
```diff
@@ -284,6 +284,66 @@ inline size_t malloc_usable_size(void* ptr) {
...
+
+inline bool operator !=(pthread_t ptA, pthread_t ptB) {
+  return pthread_equal(ptA, ptB) == 0;
+}
+
+inline bool operator !(pthread_t ptA) {
+  return ptA == 0;
+}
+
+#define pthread_t_init (pthread_t{NULL, 0})
```
sgolemon: Note from @yfeldblum
```
constexpr. Maybe name this pthread_zero.

(IIRC the _t stands for typedef, i.e., typedef struct pthread pthread_t;, and probably shouldn't be present in value names.)
```

Note from @meyering:
```
I too prefer pthread_zero or even pthread_0 -- sounds more like a value than pthread_t_init, which might evoke something that performs an initialization. That's a little closer to the uintmax_t -> UINTMAX_MAX mapping.
```

folly/Portability.h
```diff
@@ -284,6 +284,66 @@ inline size_t malloc_usable_size(void* ptr) {
...
+
+inline int pthread_attr_getstack(
+  pthread_attr_t* attr,
+  void** stackaddr,
+  size_t* stacksize) {
+  pthread_attr_getstackaddr(attr, stackaddr);
+  pthread_attr_getstacksize(attr, stacksize);
```
sgolemon: Note from @meyering:
```
Don't we want to return nonzero if either of these functions fails?
```

folly/Portability.h
```diff
@@ -284,6 +284,66 @@ inline size_t malloc_usable_size(void* ptr) {
...
+  pthread_attr_setstacksize(attr, stacksize);
+  return 0;
+}
+
+inline int pthread_attr_getguardsize(pthread_attr_t* attr, size_t* guardsize) {
+  *guardsize = 0;
+  return 0;
+}
+#else
+#define pthread_t_init ((pthread_t)0)
```
sgolemon: Note from @yfeldblum: `same`

folly/Portability.h
Orvid: I'll go with pthread_zero then. I believe I have an additional bit locally for this, and I need to implement a hasher for it for std::hash, as HHVM uses pthread_t as a key.

folly/Hash.h
```diff
@@ -88,6 +92,14 @@ size_t hash_combine_generic(const T& t, const Ts&... ts) {
 // strings or pairs
 class StdHasher {
  public:
+#ifdef _MSC_VER
+  // std::hash doesn't support structs, but, on windows,
+  // pthread_t is a struct, so combine the hash of the
+  // two members.
+  static size_t hash(const pthread_t& t) {
```
Orvid: This is what I need to relocate for proper hashing support.

Orvid:
Alright, this has been updated to my local version, and also has eliminated the need for the hasher in Hash.h, and renamed pthread_t_init to pthread_zero.

jwatzman:
@sgolemon is there an internal diff for this, sounds like it? What's the diff number, and what's the status?

Orvid:
The publicly imported diff is [here](https://reviews.facebook.net/D46365) I don't know the internal diff number though.

jwatzman:
Yeah, I eventually found it. I didn't realize we had our fancy PR process turned on for folly yet :)

Orvid:
Well, last I checked it wasn't quite on fully, as it wasn't adding the links to the Phabricator reviews to new PRs...

Orvid:
Replacing internally with D2862671.

PR #287, opened 2015-07-24
The core of the Windows port

Orvid:
This is as small as I can make it. I have stripped out every other small change, which have been submitted as 43 separate PRs.
This adds the core files, `CPortability.cpp`, `FilePortability.cpp`, `FilePortability.h`, `SocketPortability.cpp`, `SocketPortability.h`, and `WindowsPortability.h`, which make up the entirety of the portability interface for Windows.
The vast majority of the files this touches are files in which the `#include`s are changing. The only files with modifications other than `#include`s are:
```
CPortability.h
FileUtil.cpp
Makefile.am
SocketAddress.cpp
SocketAddress.h
io/async/AsyncServerSocket.cpp
io/async/AsyncSocket.cpp
io/async/AsyncUDPSocket.cpp
```
All other modifications, except for the new files, are purely changing which headers are included.

Hopefully I've managed to make this significantly more reviewable than it originally was.


folly/CPortability.cpp
```diff
@@ -0,0 +1,563 @@
...
+}
+
+char* asctime_r(const tm* tm, char* buf) {
+  char tmpBuf[64];
+  if (asctime_s(tmpBuf, sizeof(tmpBuf), tm))
+    return NULL;
+  return strcpy(buf, tmpBuf);
+}
+
+void bzero(void* s, size_t n) {
```
ariccio: Since this is Windows-only, perhaps you could annotate `s` as:

```
_Pre_writable_size_( n )
```

folly/CPortability.cpp
```diff
@@ -0,0 +1,563 @@
...
+void __stdcall __alarm_callback_func(HWND, UINT, UINT_PTR, DWORD) {
+  raise(SIGALRM);
+}
+
+unsigned int alarm(unsigned int seconds) {
+  SetTimer(NULL, 0, (UINT)(seconds * 1000), __alarm_callback_func);
+  return 0;
+}
+
+char* asctime_r(const tm* tm, char* buf) {
```
ariccio: Since this is Windows-only, perhaps you could annotate `buf` as:

```
_Out_writes_z_( 64 )
```

folly/CPortability.cpp
```diff
@@ -0,0 +1,563 @@
...
+}
+
+char* ctime_r(const time_t* t, char* buf) {
+  char tmpBuf[64];
+  if (ctime_s(tmpBuf, sizeof(tmpBuf), t))
+    return NULL;
+  return strcpy(buf, tmpBuf);
+}
+
+char* dirname(char* path) {
```
ariccio: Since this is Windows-only, perhaps you could annotate `path` as:

`_Inout_opt_z_`

...and the return value as:
`_Ret_z_`

folly/CPortability.cpp
```diff
@@ -0,0 +1,563 @@
...
+  // Could probably implement dontneed via VirtualAlloc
+  // with the MEM_RESET and MEM_RESET_UNDO flags. 
+  return 0;
+}
+
+size_t malloc_usable_size(void* addr) {
+  return _msize(addr);
+}
+
+void* memmem(const void* haystack, size_t hlen, const void* needle, size_t nlen) {
```
ariccio: Since this is Windows-only, perhaps you could annotate this function like so:

```
_Ret_maybenull_ memmem( _In_reads_bytes_( hlen) const void* haystack, size_t hlen, _In_reads_bytes_( nlen ) const void* needle, size_t nlen )
```

Orvid:
See #235 for the reason why only a couple of places have annotations.

folly/CPortability.cpp
```diff
@@ -0,0 +1,563 @@
...
+  }
+
+  return NULL;
+}
+
+void* memrchr(const void* s, int c, size_t n) {
+  const unsigned char* p = ((const unsigned char*)s) + n;
+  while (p >= (const unsigned char*)s) {
+    if (*p == (unsigned char)c) {
+      return (void*)p;
```
ariccio: Since this is Windows-only, perhaps you could annotate `s` as:

```
_In_reads_bytes_( n )
```

Side note: I assume you're intentionally casting away `const` to preserve the (horrible) behavior of the Linux `memrchr`? If you nonetheless want static analysis (and *not* the compiler) to treat it as `const`, you can annotate the return value as:

```
_Const_
```
...which exists for this very purpose.

ariccio:
Whoops.... I knew this seemed familiar.

I hope I wasn't to annoying, keep up the good work!

folly/SocketPortability.h
```diff
@@ -0,0 +1,148 @@
...
+#endif
+
+#ifdef _MSC_VER
+
+#ifndef __STDC__
+#define __STDC__ 1
+#include <io.h>
+#undef __STDC__
+#else
+#include <io.h>
```
JoelMarcey: Do we need both of these `#include <io.h>`. Our linter wants to make sure...

```
"io.h" is included multiple times. Please remove one of the #includes. To suppress this lint error, add the comment 'nolint' at the end of the include line.
```

folly/SocketPortability.h
```diff
@@ -0,0 +1,148 @@
...
+// While MSVC may not have Unix sockets, Cygwin does.
+#ifdef _MSC_VER
+# define HAVE_UNIX_SOCKETS 0
+#else
+# define HAVE_UNIX_SOCKETS 1
+#endif
+
+#ifdef _MSC_VER
+
+#ifndef __STDC__
```
JoelMarcey: We get this lint warning too

```
Warning: Symbol STDC invalid. A symbol may not begin with two adjacent underscores.
```

folly/SocketPortability.h
```diff
@@ -0,0 +1,148 @@
...
+  int getsockname(int s, struct sockaddr* name, int* namelen);
+  int getsockopt(int s, int level, int optname, char* optval, int* optlen);
+  int getsockopt(int s, int level, int optname, void* optval, int* optlen);
+  int inet_aton(const char *cp, struct in_addr *inp);
+  const char* inet_ntop(int af, const void* src, char* dst, socklen_t size);
+  int listen(int s, int backlog);
+  int poll(pollfd* fds, int nfds, int timeout);
+  int recv(int s, char* buf, int len, int flags);
+  int recv(int s, void* buf, int len, int flags);
+  int recvfrom(int s, char* buf, int len, int flags, struct sockaddr* from, int* fromlen);
```
JoelMarcey: Please make all lines 80 chars or less. So this would be something like:

```
int recvfrom(int s, char* buf, int len, int flags, struct sockaddr* from, 
                    int* fromlen);
```

There are a **bunch** of these in this pull request; I won't enumerate them all.

folly/SocketPortability.h
```diff
@@ -0,0 +1,148 @@
...
+  int shutdown(int s, int how);
+  int socketpair(int domain, int type, int protocol, int sockot[2]);
+
+  // This is the only function that _must_ be referenced via fsp::
+  // because there is no difference in parameter types to overload
+  // on.
+  int socket(int af, int type, int protocol);
+}}
+
+using namespace folly::socket_portability;
```
JoelMarcey: Got this lint error too...

```
Using directive not allowed at top level. Override this lint warning with a /* using override */ comment
```

folly/SocketPortability.cpp
```diff
@@ -0,0 +1,257 @@
+
```
JoelMarcey: Let's add a license here.... you can copy other files.

folly/FilePortability.h
```diff
@@ -0,0 +1,154 @@
...
+
+#ifdef _MSC_VER
+#include <folly/Portability.h>
+
+#ifndef __STDC__
+#define __STDC__ 1
+#include <io.h>
+#undef __STDC__
+#else
+#include <io.h>
```
JoelMarcey: Double `<io.h>` again.

folly/FilePortability.h
```diff
@@ -0,0 +1,154 @@
...
+int setmode(int fh, int md);
+int sopen(char const* fn, int of, int sf, int pm = 0);
+long tell(int fh);
+int truncate(const char* path, off_t len);
+int umask(int md);
+int write(int fh, void const* buf, unsigned int mcc);
+ssize_t writev(int fd, const iovec* iov, int count);
+}}
+
+using namespace folly::file_portability;
```
JoelMarcey: Another `using` at the top level lint error

folly/CPortability.cpp
```diff
@@ -0,0 +1,581 @@
...
+void openlog(const char*, int, int) {}
+void closelog() {}
+void syslog(int, const char*, ...) {}
+
+void __stdcall __alarm_callback_func(HWND, UINT, UINT_PTR, DWORD) {
+  raise(SIGALRM);
+}
+
+unsigned int alarm(unsigned int seconds) {
+  SetTimer(NULL, 0, (UINT)(seconds * 1000), __alarm_callback_func);
```
JoelMarcey: Is there a reason `NULL` can't be `nullptr` -- that's what we prefer. This would go for all newly added `NULL`s.

folly/CPortability.h
```diff
@@ -53,4 +53,305 @@
...
+#ifdef _MSC_VER
+
+#ifndef __STDC__
+#define __STDC__ 1
+#include <io.h>
+#include <direct.h>
+#undef __STDC__
+#else
+#include <io.h>
+#include <direct.h>
```
JoelMarcey: Same duplication of `<io.h>` and this time `<direct.h>`

JoelMarcey:
cc @yfeldblum I made some comments based on the internal D2407585. Please make any more comments here.

folly/SocketPortability.h
```diff
@@ -0,0 +1,148 @@
...
+#include <sys/socket.h>
+#include <sys/un.h>
+
+namespace folly { namespace socket_portability {
+  // Unless you have a case where you would normally have
+  // to reference the function as being explicitly in the
+  // global scope, then you shouldn't be calling these directly.
+  inline int bind(int s, const struct sockaddr* name, int namelen) { return ::bind(s, name, namelen); }
+  inline int connect(int s, const struct sockaddr* name, int namelen) { return ::connect(s, name, namelen); }
+  inline int getpeername(int s, struct sockaddr* name, int* namelen) { return ::getpeername(s, name, namelen); }
```
JoelMarcey: We are getting build errors:

```
folly/SocketPortability.h: In function â€˜int folly::socket_portability::getpeername(int, sockaddr*, int*):
folly/SocketPortability.h:126:109: error: invalid conversion from int* to socklen_t* {aka unsigned int*} [-fpermissive]
include/sys/socket.h:142:12: error:   initializing argument 3 of int getpeername(int, sockaddr*, socklen_t*) [-fpermissive]
```

This is happening in a bunch of places. Like most of the functions below.

Orvid: Looks like I messed up when I typed up the posix side of the API.... I'll go through and make sure they all get set to socklen_t where they should be.

folly/SocketPortability.h
Orvid: For the wrappers to work correctly, as I have to implement `close` in a way that also supports socket handles, I need to make sure the defines that remap them to `_*` versions are not used. Those defines are disabled by defining `__STDC__`, however, if I were to define that for everything, I would need a whole lot more wrappers, because that define would disable just about every standard C function if it's not prefixed with an underscore. This means that I have to define it for a select few files. If `__STDC__` is already defined however, I don't need to redefine it. Having the include in multiple places is shorter than wrapping the def/udef pair.

folly/SocketPortability.h
Orvid: I assume there's some way to disable this one? I need that symbol to be named that way.

folly/SocketPortability.h
Orvid: Most of the folly port was written before I got Visual Studio setup to show me a line at 80 columns, so there's going to be a lot of this. I'll go back through and deal with them.

folly/SocketPortability.h
Orvid: Alright, will add the ignores, because that using is needed for this to play nice. Without that using, every single call to any of the socket portability functions would have to be prefixed with `fsp::`. In fact, that is how I originally wrote it, before getting tired of having to change everything and realizing I can just get the compiler to resolve to the correct overload of the function.

folly/SocketPortability.cpp
Orvid: Not sure how I missed adding the license, woops.

folly/CPortability.cpp
Orvid: All of the portability headers were originally written to be header-only, which meant that the CPortability functions needed to only use C, however I split the bodies out when I started the second port, because otherwise I ended up having to do a recompile of most of HHVM if I changed the body of one of these functions.

As this is now able to use C++, I'll replace these with `nullptr`.

Orvid:
Moved internally to D2866379.

PR #294, opened 2015-08-03
Add support for recent libevent versions.

brooksbp:
Add a workaround to support recent versions of libevent.

Here is the relevant libevent commit that introduced the change: https://github.com/libevent/libevent/commit/cba59e53253bb396f192e40a002c6dd9835df51c#diff-4d5f6c7a554e6cab25684fe7194e7ba5

### DIFF
--- a/folly/io/async/EventUtil.h
+++ b/folly/io/async/EventUtil.h
@@ -22,6 +22,14 @@
 
 #include <event.h>  // libevent
 
+#if LIBEVENT_VERSION_NUMBER > 0x02010101
+#define ev_pri ev_evcallback.evcb_pri
+#define ev_flags ev_evcallback.evcb_flags
+#define ev_closure ev_evcallback.evcb_closure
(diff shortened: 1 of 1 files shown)


sgolemon:
Pulled for review as D2306034

Orvid:
Hrmm... I build on windows against a 3 month old version of libevent pulled from the master branch of the github repo, and I haven't had to make and changes like this...

Orvid:
Then again, it's perfectly possible that I don't use the header this changes in the first place.

brooksbp:
From what I can tell, `event-internal.h` is not included from any of the public headers.  So, I take this to be a breaking change that requires all code using libevent to do the equivalent of the `#define`s or modify the code directly.  I am building on Linux.

JoelMarcey:
https://github.com/facebook/folly/commit/99eb3f45bba336163f28cefe5d549731ea4a886b landed that should support this now.

cc: @yfeldblum

PR #295, opened 2015-08-06
Remove `unsigned char v = static_cast<unsigned char>(v);`

Orvid:
MSVC spotted this very suspicious line when I was running it with /analyze, and complained about using an unitialized variable. This variable isn't used anywhere in the function, and is initializing itself with itself. I have no idea how this compiled in the first place, but apparently it does.
This removes the line in question.


ariccio:
What the hell? My guess (based on keyboard proximity) is that someone meant `unsigned char c = static_cast<unsigned char>( v );`.

ariccio:
(or something like it)

PR #296, opened 2015-08-06
Fixes for upstream

doctaweeks:
Fixes compiling crc32c function on older hardware. (I believe this problem was introduced by 7ec1fc0e0e27ed56c1d18d33af9711a17e9ec750).

Also ensures is in final library since the API suggests it should be. This should correct a number of build problems for other projects like wdt and hhvm when they are compiled using a shared folly library rather than borrowing sources at compile time.

### DIFF
--- a/folly/Checksum.cpp
+++ b/folly/Checksum.cpp
@@ -29,10 +29,11 @@ namespace detail {
   #define __has_builtin(x) 0
 #endif
 
-#if (__has_builtin(__builtin_ia32_crc32qi) && \
+#if __SSE4_2__ && \
+    ((__has_builtin(__builtin_ia32_crc32qi) && \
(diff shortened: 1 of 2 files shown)


folly/Checksum.cpp
```diff
@@ -29,10 +29,11 @@ namespace detail {
   #define __has_builtin(x) 0
 #endif
 
-#if (__has_builtin(__builtin_ia32_crc32qi) && \
+#if __SSE4_2__ && \
```
Orvid: As a random note, GCC does support [Function Multiversioning](https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html) which would allow this check to be done at runtime rather than compile-time.

PR #297, opened 2015-08-07
Adjust Format-inl.h to work with MSVC

Orvid:
Originally #247, which was closed before being merged, this doesn't change the behavior of the formatting function now.


Orvid:
Ekk, another place where a change was excluded that is required for MSVC....  The change to the constraint of TryFormatValue is required.

PR #299, opened 2015-08-12
Port Folly to PPC64

lbianc:
This patch adds PPC64 specific code in order to support this architecture and be able to be used by HHVM.

### DIFF
--- a/folly/detail/MemoryIdler.cpp
+++ b/folly/detail/MemoryIdler.cpp
@@ -83,7 +83,7 @@ void MemoryIdler::flushLocalMallocCaches() {
 // Stack madvise isn't Linux or glibc specific, but the system calls
 // and arithmetic (and bug compatibility) are not portable.  The set of
 // platforms could be increased if it was useful.
-#if FOLLY_X64 && defined(_GNU_SOURCE) && defined(__linux__)
+#if (FOLLY_X64 || FOLLY_PPC64 ) && defined(_GNU_SOURCE) && defined(__linux__)
 
 static const size_t s_pageSize = sysconf(_SC_PAGESIZE);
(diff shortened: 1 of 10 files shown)


folly/detail/MemoryIdler.cpp
```diff
@@ -83,7 +83,7 @@ void MemoryIdler::flushLocalMallocCaches() {
 // Stack madvise isn't Linux or glibc specific, but the system calls
 // and arithmetic (and bug compatibility) are not portable.  The set of
 // platforms could be increased if it was useful.
-#if FOLLY_X64 && defined(_GNU_SOURCE) && defined(__linux__)
+#if (FOLLY_X64 || FOLLY_PPC64 ) && defined(_GNU_SOURCE) && defined(__linux__)
```
Orvid: Have an extra space here

cole14:
Unless you're using ppc64el, you'll also need to port fbstring_core to work correctly for big endian as well: #244

Orvid:
This is using `ppc64le` see [here](https://github.com/facebook/folly/pull/299/files#diff-0aeb0bf602d25a066eb2233e06c4a981R408).

lbianc:
Thanks @Orvid, just to confirm, the target of this change is ppc64le.

leitao:
@Orvid , we are targeting ppc64el only for the POWER. In fact, ppc64el is the new POWER ABI, also called ABI v2, that is little endian mode. This ABI is our major focus at the moment, while the ABI V1 (Big Endian) is out of focus at the moment.

gut:
Hello, how is this PR going? Is there anything blocking the merge or is the extra space the only issue?

PR #301, opened 2015-08-16
Add a compatibility shim for working with libevent on MSVC

Orvid:
MSVC builds of libevent expect `evutil_socket_t` to be `HANDLE` values, but Folly, HHVM, and Thrift all use them as file descriptors.
This adds a shim for the only method that needs it, `event_set`, and includes it where it's needed in Folly.
The shim is only enabled under MSVC, but it's safe to include the header from any platform.

### DIFF
--- a/folly/EventPortability.cpp
+++ b/folly/EventPortability.cpp
@@ -1,0 +1,51 @@
+/*
+* Copyright 2015 Facebook, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifdef _MSC_VER
+#include <folly/EventPortability.h>
+
+#include <fcntl.h>
+#include <io.h>
+#include <functional>
+
+namespace folly { namespace event_portability {
+
+// The indirect manner of casting to/from evutil_socket_t is done so
+// that we can still compile if we've adjusted evutil_socket_t to be
+// defined as a structure to break any places that are incorrectly
+// using the API. (ABI-wise, they are actually the exact same)
+
+void event_set(struct event* e, evutil_socket_t fd, short s, void(*f)(int, short, void*), void* b) {
+  // We need to dynamically wrap a function pointer, so we get this fun :(
+  auto fp = std::function<void(evutil_socket_t, short, void*)>([&](evutil_socket_t a, short b, void* c) {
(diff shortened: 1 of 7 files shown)


folly/EventPortability.cpp
```diff
@@ -0,0 +1,51 @@
+/*
```
JoelMarcey: From @yfeldblum any reason not to put the portability layer under `folly/io/async` instead of up-leveled here?

folly/EventPortability.cpp
```diff
@@ -0,0 +1,51 @@
...
+// using the API. (ABI-wise, they are actually the exact same)
+
+void event_set(struct event* e, evutil_socket_t fd, short s, void(*f)(int, short, void*), void* b) {
+  // We need to dynamically wrap a function pointer, so we get this fun :(
+  auto fp = std::function<void(evutil_socket_t, short, void*)>([&](evutil_socket_t a, short b, void* c) {
+    f(_open_osfhandle(*(intptr_t*)&a, O_RDWR | O_BINARY), b, c);
+  });
+  auto fp2 = fp.target<void(evutil_socket_t, short, void*)>();
+  ::event_set(e, fd, s, fp2, b);
+}
```
JoelMarcey: From @yfeldblum  How does all this work when there are captures?

folly/EventPortability.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+void event_set(struct event* e, int fd, short s, void(*f)(int, short, void*), void* b) {
+  intptr_t fh;
+  if (fd == -1)
+    fh = (intptr_t)INVALID_HANDLE_VALUE;
+  else
+    fh = _get_osfhandle(fd);
```
JoelMarcey: From @yfeldblum How about this instead?

```
intptr_t fh = -1 == fd
  ? INVALID_HANDLE_VALUE
  : _get_osfhandle(fd);
```

folly/EventPortability.cpp
```diff
@@ -0,0 +1,51 @@
...
+  event_set(e, *(evutil_socket_t*)&fh, s, f, b);
```
JoelMarcey: From @yfeldblum  

```
static_assert(sizeof(evutil_socket_t) == sizeof(fh), "mismatched sizes");
::event_set(e, reinterpret_cast<evutil_socket_t>(fh), s, f, b);
```

folly/EventPortability.cpp
Orvid: Only reason I put it in the root is that it's where the other portability headers will be from #287.

folly/EventPortability.cpp
Orvid: I'd ask that of whatever inside std::function is allowing me to get a `void (*)(evutil_socket_t, short, void*)` out of it. All I know is that it does run correctly.

folly/EventPortability.cpp
Orvid: I can certainly do that. It's only written this way because I dislike long ternary statements. Also, the cast is needed, as `INVALID_HANDLE_VALUE` is a `HANDLE`, which doesn't implicitly convert to an `intptr_t`.

jwatzman:
@JoelMarcey is there an internal diff for this, I assume so? What's the status?

jwatzman:
Ah nm I found it

Orvid:
Moved internally to D2874655.

PR #302, opened 2015-08-16
Don't use weird format specifiers

Orvid:
MSVC doesn't like it when you use `%1$s`, and actually emits a warning at compile time, so pass the second arugment twice instead.

### DIFF
--- a/folly/experimental/NestedCommandLineApp.cpp
+++ b/folly/experimental/NestedCommandLineApp.cpp
@@ -130,8 +130,8 @@ void NestedCommandLineApp::displayHelp(
     // Help for a given command
     auto& p = findCommand(args.front());
     if (p.first != args.front()) {
-      printf("`%1$s' is an alias for `%2$s'; showing help for `%2$s'\n",
-             args.front().c_str(), p.first.c_str());
+      printf("`%s' is an alias for `%s'; showing help for `%s'\n",
+             args.front().c_str(), p.first.c_str(), p.first.c_str());
     }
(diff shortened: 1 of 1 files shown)


folly/experimental/NestedCommandLineApp.cpp
```diff
@@ -130,8 +130,8 @@ void NestedCommandLineApp::displayHelp(
     // Help for a given command
     auto& p = findCommand(args.front());
     if (p.first != args.front()) {
-      printf("`%1$s' is an alias for `%2$s'; showing help for `%2$s'\n",
-             args.front().c_str(), p.first.c_str());
+      printf("`%s' is an alias for `%s'; showing help for `%s'\n",
```
Orvid: As a side note, what's up with the quotes in this? Why are the opening quotes backticks, but the closing quotes just single quotes?

fugalh: That's just GNU-style quotes (
https://www.gnu.org/prep/standards/html_node/Quote-Characters.html) - an
ill-advised tradition from a bygone era (in my personal opinion), but still
pops up perennially. I'm sure it was the preference of whoever wrote this
code. (An emacs user no doubt, not that there's anything wrong with that
but there seems to be a correlation.)

Orvid:
As a note, I'm pretty sure the runtime is actually fine with the current format specifiers, it's just that VS 2015 added compile-time checking of the format specifier strings, and I suspect it missed this construct, which means that it complains :(

PR #305, opened 2015-08-25
Add and implement a couple of fallback defines for MSVC - What was missed

Orvid:
When #264 was merged, some confusion appears to have occurred over whether the defines were needed for the final release of VS 2015. They are needed, so this adds them.

This also adds the change that was excluded when #297 was merged.

### DIFF
--- a/folly/AtomicLinkedList.h
+++ b/folly/AtomicLinkedList.h
@@ -19,6 +19,7 @@
 
 #include <atomic>
 #include <cassert>
+#include <folly/Portability.h>
 
 namespace folly {
 
@@ -67,7 +68,11 @@ class AtomicLinkedList {
   }
 
   bool empty() const {
+#ifdef MSVC_NO_NONVOID_ATOMIC_IF
+    return head_.load() == nullptr;
+#else
     return head_ == nullptr;
+#endif
   }
 
   /**
--- a/folly/AtomicLinkedList.h
+++ b/folly/AtomicLinkedList.h
@@ -19,7 +19,6 @@
 
 #include <atomic>
 #include <cassert>
-#include <folly/Portability.h>
 
 namespace folly {
 
@@ -68,11 +67,7 @@ class AtomicLinkedList {
   }
 
   bool empty() const {
-#ifdef MSVC_NO_NONVOID_ATOMIC_IF
     return head_.load() == nullptr;
-#else
-    return head_ == nullptr;
-#endif
   }
 
   /**
--- a/folly/Format-inl.h
+++ b/folly/Format-inl.h
@@ -774,9 +774,7 @@ class TryFormatValue {
 
 template <class T>
 class TryFormatValue<
-  T,
-  typename std::enable_if<
-      0 < sizeof(FormatValue<typename std::decay<T>::type>)>::type>
+  T, typename FormatValue<typename std::decay<T>::type>::type>
   {
  public:
   template <class FormatCallback>
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -334,6 +334,25 @@ inline size_t malloc_usable_size(void* ptr) {
 # define FOLLY_HAS_RTTI 1
 #endif
 
+#if defined(_MSC_VER) && _MSC_FULL_VER <= 190023026 // 2015 RTM or below
+// MSVC2015's initial release produces errors if you try
+// to do `if (std::atomic<char*>)`, while `if (std::atomic<void*>)`
+// works just fine.
+// Bug Report: https://connect.microsoft.com/VisualStudio/feedback/details/1464842
+# define MSVC_NO_NONVOID_ATOMIC_IF 1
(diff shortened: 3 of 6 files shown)


folly/AtomicLinkedList.h
```diff
@@ -67,7 +68,11 @@ class AtomicLinkedList {
   }
 
   bool empty() const {
+#ifdef MSVC_NO_NONVOID_ATOMIC_IF
+    return head_.load() == nullptr;
+#else
     return head_ == nullptr;
+#endif
```
JoelMarcey: If we're supporting broken compilers, just call .load() explicitly for all compilers. And add a comment explaining why.

folly/Format-inl.h
```diff
@@ -774,9 +774,7 @@ class TryFormatValue {
 
 template <class T>
 class TryFormatValue<
-  T,
-  typename std::enable_if<
-      0 < sizeof(FormatValue<typename std::decay<T>::type>)>::type>
+  T, typename FormatValue<typename std::decay<T>::type>::type>
```
JoelMarcey: Why is this necessary?

folly/Portability.h
```diff
@@ -334,6 +334,17 @@ inline size_t malloc_usable_size(void* ptr) {
...
 
+#if defined(_MSC_VER) && _MSC_FULL_VER <= 190023026 // 2015 RTM or below
+// MSVC2015's initial release produces errors if you try
+// to do `if (std::atomic<char*>)`, while `if (std::atomic<void*>)`
+// works just fine.
+// Bug Report: https://connect.microsoft.com/VisualStudio/feedback/details/1464842
+# define MSVC_NO_NONVOID_ATOMIC_IF 1
+// 2015 RTM doesn't support in-class initialization of static constexpr members.
+// Attempting to do this produces an error informing you that this is not yet supported.
+# define MSVC_NO_STATIC_INCLASS_CONSTEXPR_INITIALIZATION 1
```
JoelMarcey: Are there no newer MSVC2015 compilers? Are we going to support broken early versions of fixed compilers?

folly/Singleton.cpp
```diff
@@ -22,7 +22,11 @@ namespace folly {
 
 namespace detail {
 
+#ifdef MSVC_NO_STATIC_INCLASS_CONSTEXPR_INITIALIZATION
+const std::chrono::seconds SingletonHolderBase::kDestroyWaitTime{5};
+#else
 constexpr std::chrono::seconds SingletonHolderBase::kDestroyWaitTime;
+#endif
```
JoelMarcey: Are there no newer MSVC2015 compilers? Are we going to support broken early versions of fixed compilers?

folly/Singleton.h
```diff
@@ -200,7 +200,11 @@ class SingletonHolderBase {
   virtual void destroyInstance() = 0;
 
  protected:
+#ifdef MSVC_NO_STATIC_INCLASS_CONSTEXPR_INITIALIZATION
+  static const std::chrono::seconds kDestroyWaitTime;
+#else
   static constexpr std::chrono::seconds kDestroyWaitTime{5};
+#endif
```
JoelMarcey: If we're supporting broken compilers, just make this `const` rather than `constexpr` for all compilers. And add a comment explaining why.

Or keep this as `constexpr` but move it out of the class and into the namespace, renaming it to

`kSingletonHolderBaseDestroyWaitTime`

And add a comment why it can't go into the class.

folly/Portability.h
```diff
@@ -334,6 +334,17 @@ inline size_t malloc_usable_size(void* ptr) {
...
 #endif
 
+#if defined(_MSC_VER) && _MSC_FULL_VER <= 190023026 // 2015 RTM or below
+// MSVC2015's initial release produces errors if you try
+// to do `if (std::atomic<char*>)`, while `if (std::atomic<void*>)`
+// works just fine.
+// Bug Report: https://connect.microsoft.com/VisualStudio/feedback/details/1464842
+# define MSVC_NO_NONVOID_ATOMIC_IF 1
+// 2015 RTM doesn't support in-class initialization of static constexpr members.
+// Attempting to do this produces an error informing you that this is not yet supported.
```
JoelMarcey: Lint - make this line less than 80 chars

JoelMarcey:
@Orvid I add some comments from a reviewer internally.

folly/Portability.h
Orvid: There are no newer versions currently.

Orvid: As soon as a newer version is released that fixes these, I fully intend to remove the workarounds.

folly/Format-inl.h
Orvid: I'm not certain,  but I believe this is due to MSVC not yet supporting expression SFINAE. Whatever the actual reason MSVC doesn't like the sizeof on an invalid type.

Orvid:
Moved internally to D2901301.

PR #307, opened 2015-08-25
The last few fixes for MSVC

Orvid:
I didn't want to deal with creating separate PRs for each of these, so this is one still small, PR instead.

And no, I can't get rid of T2 in Future.h, nor can I check T2 against Unit. I don't know why, but MSVC just doesn't like it.

### DIFF
--- a/folly/futures/Future.h
+++ b/folly/futures/Future.h
@@ -62,7 +62,7 @@ class Future {
 
   template <class T2 = T, typename =
             typename std::enable_if<
-              std::is_same<Unit, T2>::value>::type>
+              std::is_same<Unit, T>::value>::type>
   Future();
 
   ~Future();
--- a/folly/json.cpp
+++ b/folly/json.cpp
@@ -663,7 +663,7 @@ void escapeString(StringPiece input,
         // checking that utf8 encodings are valid
         char32_t v = decodeUtf8(q, e, opts.skip_invalid_utf8);
         if (opts.skip_invalid_utf8 && v == U'\ufffd') {
-          out.append("\ufffd");
+          out.append(u8"\ufffd");
           p = q;
           continue;
         }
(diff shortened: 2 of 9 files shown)


folly/futures/Future.h
```diff
@@ -62,7 +62,7 @@ class Future {
 
   template <class T2 = T, typename =
             typename std::enable_if<
-              std::is_same<Unit, T2>::value>::type>
+              std::is_same<Unit, T>::value>::type>
```
JoelMarcey: Why are you doing this?

folly/json.cpp
```diff
@@ -663,7 +663,7 @@ void escapeString(StringPiece input,
         // checking that utf8 encodings are valid
         char32_t v = decodeUtf8(q, e, opts.skip_invalid_utf8);
         if (opts.skip_invalid_utf8 && v == U'\ufffd') {
-          out.append("\ufffd");
+          out.append(u8"\ufffd");
```
JoelMarcey: What is this?

JoelMarcey:
A couple of questions from @yfeldblum inline with code

Note: We are going to start our automated pull into external phabricator starting with newer pull requests, so code comments, etc. will be on external diffs instead of me being a conduit.

folly/json.cpp
Orvid: It's a utf8 string literal, added as part of c++11, and the only way to get msvc to allow codepoints over ff in a narrow string literal.

folly/futures/Future.h
Orvid: Because MSVC is dumb and this condition doesn't work if it's T2.

Orvid:
Yep, I saw the diffs get created in phabricator a while ago. I assume the links to them will be added to the PRs once the pull os complete?

JoelMarcey:
@Orvid For ones that I have already pulled manually internally, we are just going to use the *old* process. For pull requests above #309, we will start using the actually phabricator importer, etc.

Orvid:
Moved internally to D2901301.

PR #308, opened 2015-08-27
Support setThreadName under MSVC

Orvid:
Wangle tries to set a thread's name by passing an `std::thread`'s handle, which, under MSVC, is not a `pthread_t`, so we need to add an additional overload to return that we couldn't set it.
I originally tried doing this as a template enabled only if `std::thread::native_handle_type` wasn't `pthread_t`, but I just couldn't quite get it to work right, so I gave up and just special cased MSVC.

### DIFF
--- a/folly/ThreadName.h
+++ b/folly/ThreadName.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <thread>
 #include <pthread.h>
 #include <folly/Range.h>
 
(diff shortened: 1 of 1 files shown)


Orvid:
A different route to make this happen was taken in https://github.com/facebook/folly/commit/4463e2301ce61477bbe4c0982d699158db33637f.

PR #309, opened 2015-08-27
MSVC has std::underlying_type so use it when converting an enum to a string

Orvid:
And because the `<` operator causes issues with them, at least, in the way this API is used by proxygen's original cpp implementation.


JoelMarcey:
We actually have an internal diff landing soon that drops support for GCC < 4.8 in `folly/Conv.h`. That will supercede this pull request. It should land soon (maybe today).

cc: @yfeldblum

PR #311, opened 2015-08-29
Declare missing environ variable

cdolan:
Declaring `extern char** environ` in `experimental/TestUtil.cpp` fixes a compilation issue on OS X 10.10.

### DIFF
--- a/folly/experimental/TestUtil.cpp
+++ b/folly/experimental/TestUtil.cpp
@@ -28,6 +28,10 @@
 #include <folly/FileUtil.h>
 #include <folly/String.h>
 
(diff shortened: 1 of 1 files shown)


folly/experimental/TestUtil.cpp
```diff
@@ -28,6 +28,8 @@
 #include <folly/FileUtil.h>
 #include <folly/String.h>
 
+extern char** environ;
```
Orvid: This needs to be wrapped in a `#ifndef _MSC_VER`, because MSVC defines this differently, and doing this will break it.

PR #318, opened 2015-09-14
Adjust AsyncServerSocket to not use getsockname before it's connected

Orvid:
Winsock doesn't like it when you try to call `getsockname` on a socket that hasn't yet been connected or bound, because that socket doesn't have a name yet.
This only occurred because we were trying to get the family of the socket.
To solve this, I just passed the family in from the parent methods that already knew what the family was.


Orvid:
Replaced internally with D2861921.

PR #319, opened 2015-09-14
Use event_pending to check if an event is pending or registered under MSVC

Orvid:
I've only made this change for MSVC because I don't know if it will work the same way as it does under MSVC.

Under MSVC at least, `isEventRegistered` was returning `true` when it shouldn't have been, prompting me to ask in the IRC channel for libevent, and, after asking a few times over a few days, I finally got a response that said that `event_pending` is the correct way to be doing this, so that's what MSVC will now use.

I do know that this does now work correctly under MSVC, but I'm not sure about anything else.

### DIFF
--- a/folly/io/async/EventHandler.cpp
+++ b/folly/io/async/EventHandler.cpp
@@ -169,12 +169,16 @@ void EventHandler::setEventBase(EventBase* eventBase) {
 }
 
 bool EventHandler::isPending() const {
+#ifdef _MSC_VER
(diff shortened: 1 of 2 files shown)


Orvid:
I'm pretty sure the review of this is intended to be done at [D46839](https://reviews.facebook.net/D46839), but the bot hasn't commented about it being imported.

Orvid:
Moved internally.

PR #323, opened 2015-09-25
Fix incorrect variable in dynamic examples

lifei:
The examples does not work for there is an incorrect variable in the code.


fredemmott:
Sorry for the delay, our systems had unicode issues. Should have this in soon.

FB ref: D2518882

PR #325, opened 2015-10-01
Compile on MSVC 2015

korovkin:


korovkin:
@Orvid this is the combined list of commits (yours and mine) that i have used to compile this for MSVC 2015. 
(what we have been discussing during the last few days) 

the PR seems a bit dirty to be merged in. 
but i am not actually sure it needs to be merged, as this compiles only a fraction of the features folly offers for MSVC. 

if you are interested in merging some changes, i don't mind cleaning this app a bit, as soon as you share some feedback on my changes. 

cheers.

fredemmott:
We'll be merging changes individually, as needed for the next update of MSVC.

Orvid:
@korovkin I did have a few comments that I made on the commits themselves in your repo. Other than a couple things, such as the `NOMINMAX` define are reasonable to have (in this particular case, the `NOMINMAX` macro is being defined in HHVM's config step, as there are other places that try to include the windows headers) in Folly itself to allow building under MSVC outside of the HHVM environment. Most of them however shouldn't actually be needed. You are for instance, trying to build without pthreads, which I have no intention of supporting.

korovkin:
@Orvid i have no intensions pulling that into folly master, i think these changes are too dirty to be used. 

just wanted to share what we did to get it compile, so hopefully it will be useful to you / someone else. 

the reason i didn't include pthread yet is lack of need for it on WinMo which is where we want this to be compiled & used.

korovkin:
@Orvid thanks a bunch for the help and taking a look. 

please let me ping me if you need help or some other parts of folly are available for MSVC. 

cheers!

PR #326, opened 2015-10-04
Change memory_order_consume to relaxed in ProducerConsumerQueue

Hincoin:
A few of the functions within ProducerConsumerQueue use `std::memory_order_consume` to load from shared variables, but the consume guarantees aren't used since there's no earlier data dependencies on these variables that need to be propagated. Changing these loads to `std::memory_order_relaxed` offers the same guarantees that are needed for these functions and, under current compiler support, avoids the promotion to `std::memory_order_acquire`. 

Relevant post on the [Google group](https://groups.google.com/d/msg/facebook-folly/eGb5gGudGtc/nmBNLFVaJQAJ).


fredemmott:
Changes were requested here - unsure why the bot didn't comment with a phabricator link: https://reviews.facebook.net/D48141

PR #337, opened 2015-10-22
Remove GroupVarint32 tables generator dependency on x86 platform

fabioperez:
This PR is necessary for both Folly and HHVM to compile on PPC64 (https://github.com/PPC64/hhvm) and other platforms that are not compatible with SSE instructions.


fabioperez:
Please refer to https://github.com/facebook/folly/pull/339.

PR #339, opened 2015-10-27
Enable GroupVarint on PPC64

fabioperez:
This PR is necessary for both Folly and HHVM to compile on PPC64 (https://github.com/PPC64/hhvm) and other platforms that are not compatible with SSE instructions.

It also removes GroupVarint32 tables generator dependency on x86 platform.

### DIFF
--- a/folly/GroupVarint.h
+++ b/folly/GroupVarint.h
@@ -23,7 +23,7 @@
 
 #include <folly/Portability.h>
 
-#if FOLLY_X64 || defined(__i386__)
+#if FOLLY_X64 || defined(__i386__) || FOLLY_PPC64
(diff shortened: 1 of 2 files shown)


gut:
Why is GroupVarint restricted to X64 only? Except for SSE (which is protected by another macro), I don't see a reason besides little endian encoding for using that code, right?

gut:
Hello, more than a month elapsed and no feedback on this pull request. Do you need any clarification?

gut:
6 weeks. Please support us as the HHVM port will not be sucessfull on PPC64 if this is not handled in Folly.

fredemmott:
Not sure what happened to the phabricator sync, this is https://reviews.facebook.net/D49521

PR #347, opened 2015-12-19
Implements make_unique_inherit for use with polymorphism

alexst07:
In some cases is useful to use unique_ptr with polymorphism. Using make_unique_inherit is possible uses unique_ptr to do in a modern way operations like A *b = new B(); where B is derived from A.

### DIFF
--- a/folly/Memory.h
+++ b/folly/Memory.h
@@ -65,6 +65,15 @@ make_unique(Args&&...) = delete;
 
 #endif
 
+// Allows 'make_unique' for the inherited class to use polymorphism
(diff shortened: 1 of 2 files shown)


nbronson:
Thanks for your pull request. I don't think this is needed, because there is already a move constructor of std::unique_ptr that allows this pattern to work whenever std::unique_ptr<B>::pointer is convertible to std::unique_ptr<A>::pointer. That move constructor lets you write

`std::unique_ptr<A> a = folly::make_unique<B>(args);`

PR #349, opened 2015-12-29
Use DCHECK to warn fallback implementation in qfind_first_byte_of_sse

skyan:
qfind_first_byte_of  calls qfind_first_byte_of_sse42 if CpuId().sse42() return true, but qfind_first_byte_of_sse42 checks error if FOLLY_SSE_PREREQ(4, 2) is false.

So if gcc turn off msse4.2 by default, but CPU supports SSE 4.2 actually, qfind_first_byte_of will throw exception.

### DIFF
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -1003,11 +1003,10 @@ namespace detail {
 
 inline size_t qfind_first_byte_of(const StringPiece haystack,
                                   const StringPiece needles) {
-#if FOLLY_SSE_PREREQ(4, 2)
-  return qfind_first_byte_of_sse42(haystack, needles);
-#else
-  return qfind_first_byte_of_nosse(haystack, needles);
-#endif
+  static auto const qfind_first_byte_of_fn =
+    folly::CpuId().sse42() ? qfind_first_byte_of_sse42
+                           : qfind_first_byte_of_nosse;
+  return qfind_first_byte_of_fn(haystack, needles);
 }
 
 } // namespace detail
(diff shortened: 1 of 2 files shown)


Orvid:
I don't like this being done at compile-time. The way it's setup now only does the check for SSE 4.2 once. I will admit however that this could be done more efficiently, because the initialization is both guarded by a mutex and is checked every time the function is called. Ideally the selection of which version of the function to call would be done once on startup.

As an alternate, and more maintainable way to do this, you could adjust your local copy of `.sse42` to be `constexpr` and always return `true`. That would allow the compiler to eliminate other places that checks were being done at runtime.

skyan:
Thanks for comments. My main purpose is to fix aborting issue when calling `qfind_first_byte_of` under platform where compiler turn off sse4.2 but `CpuId().sse42()` return true. 

In fact, the root cause is in function `qfind_first_byte_of_sse42`, it will call `CHECK(false)` if `FOLLY_SSE_PREREQ(4, 2)` is false, and it will exit the program. But I think it should not abort and could silently choose fallback implementation. 

So maybe the better fix is to eliminate `CHECK(false)` in `qfind_first_byte_of_sse42` or use `CHECK(folly::CpuId().sse42())` instead?

sangechen:
I fix this by  `./configure CXX='g++ -msse4'`

ramrengaswamy:
Agree with @skyan 
Not sure why a `CHECK(false)` is necessary in the implementation of `qfind_first_byte_of_sse42`

Orvid:
We ended up going with a fix in https://github.com/facebook/folly/commit/5c3799297109076c31ee801f00312cfaa0bbc605 that just removes the `CHECK` altogether.

Issue #348, opened 2015-12-28
Commit 2a0cb937 breaks compilation on OS X Yosemite


cleeland:
OS X Yosemite doesn't seem to know about __NR_gettid, but does know about SYS_gettid.

Orvid:
Already fixed.

PR #350, opened 2015-12-30
Fix issue #348: OSX doesn't know about __NR_gettid.

cleeland:
This is one way to resolve issue #348.  It may not be the best way, but I can confirm that it works on OSX Yosemite.

### DIFF
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -63,6 +63,12 @@
   #endif
 #endif
 
+#if defined(__APPLE__)
+  #if !defined(__NR_gettid)
+    #define __NR_gettid SYS_gettid
+  #endif
+#endif
+
 // compiler specific attribute translation
 // msvc should come first, so if clang is in msvc mode it gets the right defines
 


elbamos:
I tried this, and it works.

cleeland:
@elbamos did you try this change on yosemite (10.10) or some other version of OSX?

elbamos:
@cleeland el capitan.

cleeland:
Good to know that (a) it's not a yosemite-only problem and (b) the change
works on the two most likely OSX versions around.

I presume iOS will be in the same situation, though I suppose that somebody
should check it.

On Sun, Jan 3, 2016 at 1:41 AM, elbamos <notifications@github.com> wrote:

> @cleeland <https://github.com/cleeland> el capitan.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/folly/pull/350#issuecomment-168474811>.
>



-- 
Chris Cleeland

Orvid:
This became the [sys/syscall.h portability header](https://github.com/facebook/folly/blob/b1e0c2c8be1d8355617929df0acf618af30dc09c/folly/portability/Syscall.h), so this is no longer needed.

PR #351, opened 2016-01-04
Allow clients of FBVector to customize the growth strategy

gmbeard:
As per the suggestion in the comments of `FBVector.h`, I've added the ability to customize the growth strategy using the `fbvector`'s allocator. If the allocator used implements a non-static member function of type `size_type calculate_capacity(size_type) const` then this will be called in every case where the `fbvector` instance needs to grow. E.g. to use an `N+1` strategy the allocator would look something like...

``` c++
template<typename T>
struct growth_policy_allocator
{
  typedef T value_type;

  T* allocate(size_t n)
  {
    ...
  }

  void deallocate(void *p, size_t n)
  {
    ...
  }

  size_t calculate_capacity(size_t current) const // const is required
  {
    return current + 1;
  }
};
```

If the allocate doesn't implement `calculate_capacity` then `fbvector` falls back to its default growth strategy.

I've also added a simple test case for this feature. Compiles and passes on Ubuntu 14.04

### DIFF
--- a/folly/FBVector.h
+++ b/folly/FBVector.h
@@ -1133,7 +1133,7 @@ private:
   // std::vector implements a similar function with a different growth
   //  strategy: empty() ? 1 : capacity() * 2.
   //
-  // fbvector grows differently on two counts:
+  // fbvector grows differently on three counts:
   //
   // (1) initial size
   //     Instead of grwoing to size 1 from empty, and fbvector allocates at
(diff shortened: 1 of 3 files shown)


thedrow:
Shame this wasn't merged. Anything I can do about it?

Orvid:
It really just needs to be rebased; Our tooling won't let me pull in a revision for review internally if the revision has conflicts with trunk :(

gmbeard:
I've rebased and resolved the conflicts in a separate branch in my fork but I no longer have access to the original environment that I made this change in (I'm on MacOS now). I'll have to do some setup and run the tests when I have some spare time... Unless it's ok for me to just take a punt and update the PR's branch?

PR #356, opened 2016-01-16
Copyright, 2015 > 2016, for files A-D

siddharthasharma:
Updated copyright notices in the source from 2015 to 2016, for all files beginning from A to D.

### DIFF
--- a/folly/ApplyTuple.h
+++ b/folly/ApplyTuple.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Facebook, Inc.
+ * Copyright 2016 Facebook, Inc.
  *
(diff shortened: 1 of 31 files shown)


lbrandy:
Unfortunately, I think someone might have beaten you to this internally. I'll double check.

siddharthasharma:
@lbrandy, what do you intend to state? I updated Haxl's notices last time and this time it's folly. Do double check, and update me. And, if you want A-Z, give me a day. L/F to response.

Orvid:
@lbrandy I think the internal diff already landed and missed Folly completely, due to its use of a single year rather than range of years.

Orvid:
Internal diff landed. See 321542683a01c3f334047531e9b487f047129775.

PR #359, opened 2016-01-29
Typo

JVillella:
We should be [showcasing the almost-backronym that the word "folly" is supposed to be](https://github.com/facebook/folly/pull/110#issuecomment-67681553). @tudor

Breaking change: https://github.com/facebook/folly/commit/b85957465746e9275528ceeb4d6646094a586874


Orvid:
As part of the discussion of b859574 internally, it was decided that we wanted to keep the current casing.

PR #378, opened 2016-03-23
Add memory padding and alignment to prevent false sharing

rigtorp:
Ensure no thirdparty thread writes to a cacheline shared by our ringbuffer, readIndex and writeIndex.
Ensure readIndex and writeIndex are located on different cachelines.

### DIFF
--- a/folly/ProducerConsumerQueue.h
+++ b/folly/ProducerConsumerQueue.h
@@ -27,6 +27,8 @@
 #include <type_traits>
 #include <utility>
 
+#include <folly/detail/CacheLocality.h>
+
 namespace folly {
 
 /*
@@ -47,7 +49,7 @@ struct ProducerConsumerQueue {
   // isFull() will return true after size-1 insertions.
   explicit ProducerConsumerQueue(uint32_t size)
     : size_(size)
-    , records_(static_cast<T*>(std::malloc(sizeof(T) * size)))
+    , records_(static_cast<T*>(std::malloc(sizeof(T) * size + detail::CacheLocality::kFalseSharingRange)))
     , readIndex_(0)
     , writeIndex_(0)
   {
@@ -169,8 +171,12 @@ private:
   const uint32_t size_;
   T* const records_;
 
-  std::atomic<unsigned int> readIndex_;
-  std::atomic<unsigned int> writeIndex_;
+  // Align to avoid false sharing between readIndex_ and writeIndex_
+  FOLLY_ALIGN_TO_AVOID_FALSE_SHARING std::atomic<unsigned int> readIndex_;
+  FOLLY_ALIGN_TO_AVOID_FALSE_SHARING std::atomic<unsigned int> writeIndex_;
+
+  // Padding to avoid adjacent allocations to share cache line with writeIndex_
+  char padding_[detail::CacheLocality::kFalseSharingRange - sizeof(writeIndex_)];
 };
 
 }


nbronson:
@rigtorp Was there a particular use case where this made a difference?

rigtorp:
False sharing between readIndex and writeIndex kills performance, this is a
small space trade off (~256 bytes) for a big performance gain. The cache
line ping-ponging also affects performance of unrelated threads, due to
memory bus traffic.

nbronson:
I was hoping to reproduce the case that mattered to you. In this code readIndex_ and writeIndex_ are always accessed near each other, so I was going to see if colocating them on a cache line gave the same benefit with a smaller space penalty.

rigtorp:
If readIndex_ and writeIndex_ are colocated on the same cache line it's
exclusive status will have to ping-pong between the reading and writing
cores. This causes memory bus level synchronization of reader and writer,
needlessly lowering performance.

nbronson:
If readIndex_ and writeIndex_ are separated then there will be two ping-pongs per enqueue or dequeue. If readIndex_ and writeIndex_ are colocated there could be less total ping-ponging, because a miss for one causes both to be filled. On the other hand, two independent misses can be overlapped, so maybe this would be slower even though it has fewer cache misses. Modern microarchitectures are very sophisticated and non-linear, so I'd like to test it on real hardware. It seems that the interleavings mainly depending on what fraction of time the program spends accessing the queue and the "heft" of T (the cost of T construction, access, and destruction, which goes in between the first and last access to the indexes during enqueuing and dequeuing). If you can give me an idea of your use case then I can make sure that I test it.

Hincoin:
I ran some performance tests on an Intel Xeon E3-1505M v5 MB back when @rigtorp first submitted this PR and recently that tested the runtime of `ProducerConsumerQueue` across various high-utilization tests both with and without the cacheline separation on `writeIndex_` and `readIndex_`. 

On 1000 trials running each version of `ProducerConsumerQueue`, I found @rigtorp's proposal to be significant at the t = 4.786 level, which is a p-value of 1.163 * 10^-6, virtually zero. This makes sense too, since the separation would allow for various parallelization opportunities such as permitting a consumer to start reading from `readIndex_` while a producer is concurrently storing to `writeIndex_` and vice-versa. Further, it would allow for a concurrent producer and consumer to store to their respective `Index_` variables without worrying about ping-ponging the cacheline, among others. I would imagine that colocating the two variables upon the same cacheline prevents these sorts of opportunities for performance gains.

However, to quote Walter Bright, " measuring gives you a leg up on experts who are too good to measure.", so I'll run some tests to measure the difference in cache misses for each version and will post results back here later.

rigtorp:
@nbronson You are mistaken in how cache coherency works: https://software.intel.com/en-us/articles/avoiding-and-identifying-false-sharing-among-threads and https://en.wikipedia.org/wiki/MOESI_protocol

@Hincoin It's good to measure, but I've implemented and benchmarked this so many times already. I added a figure to show all the false sharing mitigation present in the implementation I use https://github.com/rigtorp/SPSCQueue

rigtorp:
@nbronson Also good resource: https://www.youtube.com/watch?v=WDIkqP4JbkE (slides http://www.aristeia.com/TalkNotes/codedive-CPUCachesHandouts.pdf)

elliottneilclark:
@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/491829421016186/int_phab) to review.

elliottneilclark:
What benchmarks are you using to test this? Can you include them with the PR?

For the existing benchmark that's in the folly directory I see this:


# Before Applying the PR
```
7	8	28127	196889
8	9	71014	568112
9	10	482	4338
10	11	55	550
11	12	156	1716
12	13	117	1404
13	14	38	494
14	15	10	140
16	17	1	16
============================================================================
folly/test/ProducerConsumerQueueBenchmark.cpp   relative  time/iter  iters/s
============================================================================
----------------------------------------------------------------------------
BM_ProducerConsumer(1048574)                               101.14ns    9.89M
BM_ProducerConsumerAffinity(1048574)                        89.32ns   11.20M
BM_ProducerConsumerLatency(1048574)                         1.67min   10.00m
============================================================================
```

# After Applying the PR

```
7	8	62230	435610
8	9	37130	297040
9	10	250	2250
10	11	60	600
11	12	125	1375
12	13	88	1056
13	14	96	1248
14	15	18	252
15	16	2	30
16	17	1	16
============================================================================
folly/test/ProducerConsumerQueueBenchmark.cpp   relative  time/iter  iters/s
============================================================================
----------------------------------------------------------------------------
BM_ProducerConsumer(1048574)                               100.29ns    9.97M
BM_ProducerConsumerAffinity(1048574)                       106.94ns    9.35M
BM_ProducerConsumerLatency(1048574)                         1.67min   10.00m
============================================================================
```

So from what I see this doesn't seem to make anything faster. It makes the case where threads have affinity worse.

rigtorp:
I don't have any Ubuntu server anymore and I haven't been able to compile Folly on Centos 7. I think the benchmarks are flawed:
As far as I understand the folly benchmark has a queue size (1048574) that is greater then the number of iterations (100000). That means that the throughput benchmarks are just a benchmark for a very inefficient std::memcpy. 

For the latency benchmark: I'm not sure it's a good idea to measure latency like this using `clock_gettime()`. If you use two queues and measure the average RTT of communicating between two threads you can better isolate what you are measuring.

I'm using a similar implementation (https://github.com/rigtorp/SPSCQueue) and I see a big difference in my benchmarks. 

With co-located head and tail pointers:
```
$ perf stat -e L1-dcache-store-misses,instructions,cycles ./SPSCQueueBenchhmark 0 1
25477 ops/ms
282 ns RTT

 Performance counter stats for './SPSCQueueBenchmark 0 1':

         6,310,580      L1-dcache-store-misses                                      
   307,724,002,611      instructions              #    0.94  insns per cycle        
   328,824,617,155      cycles                                                      

      67.516649978 seconds time elapsed
```

With head and tail pointers on separate cache lines:
```
$ perf stat -e L1-dcache-store-misses,instructions,cycles ./SPSCQueueBenchhmark 0 1
28019 ops/ms
247 ns RTT

 Performance counter stats for './SPSCQueueBenchmark 0 1':

           284,599      L1-dcache-store-misses                                      
   243,181,059,980      instructions              #    0.83  insns per cycle        
   292,396,495,466      cycles                                                      

      60.471977493 seconds time elapsed
```

alantsev:
Hello,

@elliottneilclark , could you please reconsider adding paddings for the ProducerConsumerQueue fields?

I think you did not see any improvements in your benchmark, because the benchmark was affected by the false sharing problem itself.

In order to avoid it it's necessary to add additional paddings before and after ProducerConsumerQueue fields.

Here https://github.com/alantsev/folly/commit/a9f26a1aa68df9e1dcc0d4bf31ede01012266d03 I made necessary changes, and wrote a simple benchmark.

Here is the output of the benchmark before adding paddings -

```
$ g++ -I../.. -pthread -I/opt/local/include -I/opt/local/include -std=gnu++1y -g -O2 -c ProducerConsumerQueueBenchmarkSimple.cpp
$ g++ -lpthread -L/opt/local/lib ProducerConsumerQueueBenchmarkSimple.o -o ProducerConsumerQueueBenchmarkSimple
$ ./ProducerConsumerQueueBenchmarkSimple

elapsed time: 1.18136s
elapsed time: 1.26305s
elapsed time: 1.18005s
elapsed time: 1.26775s
elapsed time: 1.17047s
elapsed time: 1.22177s
elapsed time: 1.17361s
elapsed time: 1.21992s
elapsed time: 1.22741s
elapsed time: 1.22133s
^C
$

```

, and here is the output of the benchmark after adding paddings -

```
$ g++ -I../.. -pthread -I/opt/local/include -I/opt/local/include -std=gnu++1y -g -O2 -c ProducerConsumerQueueBenchmarkSimple.cpp
$ g++ -lpthread -L/opt/local/lib ProducerConsumerQueueBenchmarkSimple.o -o ProducerConsumerQueueBenchmarkSimple
$ ./ProducerConsumerQueueBenchmarkSimple
elapsed time: 0.719548s
elapsed time: 0.839374s
elapsed time: 0.834567s
elapsed time: 0.839941s
elapsed time: 0.830283s
elapsed time: 0.833212s
elapsed time: 0.834319s
elapsed time: 0.829544s
elapsed time: 0.837796s
elapsed time: 0.837368s
^C

```

rigtorp:
@alantsev What are the advantages of padding size_ and records_ since they are read only? I ended up rolling my own which also pads the buffer itself (prevents false sharing if the buffers are allocated next to each other): https://github.com/rigtorp/SPSCQueue

alantsev:
@rigtorp, you are absolutelly right - there is no reason for the padding (field pad1 in my change) between size_ and records_ fields.

Although, the point of my initial comment was to bring attention of the owner to the problem in the original benchmark code.

@elliottneilclark, there is a false sharing between test's iters_ field and the queue's writeIndex_ field in the original test. And because of that, the original benchmark does not allow you to see any improvements after applying the current PR.

I addressed this problem with benchmark by adding the padding after the last field in the queue. Alternativelly, you can slightly change the benchmark code -

if you change the loops in the producer / consumer methods like that -

replace -

`for (int i = 0; i < iters_; ++i)`

with

`for (int i = 0, e = iters_; i < e; ++i)`

you will eliminate false sharing from the benchmark and will be able to see the benifits of the current PR.

A.

rigtorp:
@alantsev I get it now. The code should prevent that to begin with by adding at the end of the class:
```cpp
char pad_[detail::CacheLocality::kFalseSharingRange - sizeof(writeIndex_)];
```
If you are near Мінск send me an email and we could meet next time I'm there. I might visit Moskva soon too.

yfeldblum:
a9d90ea14f527b0b719a90f014c8c20e13d54866 basically does this.

PR #396, opened 2016-04-18
fix(FBString): fix FBString.insert(const_iterator p, const value_type c) SIGSEGV when on large fbstring

pp-qq:


folly/FBString.h
```diff
@@ -1484,9 +1484,9 @@ class basic_fbstring {
   iterator insertImplDiscr(const_iterator i,
                            size_type n, value_type c, Selector<1>) {
     Invariant checker(*this);
-
-    assert(i >= begin() && i <= end());
-    const size_type pos = i - begin();
+
```
Orvid: At the very least, the trailing whitespace here needs to go away.

Issue #397, opened 2016-04-18
HHWheelTimer assert-fails when destroy() happens after cancelAll()


fabbree:
`cancelAll()` does not properly reset `count_`.
`HHWheelTimer` ends up in an incoherent state causing `assert(count_ == 0)` failure on destruction.

Zeroing `count_` before returning fixes the issue.

Orvid:
Could you open a PR to make this change rather than reporting the change in an issue?

fabbree:
Done!

PR #399, opened 2016-04-20
Fix for HHWheelTimer assert-fails when destroy() happens after cancel…

fabbree:
…All() #397


afrind:
Shouldn't calling cancelAll() leave count_ equal to zero?

fabbree:
In my understanding it should but currently it doesn't (count_ is not written) .

afrind:
count_ is decremented in the callback’s cancelTimeoutImpl.  Assuming the logic is correct to invoke cancel on all pending timeouts, count_ should get back to 0.  Do you have a simple repro for this?

PR #401, opened 2016-04-21
fix: readvNoInt use wrong read function

chenkan1994:


snarkmaster:
Thanks for the fix! Apparently, this function is unused internally, which is why it could get away with such an egregious bug. I'm going to land this PR, but you'd like to help `readvNoInt` stay non-broken, consider sending a PR with a tiny unit test :)

chenkan1994:
OK, I'll try later.

PR #427, opened 2016-06-19
fix 'make check' error from ssl test

brunomorishita:
avoid this error:

make[3]: *** No rule to make target '../ssl/OpenSSLHashTest.cpp', needed by '../ssl/OpenSSLHashTest.o'.


yfeldblum:
This was fixed in f285646c46f707a602b41ce5c497bb464f18bbc1, but thank you for the pull request.

PR #434, opened 2016-06-30
Proposed changes to (or replacement for) folly::Synchronized

simpkins:
These two commits add a new `Locked<T>` class.  `Locked<T>` is basically like `Synchronized<T>`, with a few API tweaks.  The major differences are:

- It eliminates the `SYNCHRONIZED` macro.  Apart from the normal issues with preprocessor macros, `SYNCHRONIZED` also has some issues when used inside for/while loops, and can also trigger false positive compiler warnings in some cases.
- When using shared mutex types, it makes the choice to acquire a shared vs exclusive lock more explicit.  This helps avoid some anti-patterns mentioned in the Synchronized docs where callers could unintentionally acquire an exclusive lock when they only needed shared access.
- It defaults to a non-shared mutex type (`std::mutex`).  The `RWLocked` alias provides a shared mutex type by default (`folly::SharedMutex`) for users that need separate read/write locking.

This isn't quite ready to go as is (in particular the tests currently require at least gcc-4.9.  I'll update them to make sure they continue to work with gcc-4.8).

I'm interested to get any feedback, but especially on the following items:

- Should this new API simply replace Synchronized<T> instead of being defined as a new Locked<T> type?  Logically it's pretty much the same thing, but the API isn't quite compatible.  Internally at FB we are fine code-modding all existing Synchronized users to OldSynchronized or something, and then adding this new class as Synchronized<T>.  However, this may cause a bit more pain for open source users, who will have to do their own code mods immediately.  Using the new Locked<T> name gives users a slightly longer transition period, but it does introduce a new name for what is largely the same thing.
- There was some discussion about the `wlock()` and `rlock()` method names for `RWLocked<T>`, as opposed to `lock()` and `lockShared()`.  One argument against the latter is that `lock()` is slightly more ambiguous--with a shared mutex type should it acquire an exclusive lock on a non-const object and a shared lock on a const object?  The `wlock()` and `rlock()` names avoid this ambiguity.
- I dropped swap() and the move constructor and assignment operators, since it seemed potentially confusing that these operate only on the data and not the mutex.  I also dropped the copy constructor/assignment as well, since this also would have allowed things like `vector<Locked<int>>`, which doesn't really seem like a sound construction due to object invalidation on vector resizing.  `Locked<T>` can still be copy/move constructed or assigned from another `T` object, just not `Locked<T>`.  That said, I'm open to adding these operators back if there is a compelling argument to keep them.

### DIFF
--- a/folly/LockTraits.h
+++ b/folly/LockTraits.h
@@ -1,0 +1,250 @@
+/*
+ * Copyright 2016 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This module provides traits classes for describing properties about mutex
+ * classes.
+ *
+ * This allows custom mutex/lock classes to be used with folly::Locked.
+ */
+#pragma once
+
+#include <chrono>
+#include <type_traits>
+
+// Android, OSX, and Cygwin don't have timed mutexes
+#if defined(ANDROID) || defined(__ANDROID__) || defined(__APPLE__) || \
+    defined(__CYGWIN__)
+#define FOLLY_LOCK_TRAITS_HAVE_TIMED_MUTEXES 0
+#else
+#define FOLLY_LOCK_TRAITS_HAVE_TIMED_MUTEXES 1
+#endif
+
+namespace folly {
+namespace detail {
+
+/**
+ * An internal helper class for identifying if a lock type supports
+ * lock_shared() and try_lock_for() methods.
+ */
+template <class Mutex>
+class LockTraitsImpl {
+ private:
+  // Helper functions for implementing the traits using SFINAE
+  template <class T>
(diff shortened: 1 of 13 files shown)


simpkins:
@andralex : If you have time, I'd love to hear any feedback you have about this proposed change.

simpkins:
Just an update, but the current plan is to avoid checking this in as a new class, but instead land this as a series of smaller diffs to gradually transform folly::Synchronized to behave more like the new Locked class from this diff.  I'll initially add the new wlock()/rlock()/lock() methods, then withLocked(), and then work on deprecating operator->() and the SYNCHRONIZED macros.

Orvid:
Planned changes have landed, closing this.

andralex:
@simpkins just saw this, thanks for the ping. Hope all is well. The high-level change is the natural consequence of wanting to eliminate the macro (I like it the way one likes an ugly toad, but indeed it may be too jenesaisquoi for its own good.)

A couple of notes:

* If you eliminate the move ctors/move assignments you trigger a bunch of unnecessary copying - and copying inside critical sections that is. Often `Synchronized` objects are members inside other objects, and the compiler won't be able to generate an efficient move constructor for it. I'm not sure I understand the motivation for removal (people are confused thinking... they swap the mutexes?)

* The change is quite hefty (adds 2400 lines) for something that purportedly just removes something. I see there's changes to the documentation, so in balance all might be good. Code should have good power/size ratio, and the fact that it took this much code to actually remove something and add no major functionality may warrant a closer look.

* I see `SYNCHRONIZED_DUAL` is gone, which is a bummer because it does something subtle, well. I'm not sure whether it's intended as an explanation, but the paragraph "Synchronizing several data items with one mutex" in https://github.com/facebook/folly/blob/master/folly/docs/Synchronized.md really should go. It's addressed to folks who want to do multithreaded programming but have no idea how to compose data by putting it in a `struct` or a `tuple`, which is a set with size zero.

PR #443, opened 2016-07-18
Fix compiler warnings

gut:
- unused variables
- suppressing "warning: variable length array ‘vec’ is used [-Wvla]"

### DIFF
--- a/folly/io/async/AsyncSocket.cpp
+++ b/folly/io/async/AsyncSocket.cpp
@@ -19,6 +19,7 @@
 #include <folly/ExceptionWrapper.h>
 #include <folly/SocketAddress.h>
 #include <folly/io/IOBuf.h>
+#include <folly/Portability.h>
 #include <folly/portability/Fcntl.h>
 #include <folly/portability/Sockets.h>
 #include <folly/portability/SysUio.h>
@@ -662,7 +663,13 @@ void AsyncSocket::writeChain(WriteCallback* callback, unique_ptr<IOBuf>&& buf,
   constexpr size_t kSmallSizeMax = 64;
   size_t count = buf->countChainElements();
   if (count <= kSmallSizeMax) {
+
+    // suppress "warning: variable length array ‘vec’ is used [-Wvla]"
+    FOLLY_PUSH_WARNING;
+    FOLLY_GCC_DISABLE_WARNING(vla);
     iovec vec[BOOST_PP_IF(FOLLY_HAVE_VLA, count, kSmallSizeMax)];
+    FOLLY_POP_WARNING;
+
     writeChainImpl(callback, vec, count, std::move(buf), flags);
   } else {
     iovec* vec = new iovec[count];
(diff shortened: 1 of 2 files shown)


folly/io/async/AsyncSocket.cpp
```diff
@@ -1762,18 +1766,6 @@ AsyncSocket::sendSocketMessage(int fd, struct msghdr* msg, int msg_flags) {
     msg->msg_namelen = len;
     totalWritten = tfoSendMsg(fd_, msg, msg_flags);
     if (totalWritten >= 0) {
-      // Call tfo_succeeded to check if TFO was used.
```
Orvid: This probably still needs to be here :P

folly/io/async/HHWheelTimer.cpp
```diff
@@ -236,9 +236,7 @@ size_t HHWheelTimer::cancelAll() {
         if (bucket.empty()) {
           continue;
         }
-        for (auto& cb : bucket) {
-          count++;
-        }
+        count += bucket.size();
```
Orvid: This change was already made.

gut: That's right. They just uploaded. Funny thing huh?
I'll remove that from this PR so the other useful things are focused.

folly/io/async/AsyncSocket.cpp
```diff
@@ -662,7 +662,11 @@ void AsyncSocket::writeChain(WriteCallback* callback, unique_ptr<IOBuf>&& buf,
   constexpr size_t kSmallSizeMax = 64;
   size_t count = buf->countChainElements();
   if (count <= kSmallSizeMax) {
+    // suppress "warning: variable length array ‘vec’ is used [-Wvla]"
+#pragma GCC diagnostic push
```
yfeldblum: Other compilers don't know about this pragma. 

https://github.com/facebook/folly/blob/v2016.07.29.00/folly/Portability.h#L136-L159

folly/io/async/AsyncSocket.cpp
```diff
@@ -662,7 +662,11 @@ void AsyncSocket::writeChain(WriteCallback* callback, unique_ptr<IOBuf>&& buf,
...
+#pragma GCC diagnostic ignored "-Wvla"
```
yfeldblum: If you are trying to build without VLAs, what about unsetting `FOLLY_HAVE_VLA` instead?

Orvid: We have that warning enabled enabled even internally. Completely unsetting it will cause it to no longer compile. It has to be set to either 0 or 1.

folly/io/async/AsyncSocket.cpp
Orvid: We have those, but we actually only use it in a couple of places. I added them but never got around to switching everything over to them.

gut:
Thanks for pointing out the Portability.h. I updated the diff

PR #445, opened 2016-07-27
Fix build in g++ 6.1.1

guoxiao:
In file included from fibers/Fiber.cpp:16:0:
fibers/Fiber.h: In member function ‘std::pair<void*, long unsigned int> folly::fibers::Fiber::getStack() const’:
fibers/Fiber.h:67:42: error: expected primary-expression before ‘>’ token
     const size_t size = std::abs<intptr_t>(
                                                            ^

### DIFF
--- a/folly/fibers/Fiber.h
+++ b/folly/fibers/Fiber.h
@@ -15,6 +15,7 @@
  */
 #pragma once
 
+#include <cstdlib>
 #include <functional>
 #include <thread>
 #include <typeinfo>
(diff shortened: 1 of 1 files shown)


folly/fibers/Fiber.h
```diff
@@ -15,6 +15,7 @@
  */
 #pragma once
 
+#include <complex>
```
yfeldblum: Not `cstdlib`?

yfeldblum: Ah, so that we can implicitly convert to `complex<intptr_t>`.

guoxiao: Sorry. I didn't notice that. 
I just added the header so that it can be build, but now I think it's not a right way to solve the problem.
I couldn't find std::abs<T> in any C++ headers. I can only find std::abs() and std::abs<complex<T>>. I think the author may want to use the former rather than the latter.

guoxiao:
Updated to use `cstdlib` and std::abs().

Orvid:
Believe it or not, the best solution for this is even more fun, as `std::abs` is defined in `<cmath>` (unless there's an overload in `<cstdlib>` that I'm missing?), however, by both including `<complex>` and removing the template parameter, we get codegen that doesn't attempt to convert to a double first. We have a diff up internally already to deal with this due to the current code not being able to be compiled against libc++, but we'll see what happens.

Orvid:
This was fixed in a slightly different way already.

PR #452, opened 2016-08-05
Allow Synchronized.h use under C++11

suspend0:
This is somewhat hacky, but it allows existing users of Synchronized.h which are still compiling under C++11 compilers to continue to work.

In my use I'm using EventBaseLocal from a C++11 compiler.

The code being conditionally disabling was added in b179601dfc42d8e3230d6477d7db8f3d5a8f64dc

### DIFF
--- a/folly/Synchronized.h
+++ b/folly/Synchronized.h
@@ -135,6 +135,7 @@ class SynchronizedBase<Subclass, true> {
    *     return data.getValue();
    *   });
    */
+#if __cplusplus > 201103L
   template <class Function>
   auto withWLock(Function&& function) {
     LockedGuardPtr<Subclass, LockPolicyExclusive> guardPtr(
@@ -147,6 +148,7 @@ class SynchronizedBase<Subclass, true> {
         static_cast<const Subclass*>(this));
     return function(*guardPtr);
   }
+#endif
 
   /**
    * Invoke a function while holding the lock exclusively.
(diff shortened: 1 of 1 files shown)


folly/Synchronized.h
```diff
@@ -135,6 +135,7 @@ class SynchronizedBase<Subclass, true> {
    *     return data.getValue();
    *   });
    */
+#if __cplusplus > 201103L
```
Orvid: This check would currently disable this code under MSVC 2015 :(

suspend0: After seeing your comment, I gave this a shot, but wasn't successful.  The best I could come up with is below, but in the meantime I did find this commit https://github.com/facebook/folly/commit/9fc18f0ac3b21dc2abb5700fba9fc0c27ffd190d#diff-0aeb0bf602d25a066eb2233e06c4a981R48 in which you move away from C++11.  

My PR is based on the condition that C++11 is supported, but if it's not (and that's totally reasonable) then I'm happy to withdraw the PR

-> typename std::result_of<Function(Subclass::DataType&)>::type

yfeldblum: Folly code uses the subset of C++14 that is supported by GCC 4.8.

This particular case can be tweaked to build with C++11 by adding a return type before the opening brace, using either `decltype` or `std::result_of`. Both would work.

suspend0:
cleanup old PRs

PR #456, opened 2016-08-14
Fix build in linux

guoxiao:
The macro `__NR_membarrier` is in `unistd.h`, but `MEMBARRIER_CMD_QUERY`
and `MEMBARRIER_CMD_SHARED` is in `linux/membarrier.h` rather than
`unistd.h`. So we must check them seperately.


lukenels:
I believe https://github.com/facebook/folly/pull/455 solves the same issue as this commit.

PR #458, opened 2016-08-20
Fix a typo in ScopeGuard.h

sas:


suspend0:
`iff` means "if and only if".  It's probably not a typo so much as a somewhat obscure usage.  In my judgement (not a folly maintainer) expanding "iff" to "if and only if" would be clearer than the original.

sas:
Makes sense, didn't think about that. I'll change this.

PR #468, opened 2016-08-30
Fix joining threads twice when exiting EventBaseTest.RunInThread normally

zonr:
10e9cd3 set up SCOPE_EXIT to join threads (wait for them to exit) on
exceptions. However, it is also called on normal exit. Remove the
original code at the end for joining threads.

This fixes EventBaseTest regression.


Orvid:
Whoops, that would be my fault. I missed that when I brought the change over from my Windows side of things :(

PR #478, opened 2016-09-11
Optimize toJson() Performance

### LINKED ISSUE #477: folly::toJson() performance
While working on [nativejson-benchmark](https://github.com/miloyip/nativejson-benchmark), I found that the runtime performance of `folly::toJson()` is extremely low for big JSONs (OS X 10.11.6, clang7.0, 6a6ac91e1fda65d7871390f03be8d19b3591ba45):

Before

After profiling, it is found that the time killer is in:
~~~
// json.cpp
void escapeString(
    StringPiece input,
    std::string& out,
    const serialization_opts& opts) {
  auto hexDigit = [] (int c) -> char {
    return c < 10 ? c + '0' : c - 10 + 'a';
  };

  out.reserve(out.size() + input.size() + 2); // <-- this line
  out.push_back('\"');
  // ...
~~~

This line causes `O(n^2)` time complexity, because every time it expands **just fit** for the string to be appended. This may be only the behavior of [libc++](http://libcxx.llvm.org/). And [C++](http://en.cppreference.com/w/cpp/string/basic_string/reserve) says:

> If new_cap is greater than the current `capacity()`, new storage is allocated, and `capacity()` is made equal or greater than `new_cap`.

The `std::string::push_back()` and `std::string::append()` expand double (or other multiples) of the current capacity, which makes amortized `O(n)`.

After removing that `reserve()` call, the result looks much more normal:

After

I will prepare a pull request on this.

miloyip:
Removing string::reserve() which causes O(n^2) penalty.

Fixes #477

### DIFF
--- a/folly/json.cpp
+++ b/folly/json.cpp
@@ -693,7 +693,6 @@ void escapeString(
     return c < 10 ? c + '0' : c - 10 + 'a';
   };
 
-  out.reserve(out.size() + input.size() + 2);
   out.push_back('\"');
 
   auto* p = reinterpret_cast<const unsigned char*>(input.begin());


facebook-github-bot-1:
Thank you for your pull request.  As you may know, we require contributors to sign our Contributor License Agreement, and we don't seem to have you on file and listed as active anymore.  In order for us to review and merge your code, please email cla@fb.com with your details so we can update your status.

miloyip:
I have just signed the CLA. Thanks bot.

facebook-github-bot-0:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

PR #480, opened 2016-09-15
Find dwarf.h in libdwarf subdirectory

tedjp:
libdwarf headers moved from `/usr/include` to `/usr/include/libdwarf` in [dwarfutils 20160613-2](http://metadata.ftp-master.debian.org/changelogs/main/d/dwarfutils/unstable_changelog).

Adds a check for the library as well, and fixes a tiny typo in the `libunwind` check.

No longer checks for `libdwarf.h`; only `dwarf.h` is used.

### DIFF
--- a/folly/configure.ac
+++ b/folly/configure.ac
@@ -531,12 +531,19 @@ AC_ARG_ENABLE([follytestmain],
    [follytestmain=${enableval}], [follytestmain=no])
 
 use_follytestmain=yes
+# libdwarf used to install in /usr/include, now installs in /usr/include/libdwarf.
+AC_SEARCH_LIBS([dwarf_init], [dwarf])
+AC_CHECK_HEADERS([libdwarf/dwarf.h dwarf.h], [break])
+# Check whether we have both the library and the header
+have_libdwarf=no
+AS_IF([test "x${ac_cv_search_dwarf_init}" != xno && test "x${ac_cv_header_libdwarf_dwarf_h}" = xyes], [have_libdwarf=yes])
+AS_IF([test "x${ac_cv_search_dwarf_init}" != xno && test "x${ac_cv_header_dwarf_h}" = xyes], [have_libdwarf=yes])
 if test "x${follytestmain}" = "xyes"; then
-   AC_CHECK_HEADERS([libdwarf.h dwarf.h],, AC_MSG_ERROR([Please install libdwarf development package]))
+   AS_IF([test "x${have_libdwarf}" = xno], [AC_MSG_ERROR([Please install libdwarf development library and headers])])
    AC_CHECK_HEADERS([libelf.h elf.h],, AC_MSG_ERROR([Please install libelf development package]))
-   AC_CHECK_HEADERS([libunwind.h],, AC_MSG_ERROR([Please install libinwind development package]))
+   AC_CHECK_HEADERS([libunwind.h],, AC_MSG_ERROR([Please install libunwind development package]))
 else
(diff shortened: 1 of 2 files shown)


folly/experimental/symbolizer/Dwarf.cpp
```diff
@@ -19,7 +19,11 @@
 
 #include <type_traits>
 
-#include <dwarf.h>
+#if FOLLY_HAVE_LIBDWARF_DWARF_H
+# include <libdwarf/dwarf.h>
```
Orvid: Any reason we can't just add `/usr/include/libdwarf/` as an extra include path? That would mean these macro checks aren't needed.

folly/configure.ac
```diff
@@ -531,12 +531,19 @@ AC_ARG_ENABLE([follytestmain],
...
 if test "x${follytestmain}" = "xyes"; then
-   AC_CHECK_HEADERS([libdwarf.h dwarf.h],, AC_MSG_ERROR([Please install libdwarf development package]))
+   AS_IF([test "x${have_libdwarf}" = xno], [AC_MSG_ERROR([Please install libdwarf development library and headers])])
    AC_CHECK_HEADERS([libelf.h elf.h],, AC_MSG_ERROR([Please install libelf development package]))
-   AC_CHECK_HEADERS([libunwind.h],, AC_MSG_ERROR([Please install libinwind development package]))
+   AC_CHECK_HEADERS([libunwind.h],, AC_MSG_ERROR([Please install libunwind development package]))
 else
-   AC_CHECK_HEADERS([libdwarf.h dwarf.h],, [use_follytestmain=no])
+   AS_IF([test "x${have_libdwarf}" = xno],, [use_follytestmain=no])
    AC_CHECK_HEADERS([libelf.h elf.h],, [use_follytestmain=no])
```
Orvid: As long as we're poking at this, could you kill the checks for `libelf.h`? We unconditionally reference it as `elf.h` when we `#include` it.

tedjp: The elf.h check is still used to determine whether to use follytestmain - at least when it is requested via --enable-follytestmain. Best to fail/check during configure rather than during build, right?

folly/experimental/symbolizer/Dwarf.cpp
tedjp: I didn't want to add it to the header search directory for every compilation. Not that there would be anything else in there, but it's more in the style of boost/folly/etc to include its files via the name of the project. This case just happens to be messier because it historically didn't ship its headers in a subdir.

Also seems tidier in case it's installed in `/usr/local/include` or wherever the user has installed it — they don't need to explicitly add `-I/usr/local/include/libdwarf`, they can just configure with `CPPFLAGS=-I/usr/local/include`.

tedjp: Alternative implementation that just appends to CPPFLAGS: tedjp/libdwarf-moved-cppflags-fb

folly/configure.ac
Orvid: What I meant was adjusting the checks so they only check for `elf.h` not `libelf.h`.

folly/experimental/symbolizer/Dwarf.cpp
```diff
@@ -19,7 +19,11 @@
...
+#elif FOLLY_HAVE_DWARF_H
```
Orvid: Make this just `#else`, otherwise it will silently not include `dwarf.h` at all. It's much clearer what an issue is when it says it can't find the header rather than not finding the symbols.

PR #483, opened 2016-09-22
Branch-free versions of size() and c_str()

b-pass:
I saw @Gownta's talk on fbstring at cppcon '16 on Tuesday.  He mentioned that it was unfortunate that FBString has a branch in size() and c_str() because these are commonly called functions that we'd all like to be pretty fast.

So here's versions of those functions which don't branch.

### DIFF
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -428,13 +428,12 @@ public:
 
   const Char * c_str() const {
     auto const c = category();
-    if (c == Category::isSmall) {
-      FBSTRING_ASSERT(small_[smallSize()] == '\0');
-      return small_;
-    }
-    FBSTRING_ASSERT(c == Category::isMedium || c == Category::isLarge);
-    FBSTRING_ASSERT(ml_.data_[ml_.size_] == '\0');
-    return ml_.data_;
+    FBSTRING_ASSERT(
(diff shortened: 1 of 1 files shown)


facebook-github-bot-9:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3909502).

Gownta:
We need to run some non-trivial performance tests on string code before we take in any optimizations. The branching is bad, but so is lengthier lists of assembly instructions. (godbolt assembly for this pattern: https://godbolt.org/g/sq29Qa)

I'll keep this diff in mind when we get gcc-5's comparison benchmarks up and running in production. Testing gcc-5 strings is the build team's first string-related priority. Once we've got those numbers, we'll start tuning fbstring.

Thanks for the taking an interest in fbstring code! I'm glad the talk got you interested in the code :)

Orvid:
Also, this appears to have a few bugs somewhere, as it's causing a significant amount of tests to fail.

b-pass:
Thanks @Orvid, I had a stupid typo in there.  Fixed it.

Orvid:
We ended up going with a slightly different form to get the same results, see be4c6d6b3e21914df8a0207ef4f716a09a648745

PR #493, opened 2016-10-12
Fix call to SSL_SESSION_up_ref().

zonr:
`SSL_SESSION_up_ref()` in both BoringSSL and OpenSSL 1.1.0 takes a
`SSL_SESSION*`.


zonr:
+@anirudhvr

PR #501, opened 2016-10-24
No need for libdwarf, only the header file matters

tudor:
folly/experimental/symbolizer doesn't actually link with libdwarf, it only requires dwarf.h.


tudor:
(This allows folly to use either libdwarf or libdw, both of which provide /usr/include/libdwarf.h. libdw is the new hotness, apparently.)

PR #512, opened 2016-11-10
Add include files necessary to use symbolizer

tudor:
Some include files are not installed, causing including symbolizer/Symbolizer.h to fail.


luciang:
Thanks Tudor :)

PR #514, opened 2016-11-13
correct the description about the thread to run

yizhiren:
exchange A and B in this sentence :  
"Or, maybe x will execute in Thread B, but y and/or z will execute in Thread A".

I think x in B and y,z in A is impossible. but x in A and y,z in B is possible, in that example.


yizhiren:
@facebook-github-bot CLA signed !!

PR #534, opened 2016-12-30
Add FOLLY_OPENSSL_HAS_ALPN and FOLLY_OPENSSL_HAS_SNI.

zonr:
This expresses the intention of long predicate on OpenSSL version and
OPENSSL_NO_TLSEXT more clearly.

This also enables ALPN and SNI support when BoringSSL is in use.

ALPN is an essential function to make HTTP/2 work when building Proxygen
with BoringSSL.

### DIFF
--- a/folly/portability/OpenSSL.h
+++ b/folly/portability/OpenSSL.h
@@ -35,6 +35,24 @@ namespace ssl {
 #define FOLLY_OPENSSL_IS_110 (OPENSSL_VERSION_NUMBER >= 0x10100000L)
 #endif // !defined(OPENSSL_IS_BORINGSSL)
 
+// BoringSSL and OpenSSL 1.0.2 later with TLS extension support ALPN.
+#if defined(OPENSSL_IS_BORINGSSL) ||          \
+    (OPENSSL_VERSION_NUMBER >= 0x1000200fL && \
+     !defined(OPENSSL_NO_TLSEXT))
+#define FOLLY_OPENSSL_HAS_ALPN 1
+#else
+#define FOLLY_OPENSSL_HAS_ALPN 0
+#endif
+
+// BoringSSL and OpenSSL 0.9.8f later with TLS extension support SNI.
+#if defined(OPENSSL_IS_BORINGSSL) ||          \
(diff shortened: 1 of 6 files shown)


folly/portability/OpenSSL.h
```diff
@@ -35,6 +36,16 @@ namespace ssl {
 #define FOLLY_OPENSSL_IS_110 (OPENSSL_VERSION_NUMBER >= 0x10100000L)
 #endif // !defined(OPENSSL_IS_BORINGSSL)
 
+// BoringSSL and OpenSSL 1.0.2 later with TLS extension support ALPN.
+#define FOLLY_OPENSSL_HAS_ALPN      \
+  (defined(OPENSSL_IS_BORINGSSL) || \
```
yfeldblum: It is undefined behavior to use `defined` in the definition of a preprocessor symbol.

This form is, if less elegant, at least defined behavior:

```
#if defined(OPENSSL_IS_BORINGSSL) || \
    (OPENSSL_VERSION_NUMBER >= 0x1000200fL && \
     !defined(OPENSSL_NO_TLSEXT))
#define FOLLY_OPENSSL_HAS_ALPN 1
#else
#define FOLLY_OPENSSL_HAS_ALPN 0
#endif
```

Same for `FOLLY_OPENSSL_HAS_SNI` below.

folly/portability/OpenSSL.h
```diff
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <openssl/base.h>
```
yfeldblum: This is causing build failures. Looks like this is only in BoringSSL. Needs a guard.

anirudhvr: @zonr This is automatically included by other headers; you shouldn't need this.

zonr: Done. Follow @anirudhvr's comment to remove the inclusion. Sorry that I forget to verify the build (outside our custom build environment) before uploading. :(

folly/portability/OpenSSL.h
zonr: Done. Thanks a lot! (I should take care of Clang's complaints!)

PR #542, opened 2017-01-30
portability: handle libstdc++ redeclarations

compnerd:
The interfaces here are decorated with `__attribute__((__noreturn__))`
rather than the C++11 attribute `[[noreturn]]`.  The two are subtly
different in behaviour and thus do not compare equal.  The redeclaration
therefore is treated as changing the attributes (rather than just the
attribute spelling).  This will fail to compile as a result.

### DIFF
--- a/folly/portability/BitsFunctexcept.h
+++ b/folly/portability/BitsFunctexcept.h
@@ -23,13 +23,18 @@
 #else
 #include <new> // Some platforms define __throw_bad_alloc() here.
 #include <folly/Portability.h>
+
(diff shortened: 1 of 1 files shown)


Orvid:
We shouldn't be declaring these at all if they are already defined elsewhere. These are intended as backports.

yfeldblum:
So 55e591f927b937b6ca3f23e7f9e5b2af708eb49f does many things, tested internally. But does it also solve the problem in this issue?

yfeldblum:
Assuming this is resolved.

PR #547, opened 2017-02-21
Update googletest to 1.8.0

tedjp:
`make check` was failing because <gmock.h> is included but not supplied. Updating to googletest 1.8.0 includes both googletest & googlemock, so `make check` now runs (and passes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,14 @@ Folly is published on Github at https://github.com/facebook/folly
 folly requires gcc 4.8+ and a version of boost compiled with C++11 support.
 
 Please download googletest from
-https://github.com/google/googletest/archive/release-1.7.0.zip and unzip it in the
-folly/test subdirectory.
+https://github.com/google/googletest/archive/release-1.8.0.zip and unzip it in the
+folly/test subdirectory as `gtest-1.8.0`.
+
+    (cd folly/test && \
+     rm -rf gtest-1.8.0 && \
+     wget https://github.com/google/googletest/archive/release-1.8.0.zip && \
+     unzip release-1.8.0.zip && \
+     mv googletest-release-1.8.0 gtest-1.8.0)
 
 #### Ubuntu 12.04
 
--- a/README.md
+++ b/README.md
@@ -73,14 +73,14 @@ Folly is published on Github at https://github.com/facebook/folly
 folly requires gcc 4.8+ and a version of boost compiled with C++11 support.
 
 Please download googletest from
-https://github.com/google/googletest/archive/release-1.8.0.zip and unzip it in the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -73,8 +73,14 @@ Folly is published on Github at https://github.com/facebook/folly
...
-https://github.com/google/googletest/archive/release-1.7.0.zip and unzip it in the
-folly/test subdirectory.
+https://github.com/google/googletest/archive/release-1.8.0.zip and unzip it in the
+folly/test subdirectory as `gtest-1.8.0`.
+
+    (cd folly/test && \
+     rm -rf gtest-1.8.0 && \
+     wget https://github.com/google/googletest/archive/release-1.8.0.zip && \
+     unzip release-1.8.0.zip && \
+     mv googletest-release-1.8.0 gtest-1.8.0)
```
Orvid: Forwarding internal feedback from @igorsugak:

> Let's unpack it into googletest directory. And update the other part of the patch to remove the version number from the header path please.
> `mv googletest-release-1.8.0 googletest)`

tedjp:
Removed version number from `gtest` directory. Called it `gtest` to avoid confusion with the `googletest` subdirectory (alongside `googlemock`).

Changed from .zip to .tar.gz. Can change back to zip if that's preferred.

Happy to move the block of download-and-unpack into a Makefile target if that's helpful. I think it used to be that way.

folly/test/Makefile.am
```diff
@@ -2,7 +2,7 @@ SUBDIRS = . function_benchmark
 
 ACLOCAL_AMFLAGS = -I m4
 
-CPPFLAGS += -Igtest-1.7.0/include
+CPPFLAGS += -Igtest/googletest/include -Igtest/googlemock/include
```
Orvid: Via @igorsugak:

> Just one more thing. Would you include these via -isystem? That will prevent warnings from googletest showing up in folly builds, which is what we do internally.

folly/test/Makefile.am
```diff
@@ -20,16 +20,16 @@ check_LTLIBRARIES = libfollytestmain.la libgtest.la
 check_PROGRAMS =
 noinst_LTLIBRARIES = thread_local_test_lib.la
 
-libgtest_la_CPPFLAGS = -Igtest-1.7.0 -Igtest-1.7.0/src
-libgtest_la_SOURCES = gtest-1.7.0/src/gtest-all.cc
+libgtest_la_CPPFLAGS = -Igtest/googletest
```
Orvid: Via @igorsugak:
> And here: `s/-I/-isystem /`

PR #549, opened 2017-02-22
Re-enable io tests

tedjp:
Looks like some tests for the `io` directory were left behind when they were moved out of `experimental`. Re-enabled them in the new location, except for the compression test because it takes a lot longer. It's still built and can be executed manually.

Also added some preprocessor guards around compression codecs that might not be compiled in (and would therefore fail during test SetUp).


tedjp:
Needs adjustment; will resubmit.

PR #550, opened 2017-02-22
Re-enable io tests

tedjp:
Looks like some tests for the `io` directory were left behind when they were moved out of experimental. Re-enabled them in the new location, except for the compression test because it takes a lot longer. It's still built and can be executed manually.

Also added some preprocessor guards around compression codecs that might not be compiled in (and therefore fail during test SetUp).

Depends on #547 because the path to gtest appears in the new Makefile.am's `CPPFLAGS`.

### DIFF
--- a/folly/experimental/io/test/Makefile.am
+++ b/folly/experimental/io/test/Makefile.am
@@ -1,12 +1,5 @@
 ACLOCAL_AMFLAGS = -I m4
 
-TESTS = iobuf_test \
-        iobuf_cursor_test
+TESTS =
 
 check_PROGRAMS = $(TESTS)
-
-iobuf_test_SOURCES = IOBufTest.cpp
-iobuf_test_LDADD = $(top_builddir)/libfollyio.la
-
-iobuf_cursor_test_SOURCES = IOBufCursorTest.cpp
-iobuf_cursor_test_LDADD = $(top_builddir)/libfollyio.la $(top_builddir)/libfollybenchmark.la
--- a/folly/experimental/io/test/Makefile.am
+++ b/folly/experimental/io/test/Makefile.am
@@ -1,5 +1,12 @@
 ACLOCAL_AMFLAGS = -I m4
 
-TESTS =
+CPPFLAGS = -I$(top_srcdir)/test/gtest/googletest/include
+ldadd = $(top_builddir)/test/libfollytestmain.la
 
-check_PROGRAMS = $(TESTS)
+check_PROGRAMS = \
+	fs_util_test
+
+TESTS = $(check_PROGRAMS)
+
+fs_util_test_SOURCES = FsUtilTest.cpp
+fs_util_test_LDADD = $(ldadd)
--- a/folly/io/test/.gitignore
+++ b/folly/io/test/.gitignore
@@ -1,0 +1,3 @@
+*.log
+*.trs
+*-test
--- a/folly/io/test/.gitignore
+++ b/folly/io/test/.gitignore
@@ -1,3 +1,0 @@
-*.log
-*.trs
-*-test
--- a/folly/io/test/CompressionTest.cpp
+++ b/folly/io/test/CompressionTest.cpp
@@ -121,30 +121,25 @@ RandomDataHolder randomDataHolder(dataSizeLog2);
 ConstantDataHolder constantDataHolder(dataSizeLog2);
 
 TEST(CompressionTestNeedsUncompressedLength, Simple) {
-  EXPECT_FALSE(getCodec(CodecType::NO_COMPRESSION)->needsUncompressedLength());
-#if FOLLY_HAVE_LIBLZ4
-  EXPECT_TRUE(getCodec(CodecType::LZ4)->needsUncompressedLength());
-#endif
-#if FOLLY_HAVE_LIBSNAPPY
-  EXPECT_FALSE(getCodec(CodecType::SNAPPY)->needsUncompressedLength());
-#endif
-#if FOLLY_HAVE_LIBZ
-  EXPECT_FALSE(getCodec(CodecType::ZLIB)->needsUncompressedLength());
-#endif
-#if FOLLY_HAVE_LIBLZ4
-  EXPECT_FALSE(getCodec(CodecType::LZ4_VARINT_SIZE)->needsUncompressedLength());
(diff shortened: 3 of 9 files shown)


folly/io/test/CompressionTest.cpp
```diff
@@ -122,15 +122,29 @@ ConstantDataHolder constantDataHolder(dataSizeLog2);
 
 TEST(CompressionTestNeedsUncompressedLength, Simple) {
   EXPECT_FALSE(getCodec(CodecType::NO_COMPRESSION)->needsUncompressedLength());
+#if FOLLY_HAVE_LIBLZ4
```
Orvid: Via @yfeldblum:
> Can we avoid the preprocessor? Can this be done with another technique, like computing at compile time (post-preprocessor) or at runtime for each expectation whether that statement needs to be run?

Here and wherever possible.

folly/io/test/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.log
```
Orvid: Via @yfeldblum:
> Remove this file?

tedjp: All tests log their output. Same as the `folly/test` directory.

Orvid: Good enough for me :)

Orvid:
You'll need to rebase onto master as well.

tedjp:
Let me know if you'd like me to resolve the conflict. I presume you'd prefer a clean history _a la_ Linux.

tedjp:
Thanks. Wasn't sure of the protocol :)

tedjp:
One more instance to update.

tedjp:
Sorry, I squashed the last set of preprocessor guard removals into the same commit, so you'll need to import it again.

Orvid:
Re-importing is literally us hitting a button, and is needed for any changes at all, and just updates the version internally, so it's fine :) They'll all get squashed once we merge it anyways.

folly/io/test/CompressionTest.cpp
```diff
@@ -156,6 +169,10 @@ class CompressionTest
 };
 
 void CompressionTest::runSimpleIOBufTest(const DataHolder& dh) {
+  if (!codec_) {
+    return;
```
yfeldblum: This looks like falsely-passing tests. It would be better if these tests were never run. This can be accomplished by computing the list of codec types to instantiate test cases for. I think you can use a `vector` returned from some computation as an argument to `INSTANTIATE_TEST_CASE_P`.

folly/io/test/.gitignore
yfeldblum: Why is this file different from `folly/test/.gitignore`?

folly/experimental/io/test/Makefile.am
```diff
@@ -1,12 +1,5 @@
 ACLOCAL_AMFLAGS = -I m4
```
Orvid: Seeing this after the rebase, this entire file should just be removed, as it no longer does anything useful.

tedjp: Good call. I might reintroduce it with some future work.

folly/io/test/CompressionTest.cpp
```diff
@@ -121,26 +121,42 @@ RandomDataHolder randomDataHolder(dataSizeLog2);
...
   void SetUp() override {
     auto tup = GetParam();
     uncompressedLength_ = uint64_t(1) << std::tr1::get<0>(tup);
     chunks_ = std::tr1::get<1>(tup);
-    codec_ = getCodec(std::tr1::get<2>(tup));
+
+    auto codecType = std::tr1::get<2>(tup);
+    if (hasCodec(codecType)) {
+      codec_ = getCodec(codecType);
+    }
```
tedjp: This is obsolete; will remove.

folly/io/test/CompressionTest.cpp
```diff
@@ -307,14 +330,23 @@ INSTANTIATE_TEST_CASE_P(
...
   void runSimpleTest(const DataHolder& dh);
 
   std::unique_ptr<Codec> codec_;
 };
 
 void CompressionCorruptionTest::runSimpleTest(const DataHolder& dh) {
+  if (!codec_) {
+    return;
+  }
+
```
tedjp: More leftover junk; changes coming.

folly/io/test/.gitignore
tedjp: Consolidated into top-level `.gitignore` for all `test` directories.

folly/io/test/CompressionTest.cpp
```diff
@@ -120,17 +120,58 @@ constexpr size_t dataSizeLog2 = 27;  // 128MiB
...
+      std::back_inserter(supported),
+      hasCodec);
+
+  supported.shrink_to_fit();
+  return supported;
+}
+
+// All compiled-in compression codecs.
+static std::vector<CodecType> const& availableCodecs() {
+  static std::vector<CodecType> codecs;
```
yfeldblum: Remove `static` here - otherwise this will be incorrect if called twice.

Alternatively, fix the function to memoize the result while being safe to call twice, but that is a bit more work and not necessary in unit-tests.

folly/io/test/CompressionTest.cpp
```diff
@@ -120,17 +120,58 @@ constexpr size_t dataSizeLog2 = 27;  // 128MiB
...
+  codecs.reserve(static_cast<size_t>(CodecType::NUM_CODEC_TYPES));
```
yfeldblum: `reserve` and `shrink_to_fit` may be useful for efficiency in production code, but these are unit-tests. Likewise for `supportedCodecs`.

folly/io/test/CompressionTest.cpp
tedjp: Oops, thanks.

folly/io/test/CompressionTest.cpp
tedjp: I'm inclined to keep the `reserve()` call since the size is likely to be almost- or equal to `NUM_CODEC_TYPES` (and never larger). Agreed on `shrink_to_fit()` being overkill.

yfeldblum:
Thanks!

PR #554, opened 2017-03-01
IOStreamBuf (istream/basic_streambuf adaptor for IOBuf)

tedjp:
Allows chained `IOBuf`s to be read via `std::istream` without having to coalesce them into a single contiguous buffer first.

Works with any single-byte type, eg. `char` or `uint8_t`. Not currently usable on multi-byte types like `wchar_t` due to the extra complexity involved in recomposing multi-byte characters that are split across `IOBuf` boundaries. (`static_assert` prevents use with multi-byte types.)

Currently read-only; cannot be used with a `std::ostream`.

Depends the `io/test/Makefile.am` changes in #550. I've included the two necessary commits in this pull request. Happy to resubmit this after #550 completes review if that's preferred.

There are probably good performance reasons to avoid using this in general — and perhaps that should be noted in the header — but it could be useful in conjunction with APIs that can read from an `istream`. My use for this was with Proxygen: passing an incoming message body (chained `IOBuf`) to a 3rd party API.

The use of a `.tcc` file for the implementation of most of the templated functions is unusual for Folly but I wanted to keep them out of the general header. It could easily be renamed to `…-impl.h` or whatever the preferred style is.


Orvid:
`-inl.h` is the preferred style for Folly.

folly/io/IOStreamBuf-inl.h
```diff
@@ -0,0 +1,264 @@
+#include <folly/Likely.h>
```
Orvid: Via @yfeldblum:
> Missing an include guard (`#pragma once` preferably) and a license header.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <folly/io/IOBuf.h>
+
+namespace folly {
+
+template<typename CharT, typename Traits = std::char_traits<CharT> >
+class IOStreamBuf : public std::basic_streambuf<CharT, Traits> {
+  // Due to having to merge single-byte subsets of CharT across IOBuf
+  // boundaries, prevent the use of IOStreamBuf on multi-byte types for now.
+  static_assert(sizeof(CharT) == 1,
```
Orvid: Via @yfeldblum:
> Style. If the arguments must be broken across lines, then each one goes on its own line, and each one is indented +4. Here and elsewhere.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+          "IOStreamBuf doesn't yet work with multi-byte types");
+
+ public:
+  /**
+   * Construct IOStreamBuf using the provided IOBuf, which may be chained.
+   * The IOStreamBuf does not own the IOBuf nor extend the lifetime of it; you
+   * must ensure that the IOBuf provided lasts at least as long as the
+   * IOStreamBuf.
+   */
+  IOStreamBuf(const folly::IOBuf* head);
```
Orvid: Via our internal linter:
> Single-argument constructor `IOStreamBuf(const folly::IOBuf* head)` may inadvertently be used as a type conversion constructor. Prefix the function with the `explicit` keyword to avoid this, or add an `/* implicit */` comment to suppress this warning.

In this case, marking it as `explicit` is the correct solution.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+
+  IOStreamBuf(const IOStreamBuf&) = default;
```
Orvid: Via @yfeldblum:
> Style. Prefer const-on-the-right in new code. `IOStreamBuf const&`. Here and elsewhere.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+  IOStreamBuf& operator=(const IOStreamBuf&) = default;
+  void swap(IOStreamBuf<CharT,Traits>&);
```
Orvid: Via @yfeldblum:
> Style. `, ` (with space) between template arguments. Here and elsewhere.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+
+  virtual ~IOStreamBuf() override = default;
```
Orvid: Via @yfeldblum:
> Need only `override`, not also `virtual`. Here and elsewhere.

folly/io/IOStreamBuf.h
```diff
@@ -0,0 +1,67 @@
...
+  void csetg(const uint8_t* gbeg, const uint8_t* gcurr, const uint8_t* gend);
+
+ private:
+  const folly::IOBuf* head_;
+  const folly::IOBuf* gcur_; // current get IOBuf
+};
+
+}
+
+#include "IOStreamBuf-inl.h"
```
Orvid: Via @yfeldblum:
> Style. Include with the full include path, ie. `folly/io/IOStreamBuf-inl.h`.

folly/io/IOStreamBuf-inl.h
```diff
@@ -0,0 +1,264 @@
...
+  static_cast<typename IOStreamBuf<CharT,Traits>::pos_type>(
+    static_cast<typename IOStreamBuf<CharT,Traits>::off_type>(-1));
+
+// This is called either to rewind the get area (because gptr() == eback())
+// or to attempt to put back a non-matching character (which we disallow
+// on non-mutable IOBufs).
+template <typename CharT, typename Traits>
+typename IOStreamBuf<CharT,Traits>::int_type
+IOStreamBuf<CharT,Traits>::pbackfail(int_type c) {
+  if (this->gptr() != this->eback())
```
Orvid: Via @yfeldblum:
> Style. `if` requires `{}`. Here and elsewhere.

folly/io/IOStreamBuf-inl.h
```diff
@@ -0,0 +1,264 @@
...
+    return traits_type::eof();
+  }
+
+  // Find the next preceding non-empty IOBuf, back to head_
+  // Ensure the object state is not modified unless an earlier input sequence
+  // can be found.
+  const IOBuf* prev = gcur_;
+  do {
+    prev = prev->prev();
+  } while (prev->length() == 0 && prev != head_);
```
Orvid: Via @yfeldblum:
> Can also use `std::find_if` here.

folly/io/IOStreamBuf-inl.h
```diff
@@ -0,0 +1,264 @@
...
+
+  for (const IOBuf* buf = head_; buf != gcur_; buf = buf->next())
+    pos += buf->length();
+
+  return pos + (this->gptr() - this->eback());
+}
+
+template <typename CharT, typename Traits>
+typename IOStreamBuf<CharT,Traits>::pos_type
+IOStreamBuf<CharT,Traits>::seekoff(off_type off,
```
Orvid: Via @yfeldblum:
> This is very long. At the very least, split it into three private member functions for the three possible values of `way`, since the implementations for the three code paths share no code.

folly/io/test/Makefile.am
```diff
@@ -0,0 +1,39 @@
...
+		 compression-test
+
+iobuf_test_SOURCES = IOBufTest.cpp
+iobuf_test_LDADD = $(ldadd)
+
+iobuf_cursor_test_SOURCES = IOBufCursorTest.cpp
+iobuf_cursor_test_LDADD = $(ldadd)
+
+compression_test_SOURCES = CompressionTest.cpp
+compression_test_LDADD = $(top_builddir)/libfolly.la \
```
Orvid: Via @yfeldblum:
> Can drop the first entry to the next line, for style.

Orvid (changes requested):
Via @yfeldblum:
> A bunch of style nits to make this conform to the standard Folly style. I mention each nit category only once, but each mention refers implicitly to many occurrences in that category.

tedjp:
Thanks for all the feedback. I'll send an update when time permits.

folly/io/IOStreamBuf-inl.h
tedjp: Looks like IOBuf::Iterator is a forward-only iterator whereas this code needs to iterate backwards to the head. Furthermore it needs access to the IOBuf object itself but the IOBuf::Iterator only exposes a ByteRange (ie. the data within the IOBuf).

yfeldblum: :+1:

vitaut:
Closing since this is severely out-of-date and iostreams are a legacy API anyway.

PR #558, opened 2017-03-13
Fix the --disable-follytestmain option

msteinert:
This option is useful when compiling on systems that do not want to
include libunwind.


.gitignore
```diff
@@ -20,6 +20,7 @@ folly/test/*_test
 folly/test/*_test_using_jemalloc
 folly/config.*
 folly/configure
+folly/libfolly.pc
```
yfeldblum: Mind splitting this out into a separate PR?

msteinert: Sure, no problem.

PR #559, opened 2017-03-13
Fix the pkg-config filter

msteinert:
I encountered a situation where "pkg-config --libs openssl" included
some linker flags, i.e.

    $ pkg-config --libs openssl
    -Wl,-z,relro -lssl -lcrypto -ldl -lz

This resulted in these linker flags being passed to sed. This change
filters out the linker flags from the pkg-config output.

### DIFF
--- a/folly/m4/fb_filter_pkg_libs.m4
+++ b/folly/m4/fb_filter_pkg_libs.m4
@@ -2,7 +2,7 @@ AC_DEFUN([FB_FILTER_PKG_LIBS],
   [AC_REQUIRE([AC_PROG_SED])
    deps_=`for p in $PKG_DEPS; do pkg-config --libs $p; done`
    filter_=`for l in $deps_;dnl
-     do echo $l | $SED -e 's%\(-l.*\)%-e s/\1//%' -e '/^-L/d';dnl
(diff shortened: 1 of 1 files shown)


folly/m4/fb_filter_pkg_libs.m4
```diff
@@ -2,7 +2,7 @@ AC_DEFUN([FB_FILTER_PKG_LIBS],
   [AC_REQUIRE([AC_PROG_SED])
    deps_=`for p in $PKG_DEPS; do pkg-config --libs $p; done`
    filter_=`for l in $deps_;dnl
-     do echo $l | $SED -e 's%\(-l.*\)%-e s/\1//%' -e '/^-L/d';dnl
+     do echo $l | $SED -e 's%\(-l.*\)%-e s/\1//%' -e '/^-L/d' -e '/^-W/d';dnl
```
pixelb: It would be more general to filter all options except the -l we want to process. I.E.:

    do echo $l | $SED -n '-e 's%\(-l.*\)%-e s/\1//p%'

msteinert: This appears to work:

    do echo $l | $SED -ne 's%\(-l.*\)%-e s/\1//%p';dnl

I'll resubmit the change.

PR #566, opened 2017-03-24
Several fixes to the "SWMRList" example in experimental/hazptr.

Quuxplusone:
commit 4deb5497373c81b2718ad661a13310454d2212d8

    Fix a correctness bug in "SWMRList.h".
    
    Thanks to Maged for the tip! The old code omitted setting a removed node's
    "next" pointer to `nullptr`, which meant that if the writer removed node
    A and then node B = A->next while the reader was looking at B, then the reader
    might happily keep chasing pointers B->next, B->next->next,... without ever
    noticing that it was now on a "dead branch" of the linked list.
    
    (And then there's a bit of a trick: you really do have to remove the node
    first and *then* set its "next" pointer to null, because if you do the null
    assignment first, you'll truncate the list, which means that some readers
    will see a truncated list and return the wrong answer. I've added comments
    to try to explain this to future-me.)
    
    Style nit: Avoid doing multiple atomic operations on the same line of code.

commit af03649a9b207240ac34359284c3eb14c51c6182

    Modernize the parameter-passing conventions in SWMRList.h.
    
    Pass `T`s by const reference, except in `add` where we're likely to want to
    make a copy of the `T` anyway. In that case it would be more "STL-correct"
    to supply two different overloads `add(T&&)` and `add(const T&)`, but this
    is just an example so it makes sense to keep things simple.

commit 525f487312a9ba69b19bea0be153b219b7f722f0

    Fix an undefined behavior in SWMRList example.
    
    Searching an empty SWMRList<int> always invokes undefined behavior by
    comparing an uninitialized `T elem` against `val` on the final line of
    the `contains` function. Besides, this patch allows SWMRList to work
    with `T`s that aren't default-constructible or even copy-constructible.

### DIFF
--- a/folly/experimental/hazptr/example/SWMRList.h
+++ b/folly/experimental/hazptr/example/SWMRList.h
@@ -56,7 +56,7 @@ class SWMRListSet {
   hazptr_domain& domain_;
 
   /* Used by the single writer */
-  void locate_lower_bound(const T v, std::atomic<Node*>*& prev) const {
+  void locate_lower_bound(const T& v, std::atomic<Node*>*& prev) const {
     auto curr = prev->load();
     while (curr) {
       if (curr->elem_ >= v) break;
@@ -78,16 +78,16 @@ class SWMRListSet {
     }
   }
 
-  bool add(const T v) {
+  bool add(T v) {
     auto prev = &head_;
     locate_lower_bound(v, prev);
     auto curr = prev->load();
     if (curr && curr->elem_ == v) return false;
-    prev->store(new Node(v, curr));
+    prev->store(new Node(std::move(v), curr));
     return true;
   }
 
-  bool remove(const T v) {
+  bool remove(const T& v) {
     auto prev = &head_;
     locate_lower_bound(v, prev);
     auto curr = prev->load();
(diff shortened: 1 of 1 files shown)


yfeldblum:
If you are okay with 3 commits being squashed, I can import this for review. If you want 3 separate commits, can you make 3 PRs?

Quuxplusone:
@yfeldblum: I'm okay with squashing.

folly/experimental/hazptr/example/SWMRList.h
```diff
@@ -78,42 +78,45 @@ class SWMRListSet {
     }
   }
 
-  bool add(const T v) {
+  bool add(T v) {
```
djwatson: should be add(T&& v)?

Quuxplusone: From the commit message:

> Pass `T`s by const reference, except in `add` where we're likely to want to make a copy of the `T` anyway. In that case it would be more "STL-correct" to supply two different overloads `add(T&&)` and `add(const T&)`, but this is just an example so it makes sense to keep things simple.

Changing it to `add(T&& v)` without also supplying the `add(const T&)` overload would prevent people from doing e.g.

    int i = 42;
    mylist.add(i);  // error --- "i" is int&, doesn't match int&&

yfeldblum: Since this is an example of hazard pointers, not an example of optimal API design for avoiding accidental copies, this is fine.

Issue #567, opened 2017-03-27
Use of clearenv() prevents compilation on macOS


fxcoudert:
Compilation of folly fails on macOS 10.11 and 10.12 with:

```
libtool: compile:  clang++ -DHAVE_CONFIG_H -I./.. -pthread -I/usr/local/include -I/usr/local/Cellar/openssl/1.0.2k/include -std=gnu++1y -g -O2 -c experimental/EnvUtil.cpp  -fno-common -DPIC -o experimental/.libs/EnvUtil.o
experimental/EnvUtil.cpp:48:15: error: use of undeclared identifier 'clearenv'
  PCHECK(0 == clearenv());
              ^
1 error generated.
make[2]: *** [experimental/EnvUtil.lo] Error 1
```

`clearenv()` is not available on macOS.

yfeldblum:
Fixed in 2d2d3d3f923a6efb92ad2d17b10098b33ea68d02.

ilovezfs:
It looks like this has now been un-fixed.

```
experimental/EnvUtil.cpp:48:15: error: use of undeclared identifier 'clearenv'
  PCHECK(0 == clearenv());
              ^
1 error generated.
make[2]: *** [experimental/EnvUtil.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

terrelln:
Yup, the commit was reverted in 4fe88e4.

Orvid:
I had a commit that was supposed to land to fix this after the revert landed, but things broke and it didn't land.  I'll go poke the system with a stick now and get the fix landed.

ilovezfs:
thanks @Orvid

PR #568, opened 2017-03-29
Two changes to make folly able to build in CygWin

cong1920:
1) Make the generate_fingerprint_tables exec have .exe as file extension on Windows platform;
2) Make an impl of clearenv() in C++ for Windows, CygWin and Apple platforms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ folly/m4/ltoptions.m4
 folly/m4/ltsugar.m4
 folly/m4/ltversion.m4
 folly/m4/lt~obsolete.m4
-folly/generate_fingerprint_tables
+folly/generate_fingerprint_tables*
 folly/FormatTables.cpp
 folly/EscapeTables.cpp
 folly/GroupVarintTables.cpp
--- a/folly/Makefile.am
+++ b/folly/Makefile.am
@@ -599,8 +599,8 @@ libfollybase_la_LDFLAGS = $(AM_LDFLAGS) -version-info $(LT_VERSION)
 libfolly_la_LIBADD = libfollybase.la
 libfolly_la_LDFLAGS = $(AM_LDFLAGS) -version-info $(LT_VERSION) -lssl
 
-FingerprintTables.cpp: generate_fingerprint_tables
-	./generate_fingerprint_tables
+FingerprintTables.cpp: generate_fingerprint_tables$(EXEEXT)
+	./generate_fingerprint_tables$(EXEEXT)
 CLEANFILES += FingerprintTables.cpp
 
 libfollybenchmark_la_SOURCES = Benchmark.cpp
--- a/folly/portability/Stdlib.cpp
+++ b/folly/portability/Stdlib.cpp
@@ -136,3 +136,31 @@ int unsetenv(const char* name) {
 }
 }
 #endif
+
+#if defined(_WIN32) || defined (__CYGWIN__) || defined(__APPLE__)
+#include <string>
+extern char **environ;
+extern "C" {
+int clearenv(void) {
+  char **env = environ;
+  while (env[0] != nullptr) {
+    size_t len = 0;
+    while (env[0][len] != '=') ++len;
+    ++len;
+    {
+      try {
+        std::string name(env[0], len);
+        name[len - 1] = '\0';
+        int result = unsetenv(name.c_str());
+        if (result != 0) {
+          return result;
+        }
+      } catch (const std::bad_alloc&) {
+        return -1;
+      }
+    }
+  }
+  return 0;
+}
+}
+#endif
--- a/folly/portability/Stdlib.h
+++ b/folly/portability/Stdlib.h
@@ -25,7 +25,7 @@
 #endif
 
 extern "C" {
-#ifdef _WIN32
+#if defined(_WIN32)
 // These are technically supposed to be defined linux/limits.h and
 // sys/param.h respectively, but Windows defines _MAX_PATH in stdlib.h,
 // so, instead of creating two headers for a single define each, we put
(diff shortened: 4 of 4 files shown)


cong1920:
Actually, I was trying to add one more test case for the new impl of clearenv() on Windows/Apple platform. But I did not see the folly/experimental/test/EnvUtilTest.cpp is part of the build process. Can anyone point out how to build the EnvUtilTest.cpp and run for me? Thanks!

cong1920:
Haha, I was to file an issue for tracking this PR. Glad to see there is one, #567

ilovezfs:
CC @Orvid

Orvid:
Alright, so, Folly doesn't actually support cygwin or mingw. They may work, but are untested, as it is instead supported directly with MSVC. 

I've had a diff up to do basically this since last week, but the problem is how to handle the error cases. This code currently will break if `setenv` has been used to set any variable to an empty string, it will also fail when any drive options are present in the environment. Right now, with the way this, and the version I have internally handles failures results in `environ` being in a completely unknown (but valid) state, rather than a known invalid state...

What I'm probably going to do is just disable that functionality of the environment utils for now until I can put more thought into how to fail correctly.

Orvid:
I ended up landing my version of `clearenv` in https://github.com/facebook/folly/commit/a5a611496aa06ce04fe4a4411cdafbf5ab8e8ec6 as we do actually need it. As we don't currently support Cygwin, I'm going to close this PR.

wingc-china:
Cool, thanks! Like your implementation!
Not sure the Folly does not support Cygwin but I made it build in Cygwin though (with several hacks of course). First time to hear it supports MSVC directly. I was trying to build in MSVC from beginning but I had problems to get all Folly's dependency libs to build in MSVC so... If it builds in MSVC, can you help me get a guidance doc or something?
Are we planning to support Cygwin? If so I can make some my hacks to better code to make PRs then. Thx!

cong1920:
_...reposting with my real GitHub account..._

Cool, thanks! Like your implementation!
Not sure the Folly does not support Cygwin but I made it build in Cygwin though (with several hacks of course). First time to hear it supports MSVC directly. I was trying to build in MSVC from beginning but I had problems to get all Folly's dependency libs to build in MSVC so... If it builds in MSVC, can you help me get a guidance doc or something?
Are we planning to support Cygwin? If so I can make some my hacks to better code to make PRs then. Thx!

Orvid:
I landed the CMake files last week, so the easiest way to build on Windows is going to be via vcpkg (https://github.com/Microsoft/vcpkg), you want to install `libevent:x64-windows`, `boost:x64-windows`, `openssl:x64-windows`, `double-conversion:x64-windows`, `gflags:x64-windows` and `glog:x64-windows`. The test suite doesn't currently support being compiled without PThread support, but once it does, you'll need `gtest:x64-windows` to build.

Configure Folly via CMake and pass the vcpkg toolset while selecting the 64-bit platform with MSVC 2017, and build with Visual Studio.

cong1920:
Thank you, @Orvid 

I did give the vcpkg a shot to install all libs you listed above. It is so easy and clean. I like it! It really saves time and energy from building all these stuff in CygWin on Windows.

Then I tried to use CMake to build the Folly along with the vcpkg toolchain set. It still needs some changes to the folly-config.h to manually turn off some macro flags to get build pass. But it is much easier than tons of hacks I had to make in CygWin.

Anyway thanks a lot! It is much better than what I have tried. I am going to compose a tutorial if anyone else wants to try building folly and the wangle on Windows. Also I will see if I can quickly learn some CMake to make those folly-config.h changes to be part of vcxproj files generating process via the CMake.

Orvid:
What folly-config.h changes did you need? The one the CMake build uses should be fine.

cong1920:
#define FOLLY_HAVE_BITS_FUNCTEXCEPT_H 0
#define FOLLY_HAVE_CLOCK_GETTIME 0
#define FOLLY_HAVE_CPLUS_DEMANGLE_V3_CALLBACK 0
#define FOLLY_HAVE_FEATURES_H 0
#define FOLLY_HAVE_INT128_T 0
#define FOLLY_HAVE_MEMRCHR 0
#define FOLLY_HAVE_PTHREAD_ATFORK 0
#define FOLLY_HAVE_VLA 0

Besides I have to comment out the line of "#  set(CMAKE_GENERATOR_TOOLSET "v141..." in file CMakeLists.text, I read the comments above but the line itself does not do any trick but only makes CMake report an error. I still build static libs successfully with this line commented out.

Thanks!

Orvid:
Where did you run CMake from?  None of those defines should be required for the build to work.

Which version of CMake is reporting an error?

Also, even though Folly will build without that extra CMake hack, it will fail at link time because that extra CMake hack is merging the base library and the fingerprint table support into a single `folly.lib` file.

cong1920:
My CMake is from the Vcpkg's installation, which is cmake-3.8.0-rc1-win32-x86.

I am not sure it is the cause is from the CMake because if I do not define them 0 there would be build errors that some header files can not be found. These header files should be some non-Windows platform libs.

You could be right because there are multiple lib files in the end. I haven't tried to make Wangle build and link to the lib files, or the single folly.lib file. But if did not comment out that line this is what I got:
`d:\Code\Libs\folly\cmake-build2>d:\Toolz\CMake\bin\cmake .. -G"Visual Studio 15 2017 Win64" "-DCMAKE_TOOLCHAIN_FILE=d:\Code\Libs\vcpkg\scripts\buildsystems\vcpkg.cmake"
CMake Error at CMakeLists.txt:30 (project):
  Generator

    Visual Studio 15 2017 Win64

  does not recognize the toolset

    v141</PlatformToolset></PropertyGroup><ItemDefinitionGroup Condition="'$(ProjectName)'=='folly'"><ProjectReference><LinkLibraryDependencies>true</LinkLibraryDependencies></ProjectReference></ItemDefinitionGroup><PropertyGroup><PlatformToolset>v141

  that was specified.


See also "D:/Code/Libs/folly/cmake-build2/CMakeFiles/CMakeOutput.log"`

PR #587, opened 2017-05-05
Support static gflags library

KindDragon:


KindDragon:
https://github.com/gflags/gflags/blob/master/CMakeLists.txt#L510

PR #588, opened 2017-05-09
Fix FBString with MSVC

KindDragon:


KindDragon:
BTW 919d544e2482da913110b018b572232cdc17718a break MSVC build

PR #594, opened 2017-05-12
Fix compilation ChecksumDetail.cpp with MSVC

KindDragon:
MSVC doesn't support `^` and `&` operator for __m128i type


djwatson:
Thanks!

PR #596, opened 2017-05-15
Move NOMINMAX, WIN32_LEAN_AND_MEAN and _WINSOCK_DEPRECATED_NO_WARNINGS defines from FollyCompiler.cmake into portability/Windows.h

ismail:
Move NOMINMAX, WIN32_LEAN_AND_MEAN and _WINSOCK_DEPRECATED_NO_WARNINGS defines from FollyCompiler.cmake into portability/Windows.h, this way these definitions can be picked up from non-cmake projects.

### DIFF
--- a/folly/portability/Windows.h
+++ b/folly/portability/Windows.h
@@ -37,6 +37,22 @@
 #include <direct.h> // nolint
 #endif
 
+#ifndef _WINDOWS_
+// This is needed because, for some absurd reason, one of the windows headers
+// tries to define "min" and "max" as macros, which messes up most uses of
+// std::numeric_limits.
+#define NOMINMAX
+// Don't include most of Windows.h
+#define WIN32_LEAN_AND_MEAN
+#else
+#error "Folly headers must be included before including Windows.h"
+#endif
+
+#ifndef _WINSOCK2API_
+// Don't deprecate pieces of winsock
+#define _WINSOCK_DEPRECATED_NO_WARNINGS
+#endif
+
 #include <WinSock2.h>
 #include <Windows.h>
 
--- a/CMake/FindGMock.cmake
+++ b/CMake/FindGMock.cmake
@@ -1,6 +1,7 @@
 #
 # Find libgmock
 #
+#  LIBGMOCK_DEFINES     - List of defines when using libgmock.
 #  LIBGMOCK_INCLUDE_DIR - where to find gmock/gmock.h, etc.
 #  LIBGMOCK_LIBRARY     - List of libraries when using libgmock.
 #  LIBGMOCK_FOUND       - True if libgmock found.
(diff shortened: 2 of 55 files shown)


folly/portability/Windows.h
```diff
@@ -37,6 +37,7 @@
 #include <direct.h> // nolint
 #endif
 
+#define NOMINMAX
```
Orvid: This is currently being defined by the CMake build system because it's a requirement for any external code using the library to do the same, otherwise any Windows header included before this will break the assumptions of this code.

ismail: I'm using folly via vcpkg w/ msbuild hence it's not picking up this, for me it's natural to include this in the header and not in the cmake file because not everyone is supposed to use cmake (think autoconf too).

Orvid: You need quite a few other defines to prevent a lot of other warnings being generated if you want to use more than a select few parts of Folly. See https://github.com/facebook/folly/blob/master/CMake/FollyCompiler.cmake#L226-L232

In theory, some of them can be shifted into here, but some extra work will be required to shift the rest, along with more beyond that to deal with the defines to keep it from warning on all of the use of the non-underscore-prefixed version of a lot of functions.

ismail: Those can all be moved to portability/Windows.h easily and it'd be the right fix I think. If you think otherwise feel free to close this pull request. If you agree I can update the request moving rest of the defines.

ismail: @Orvid I'd appreciate a comment, thanks!

Orvid: Ah, yep, sorry, I've been mostly out of the office for the last week. The right fix is definitely to add some of the defines here, but in order to do that, an extra check needs to be added to ensure that `Windows.h` and winsock haven't been included before this header is included.

ismail: @Orvid , thank you will work on that direction!

ismail:
@Orvid Ok made the changes requested.

folly/portability/Windows.h
```diff
@@ -37,6 +37,31 @@
 #include <direct.h> // nolint
 #endif
 
+#ifndef _WINDOWS_
+// We're building in C++ 17 mode, but certain dependencies (Boost) still have
+// dependencies on unary_function and binary_function, so we have to make sure
+// not to remove them.
```
Orvid: `NOMINMAX`, `WIN32_LEAN_AND_MEAN` and `_WINSOCK_DEPRECATED_NO_WARNINGS` are the only defines that can live here. The rest of the defines need to stay where they are, as they effect things other than the headers that this includes.

If either `_WINDOWS_` or `_WINSOCK2API_` is defined, this needs to be a hard `#error`. Folly won't work if `NOMINMAX` isn't defined.

ismail: @Orvid done.

ismail:
Closing to re-work conflicts.

PR #600, opened 2017-05-20
Move NOMINMAX define from FollyCompiler.cmake into portability/Windows.h

ismail:
This way non-cmake projects can successfully use Folly on Windows.

### DIFF
--- a/CMake/FollyCompiler.cmake
+++ b/CMake/FollyCompiler.cmake
@@ -235,7 +235,6 @@ function(apply_folly_compile_options_to_target THETARGET)
   # And the extra defines:
   target_compile_definitions(${THETARGET}
     PUBLIC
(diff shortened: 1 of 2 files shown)


Orvid:
I have a slightly more lax (as long as `min` and `max` aren't defined before hand, it doesn't matter if `Windows.h` has been included in some other way) version of this that I'll put up on Monday. As it stands right now, this would break the build, as I need to make some changes to force the include order of the lib event portability header and make some changes internally to ensure it's treated as a non-portable header by our internal linter :)

ismail:
Ok then, closing this one.

Orvid:
https://github.com/facebook/folly/commit/887367cae9e579cd1b715265a4f1b089beda71da is my version of it, along with the fixes required.

ismail:
@Orvid looks good, thanks!

PR #603, opened 2017-05-23
add std::decay conversion to IsOneOf

Wilbeibi:


yfeldblum:
`IsOneOf` should offer exact comparisons only.

The caller can get the alternative behavior with this alias:

```
template <typename T, typename... Ts>
using IsOneOfDecayed = IsOneOf<T, std::decay_t<Ts>...>;
```

PR #589, opened 2017-05-09
MSVC: Export also folly_base library

KindDragon:
Support Ninja generator to speedup build

PR #606, opened 2017-05-31
Check MSVC compiler version with Ninja generator

KindDragon:


CMakeLists.txt
```diff
@@ -12,8 +12,13 @@ elseif ("${CMAKE_GENERATOR}" STREQUAL "Visual Studio 14 2015 Win64")
   set(CMAKE_GENERATOR_TOOLSET "v140</PlatformToolset></PropertyGroup><ItemDefinitionGroup Condition=\"'$(ProjectName)'=='folly'\"><ProjectReference><LinkLibraryDependencies>true</LinkLibraryDependencies></ProjectReference></ItemDefinitionGroup><PropertyGroup><PlatformToolset>v140")
   set(MSVC_IS_2017 OFF)
 elseif ("${CMAKE_GENERATOR}" STREQUAL "Ninja")
-  message("Folly is being built with Ninja, so assuming VS 2017 is being used.")
-  set(MSVC_IS_2017 ON)
+  if (MSVC_VERSION EQUAL 1910)
```
Orvid: Is this already set at this point? In the Visual Studio generators, this value doesn't get set until the first project with C++ enabled is created.

KindDragon: You right

KindDragon: Can I move `project()` above this `if`?

Orvid: Unfortunately, no, as you can't change the generator toolset after the compiler has been tested :( Once I get the reason #589 was needed resolved, it should be possible to move this entire check into `FollyCompiler.cmake`.

KindDragon: Ahh. ok

Orvid: I've put up a diff internally with this commit: https://github.com/Orvid/folly/commit/a440e8609f38b60bbc1d26cb53c7be95b5941603

It should let the ninja build use whichever compiler is available, along with getting rid of all of the special casing and the need for #589.

PR #609, opened 2017-06-12
fix build due to stats test refactor

theonajim:


Orvid:
Ooops, I thought I'd worked out the issues, but apparently not.

Issue #607, opened 2017-06-04
Folly Futures fail to build with GCC 7


mmlac:
I'm on Arch, folly won't build with GCC 7:

```
libtool: compile:  g++ -DHAVE_CONFIG_H -I./.. -pthread -I/usr/include -std=gnu++1y -g -O2 -MT futures/test/TestExecutor.lo -MD -MP -MF futures/test/.deps/TestExecutor.Tpo -c futures/test/TestExecutor.cpp -o futures/test/TestExecutor.o >/dev/null 2>&1
In file included from ./../folly/futures/Future.h:499:0,
                 from futures/Future.cpp:16:
./../folly/futures/Future-inl.h:236:6: error: prototype for ‘decltype (((folly::Future<T>*)this)->folly::Future<T>::then(forward<Arg>(arg), (forward<Args>)(folly::Future::then::args)...)) folly::Future<T>::then(Executor*, Arg&&, Args&& ...)’ does not match any in class ‘folly::Future<T>’
 auto Future<T>::then(Executor* x, Arg&& arg, Args&&... args)
      ^~~~~~~~~
```

Full log here:  https://gist.github.com/mmlac/213eac9b526e7703ecf8b9ded45db179

Clang still builds. Building with GCC 5 works:  `CC=gcc-5 CXX=g++-5`  (for anyone running into that issue, `sudo pacman -S gcc5` to get version 5 back on arch).

ahkok:
Yes, this is the same issue we're seeing as part of hhvm-3.19.2.

ahkok:
Seems to fix the issue for me. Now it breaks on `mcrouter` though :)

Orvid:
The Pull Request that was originally submitted to fix this broke a few things internally, so I took an alternate approach in b0ec99828f0c782a38947513c5e0c242984e269e that should get things working on GCC 7 again.

PR #621, opened 2017-06-23
Addresses GCC 7 regression w.r.t this-> in decltype

remyabel:
See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81188

Ref: #607


yfeldblum:
Also fix the alignment.

Orvid:
This change seems to be breaking a few things internally (unfortunately none of the things that break are open-source). I'll go digging and also try an alternate approach and see if that makes a difference.

Orvid:
It looks like the alternate approach kept things working internally, so I landed that version instead as b0ec99828f0c782a38947513c5e0c242984e269e. It should get things working on GCC 7 again.

PR #640, opened 2017-07-21
Handle 4-byte unicode code points

### LINKED ISSUE #632: json serialization broken for 4byte UTF-8 sequences?
The following test program bugs me. As it is, it prints garbage (the last cout). But if I remove the "skip_invalid_utf8" option, it works just fine. I am pretty confident that there is no invalid utf8, so what is going on here? In particular, why does parsing work fine, printing also, but not serialization?

```
#include <iostream>
#include <string>
#include <folly/dynamic.h>
#include <folly/json.h>

int main() {
    const std::string test = "\xf0\x90\x8c\xbc"; // 'Valid 4 Octet Sequence'

    std::cout << "std::string: " << std::endl;
    std::cout << test << std::endl;

    folly::json::serialization_opts opts;
    opts.skip_invalid_utf8 = true; // commenting works ?

    const auto d = folly::parseJson("\"" +  test + "\"", opts);

    std::cout << "folly::dynamic: " << std::endl;
    std::cout << d << std::endl;

    const auto r = folly::json::serialize(d, opts);
    std::cout << "folly::json: " << std::endl;
    std::cout << r << std::endl;
}
```

choeger:
fixes #632

Signed-off-by: Christoph Höger <christoph.hoeger@celeraone.com>

### DIFF
--- a/folly/Unicode.cpp
+++ b/folly/Unicode.cpp
@@ -94,7 +94,7 @@ char32_t utf8ToCodePoint(
 
   fst <<= 1;
 
-  for (unsigned int i = 1; i != 3 && p + i < e; ++i) {
+  for (unsigned int i = 1; i != 4 && p + i < e; ++i) {
     unsigned char tmp = p[i];
 
     if ((tmp & 0xC0) != 0x80) {


Orvid:
Could you add a few tests to `test/UnicodeTest.cpp` for this?

folly/Unicode.cpp
```diff
@@ -94,7 +94,7 @@ char32_t utf8ToCodePoint(
 
   fst <<= 1;
 
-  for (unsigned int i = 1; i != 3 && p + i < e; ++i) {
+  for (unsigned int i = 1; i != 4 && p + i < e; ++i) {
     unsigned char tmp = p[i];
 
     if ((tmp & 0xC0) != 0x80) {
```
Orvid: Github won't let me add the comment there, but does this also need to check for surrogates (see line 120)

luciang:
Fixed by https://github.com/facebook/folly/commit/07a29570c3d9b4bf560c08d337fb7d76e961eb76

PR #642, opened 2017-07-21
add import to main in logging example

nicksbyman:


nicksbyman:
Hello, it looks like the #define for the DEFINE_string macro is defined in the header file I imported. Not sure how it would compile without this line of code, unless you ran a make install for the necessary header files, which doesn't seem like the right way to do it. Cheers!

folly/experimental/logging/example/main.cpp
```diff
@@ -16,7 +16,7 @@
 #include <folly/experimental/logging/Init.h>
 #include <folly/experimental/logging/xlog.h>
 #include <folly/init/Init.h>
-
+#include <folly/portability/GFlags.h>
```
yfeldblum: Style nit. Please leave two groups of includes as before, separated by a blank line. This added include should be in the first group.

nicksbyman: fixed :)

PR #643, opened 2017-07-24
Prevent IsOneOf unused template specialization instantiation

stryku:
Current `IsOneOf` implementation does unnecessary work because it instantiates all of the possible template specializations, even if type is same as the first of the tested ones. E.g.
`IsOneOf<char, char, int, float>` will instantiate:
```
IsOneOf<char, char, int, float>
IsOneOf<char, char, int>
IsOneOf<char, char>
IsOneOf<char>
```

With the proposed inheritance, compiler will stop initializing at the first match.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -386,8 +386,14 @@ struct Bools {
 
 // Lighter-weight than Conjunction, but evaluates all sub-conditions eagerly.
 template <class... Ts>
-struct StrictConjunction
-    : std::is_same<Bools<Ts::value..., true>, Bools<true, Ts::value...>> {};
+struct StrictConjunction 
+  : std::is_same<Bools<Ts::value...>, Bools<(Ts::value || true)...>> {};
+
+template <class... Ts>
+struct StrictDisjunction 
+  : Negation<
+      std::is_same<Bools<Ts::value...>, Bools<(Ts::value && false)...>>
+    > {};
 
 } // namespace folly
 
@@ -509,15 +515,8 @@ struct IsRelocatable< std::pair<T, U> >
         IsRelocatable<U>::value> {};
 
 // Is T one of T1, T2, ..., Tn?
-template <class T, class... Ts>
-struct IsOneOf {
-  enum { value = false };
-};
-
-template <class T, class T1, class... Ts>
-struct IsOneOf<T, T1, Ts...> {
-  enum { value = std::is_same<T, T1>::value || IsOneOf<T, Ts...>::value };
-};
+template <typename T, typename... Ts>
+using IsOneOf = StrictDisjunction<std::is_same<T, Ts>...>;
 
 /*
  * Complementary type traits for integral comparisons.


yfeldblum:
How would this compare to using a param-pack expansion over all of `std::is_same<T, Ts>...`, but with no recursion?

stryku:
Let's compare two approaches. One with pack expansion, and the one with recursion.

1. Pack expansion
We can use ready-made `StrictConjuction` here:
```
template <typename T, typename ...Ts>
using IsOneOf = Negation<StrictConjunction<(!std::is_same<T, Ts>:::value)...>>;
```
with initial example, this one will aproximately instantiate such types
```
std::is_same<char, char>
std::is_same<char, int>
std::is_same<char, float>
Bools<true, false, true, true>
Bools<true, true, false, true>
std::is_same<Bools<true, false, true, true>, Bools<true, true, false, true>>
Negation<std::is_same<Bools<true, false, true, true>, Bools<true, true, false, true>>
an alias: IsOneOf<char, char, int, float> to above Negation
```

2. Recursion would create:
```
std::is_same<char, char>
IsOneOf<char, char, int, float>
IsOneOf<char, int, float>//I think this one too
std::conditional<true, std::true_type, IsOneOf<char, int, float>>
```
That was best case. In worst case - without the match e.g.: `IsOneOf<char, int, float, bool>`: 
1. param-pack
```
std::is_same<char, int>
std::is_same<char, float>
std::is_same<char, bool>
Bools<true, true, true, true>
Bools<true, true, true, true>
std::is_same<Bools<true, true, true, true>, Bools<true, true, true, true>>
Negation<std::is_same<Bools<true, true, true, true>, Bools<true, true, true, true>>
```

2. Recursion:
```
std::is_same<char, int>
std::is_same<char, float>
std::is_same<char, bool>
IsOneOf<char, int, float, bool>
IsOneOf<char, float, bool>
IsOneOf<char, bool>
IsOneOf<char>
std::conditional<false, std::true_type, IsOneOf<char, float, bool>>
std::conditional<false, std::true_type, IsOneOf<char, bool>>
std::conditional<false, std::true_type, IsOneOf<char>>
```

Now I can see that implementation with `StrictConjunction` would be cleaner in the code but I'm not sure if we want to instantiate useless `std::is_same`s. What do you think?

yfeldblum:
I did some quick timings with clang in:
https://gist.github.com/yfeldblum/ffae5374aaaa11b03f08919d25b1555b.

stryku:
Nice, I'll change implementation to use `StrictConjunction` then

yfeldblum:
Add `StrictDisjunction` while you are here?

stryku:
I can do this in next ~1,5h, if you want this now, feel free (:

stryku:
Updated commit with proper implementation

stryku:
@yfeldblum one question, could you tell me what does it mean that it's imported? It'll be reviewed or what?

yfeldblum:
Yes. Internal code review is done in our internal Phabricator, so we import patches from Github, preserving original author metadata. After being accepted internally, patches are merged into master in our internal monorepo (monolithic repository). Once merged, commits affecting open-source projects on Github, including those originating in PRs on Github, are synced to Github and PRs are closed.

folly/Traits.h
```diff
@@ -509,15 +509,8 @@ struct IsRelocatable< std::pair<T, U> >
...
-template <class T, class... Ts>
-struct IsOneOf {
-  enum { value = false };
-};
-
-template <class T, class T1, class... Ts>
-struct IsOneOf<T, T1, Ts...> {
-  enum { value = std::is_same<T, T1>::value || IsOneOf<T, Ts...>::value };
-};
+template <typename T, typename ...Ts>
```
yfeldblum: Style nit: `typename... Ts` v.s. `typename ...Ts`. We would like Folly to be `clang-format`-clean.

folly/Traits.h
```diff
@@ -509,15 +509,8 @@ struct IsRelocatable< std::pair<T, U> >
...
+using IsOneOf = Negation<StrictConjunction<Negation<std::is_same<T, Ts>>...>>;
```
yfeldblum: Add `StrictDisjunction`, which should probably exist anyway, and then have `IsOneOf` be defined in terms of that?

yfeldblum: Ah! Right.

folly/Traits.h
stryku: done

folly/Traits.h
yfeldblum: It would still be cleaner to define `IsOneOf` here in terms of `StrictDisjunction`, rather than using double-`Negative` here.

stryku: Ahh, sorry. I read `StrictConjunction` instead of `StrictDisjunction` and wasn't able to understand how would you like to create `StrictConjunction` without the `Negative`s, my bad.

And yes, there is no `StrictDisjunction` yet. I'll create one and update the pr.

stryku: Updated

folly/Traits.h
```diff
@@ -388,6 +388,9 @@ struct Bools {
 template <class... Ts>
 struct StrictConjunction
     : std::is_same<Bools<Ts::value..., true>, Bools<true, Ts::value...>> {};
+    
+template <class... Ts>
+struct StrictDisjunction : Negation<StrictConjunction<Negation<Ts>...>> {};
```
ericniebler: This causes N instantiations of `Negation`. Bad. How about this:

```c++
template <class... Ts>
struct StrictDisjunction
    : Negation<std::is_same<Bools<Ts::value...>, Bools<(Ts::value && false)...>>>;
```

stryku: Good point. I've created helpers for clarity

folly/Traits.h
```diff
@@ -384,10 +384,24 @@ struct Bools {
   }
 };
 
+template <bool... Bs>
+struct AllFalse : std::is_same<Bools<Bs...>, Bools<(Bs && false)...>> {};
```
yfeldblum: @ericniebler - Which of these should be aliases, if any?

stryku: I think none. Compiler claims that the parameter pack is not expanded if those are aliases

yfeldblum: I'm not seeing that? https://godbolt.org/g/enCFGG

yfeldblum: Asking @ericniebler because this could affect other code in folly which he knows about which actually depends on the current deferred evaluation semantics of `StrictConjunction`.

folly/Traits.h
```diff
@@ -384,10 +384,24 @@ struct Bools {
...
+
+template <bool... Bs>
+struct AllTrue : std::is_same<Bools<Bs...>, Bools<(Bs || true)...>> {};
```
ericniebler: I don't think we need another name for `StrictConjunction`. If we do, they should be called `AllOf`, `AnyOf`, and `NoneOf`, like the standard algorithms. But I don't see a need yet. I prefer to keep the API small and simple. `[Strict]Conjunction` and `[Strict]Disjunction` are all we need.

folly/Traits.h
ericniebler: I would make these aliases and internal implementation details, but my preference is to just leave them out entirely.

folly/Traits.h
```diff
@@ -374,7 +374,7 @@ struct Disjunction<T, TList...>
     : std::conditional<T::value, T, Disjunction<TList...>>::type {};
 
 template <typename T>
-struct Negation : std::integral_constant<bool, !T::value> {};
+using Negation = std::integral_constant<bool, !T::value>;
```
ericniebler: This should be a struct, not an alias. It should be possible to name the type `Negation<T>` where `T` is an incomplete type.

folly/Traits.h
stryku: Besides that I'll remove `AllTrue`, `AllFalse` etc. IMHO they're named well. Names `AllOf`, `AnyOf` etc. suggests that they take some kind predicate and check all types against it (like in the STL), don't you think?

>I prefer to keep the API small and simple. 
That was my problem. I thought about folly as a library that need to provide as wide API as possible. Now I understand that new stuff shall be added when there is a need.

folly/Traits.h
stryku: Done.

folly/Traits.h
stryku: Done. I've updated the commit.

stryku:
@yfeldblum @ericniebler Thank you for your patience, it was a learnful pull request

yfeldblum:
@stryku Thank you for contributing!

PR #649, opened 2017-08-01
Fix for building with icc.

gdbentley:


Orvid:
Where is `Futex` being used in a `constexpr` context?

lbrandy:
I too would be interested to hear why you need this for compiling in icc.

But also, this is a good change regardless. Everything that can be should be constant initializable.

gdbentley:
In answer to the questions, here is an example of the errors produced with icc before this fix:

/local/lab_dss/hhvmtest/fbicctest/18_0/fbicc18_maincp10/ws/hhvm-mainline/third-party/folly/src/folly/CallOnce.h(58): error: constexpr constructor calls non-constexpr function "folly::SharedMutexImpl<ReaderPriority, Tag_, Atom, BlockImmediately>::SharedMutexImpl() [with ReaderPriority=false, Tag_=void, Atom=std::atomic, BlockImmediately=false]"
    folly::SharedMutex mutex_;

PR #653, opened 2017-08-10
Handle 4-byte unicode code points.

dryganets:
https://github.com/facebook/folly/pull/640
reworked to have a surrogate check for the 4th symbol.

We need it for react-native to make sure that invalid symbols will be correctly skipped in case somebody passes a corrupted string to javascript core.

Needs to be backported to 2016.09.26.00 to fix the following issue in react-native:
#10756
```
Fatal Exception: java.lang.RuntimeException: Failed to create String from JSON
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
       at android.os.Looper.loop(Looper.java:234)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:193)
       at java.lang.Thread.run(Thread.java:818)
```

### DIFF
--- a/folly/Unicode.cpp
+++ b/folly/Unicode.cpp
@@ -94,7 +94,7 @@ char32_t utf8ToCodePoint(
 
   fst <<= 1;
 
-  for (unsigned int i = 1; i != 3 && p + i < e; ++i) {
+  for (unsigned int i = 1; i != 4 && p + i < e; ++i) {
     unsigned char tmp = p[i];
 
     if ((tmp & 0xC0) != 0x80) {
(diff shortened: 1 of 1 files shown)


Orvid:
As with the original PR, could you add a test for this new behavior in `test/UnicodeTest.cpp` as well?

dryganets:
I see.
I can't make tests run locally.

ie *make check* doesn't work for me on my mac.

I launched brew bootstrap script and downloaded gmock by script from README.md.
My clang version is:
```
gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
```

I fixed a couple of issues, but I'm too lazy to fix them all.
It looks like my version of the compiler is more strict than one which is used by Facebook.
What is recommended compiler version?

I have sources compiling without problem with the one I have.
Maybe there is an option to run only some group of the tests? For example Json only?
I'm new to c++/Make files and gmock so, sorry if my questions are weird to you.
looks like make check runs all of them.

PR #658, opened 2017-08-14
Fixed build issue on ppc64el

j4zzcat:


yfeldblum:
What is the build issue?

folly/detail/Crc32cDetail.cpp
```diff
@@ -281,6 +281,9 @@ uint32_t crc32c_hw(const uint8_t* buf, size_t len, uint32_t crc) {
 
 #else
 
+} // namespace crc32_detail
```
yfeldblum: Make it more "correct" by moving `#if FOLLY_SSE_PREREQ(4, 2)` above to before the `namespace ... {` lines? And then correspondingly move the `#endif`, `#else`, etc.?

Orvid:
Excluding the makefile change, the rest of this is fixed already as far as I can tell.

PR #680, opened 2017-09-20
Fix unittest compile error on futures/test/RetryingTest.cpp.

amin-jabri:
Summary:
On MacOS compiling unittests with "make chek" shows the following compile
error:
```
../futures/test/RetryingTest.cpp:168:26: error: no matching function for call to 'min'
 newMemLimit.rlim_cur = std::min(1UL << 30, oldMemLimit.rlim_max);
```

### DIFF
--- a/folly/futures/test/RetryingTest.cpp
+++ b/folly/futures/test/RetryingTest.cpp
@@ -165,7 +165,7 @@ TEST(RetryingTest, large_retries) {
   PCHECK(getrlimit(RLIMIT_AS, &oldMemLimit) == 0);
 
   rlimit newMemLimit;
-  newMemLimit.rlim_cur = std::min(1UL << 30, oldMemLimit.rlim_max);
+  newMemLimit.rlim_cur = std::min(1ULL << 30, oldMemLimit.rlim_max);
   newMemLimit.rlim_max = oldMemLimit.rlim_max;
   if (!folly::kIsSanitizeAddress) { // ASAN reserves outside of the rlimit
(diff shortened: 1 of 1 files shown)


folly/futures/test/RetryingTest.cpp
```diff
@@ -165,7 +165,8 @@ TEST(RetryingTest, large_retries) {
   PCHECK(getrlimit(RLIMIT_AS, &oldMemLimit) == 0);
 
   rlimit newMemLimit;
-  newMemLimit.rlim_cur = std::min(1UL << 30, oldMemLimit.rlim_max);
+  newMemLimit.rlim_cur =
+      std::min(static_cast<rlim_t>(1UL << 30), oldMemLimit.rlim_max);
```
JoeLoser: Do you have a preference of `static_cast`ing to the proper type to make the first and second operands to min match up? Some people prefer just specifying the template parameter to `std::min` as it is only templated on one type and so that is why the first and second operand must have the same type to get the auto template deduction to work. I don't have a strong opinion on this obviously, but didn't know if Folly had some standard that others follow for this sort of thing. 

Example:
`newMemLimit.rlim_cur =  std::min<rlim_t>(1UL << 30, oldMemLimit.rlim_max);`

yfeldblum: This is fine - thanks for checking.

Note that this forces the two arguments to have the same type, rather than allowing both arguments to be cast implicitly to a third type.

JoeLoser: Thanks for the response. Very fair point on the convertibility.

PR #686, opened 2017-10-12
Modify to use move instead of copy

hellojaewon:
Hope that it reduces minor copy cost


yfeldblum:
The copy should be elided in practice. Is it not?

http://en.cppreference.com/w/cpp/language/copy_elision

hellojaewon:
I missed copy-elision. I apologize giving annoyance to you.

PR #688, opened 2017-10-13
Fix usingJEMalloc() with MS VC++ compiler

KindDragon:
`asm volatile` only works in Gcc and Clang compilers and MSVC can't optimize away `malloc`

/cc @pixelb @Orvid


Orvid:
This was fixed in a different way.

Issue #677, opened 2017-09-16
Issues with Visual C++ compiler


KindDragon:
- [x] [folly/Malloc.h(193)](https://github.com/facebook/folly/blob/271977282fe30d3c351b5e8a6f1692d743996ed1/folly/Malloc.h#L193): error C2059: syntax error: 'volatile'  PR #688
- [x] [folly/detail/Crc32cDetail.cpp(208)](https://github.com/facebook/folly/blob/271977282fe30d3c351b5e8a6f1692d743996ed1/folly/detail/Crc32cDetail.cpp#L208): [fatal error C1009: compiler limit: macros nested too deeply](https://github.com/facebook/folly/commit/74102328f60e450de7d1dab40d1772f82549fd96#commitcomment-23423375)
- [x] [folly/executors/NamedThreadFactory.h(38)](https://github.com/facebook/folly/blob/271977282fe30d3c351b5e8a6f1692d743996ed1/folly/executors/NamedThreadFactory.h#L37-L38): error C2665: 'folly::setThreadName': none of the 2 overloads could convert all the argument types
- [x] folly can't link with static libraries https://gist.github.com/KindDragon/c4e6414bd60242d317bb0387a82dce61#file-package-x64-windows-static-rel-out-log-L2291 PR #693
- [x] [folly/futures/Future-inl.h(242)](https://github.com/facebook/folly/blob/fbc4c23895b0ee3874d9a36401d580a2a8957ba9/folly/futures/Future-inl.h#L242): error C2244: 'folly::SemiFuture<T>::SemiFuture': unable to match function definition to an existing declaration
- [x] [folly/io/async/AsyncSocket.cpp(1589)](https://github.com/facebook/folly/blob/1b0dfd663244c1d743052072a507f1b9e1f88d05/folly/io/async/AsyncSocket.cpp#L1589): error C2039: 'Control': is not a member of 'msghdr'. Win10 SDK define MSG_ERRQUEUE too https://naughter.wordpress.com/2017/07/16/changes-in-the-windows-v10-0-16232-sdk-compared-to-windows-v10-0-15063-sdk-part-one/ PR #689
- [x] [folly/io/async/AsyncSocket.cpp(939)](https://github.com/facebook/folly/blob/master/folly/io/async/AsyncSocket.cpp#L937-L946): error C2065: 'SOL_IP': undeclared identifier
[...]

PR #689, opened 2017-10-13
Windows v10.0.16232 SDK defines MSG_ERRQUEUE, but it doesn't work in …

KindDragon:
…folly

It requires different `msghdr` structure and `recvmsg` ([Sockets.cpp](https://github.com/facebook/folly/blob/2ccbaf9e215b23a357ad6f07de0d1374597b47b8/folly/portability/Sockets.cpp#L261-L305)) doesn't support this flag yet

Compiler error: `folly/io/async/AsyncSocket.cpp(1589): error C2039: 'Control': is not a member of 'msghdr'.`

/cc @maxgeorg @Orvid 
Related to #677

### DIFF
--- a/folly/io/async/AsyncServerSocket.cpp
+++ b/folly/io/async/AsyncServerSocket.cpp
@@ -40,7 +40,7 @@ namespace fsp = folly::portability::sockets;
 namespace folly {
 
 static constexpr bool msgErrQueueSupported =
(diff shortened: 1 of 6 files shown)


Orvid:
It was faster for me to make the changes needed internally after realizing that this hadn't landed yet, so I went that route.

Issue #702, opened 2017-11-02
Build fail with GCC 7.2.1


asierguti:
I am building an HTTP service with proxygen. When building the entire project, I get the following error message.

```
/usr/local/include/folly/Optional.h: In instantiation of ‘const Value* folly::Optional<Value>::Storage::value_pointer() const [with Value = std::chrono::duration<long int, std::ratio<1, 1000> >]’:
/usr/local/include/folly/Optional.h:244:12:   required from ‘constexpr const Value& folly::Optional<Value>::value() const & [with Value = std::chrono::duration<long int, std::ratio<1, 1000> >]’
/usr/local/include/proxygen/lib/http/session/HTTPTransaction.h:981:38:   required from here
/usr/local/include/folly/Optional.h:378:23: error: call of overloaded ‘launder(const std::chrono::duration<long int, std::ratio<1, 1000> >*)’ is ambiguous
         return launder(reinterpret_cast<Value const*>(this->value_));
                ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/folly/Optional.h:67:0,
                 from /usr/local/include/proxygen/lib/http/session/HTTPTransaction.h:14,
                 from /home/asier/projects/hspb/src/app/http/client/TransactionHandler.h:3,
                 from /home/asier/projects/hspb/src/app/http/client/TransactionHandler.cpp:1:
/usr/local/include/folly/lang/Launder.h:29:27: note: candidate: T* folly::launder(T*) [with T = const std::chrono::duration<long int, std::ratio<1, 1000> >]
 FOLLY_NODISCARD inline T* launder(T* in) noexcept {
                           ^~~~~~~
/usr/local/include/folly/lang/Launder.h:51:6: note: candidate: void folly::launder(const void*) <deleted>
 void launder(void const*) = delete;
      ^~~~~~~
/usr/local/include/folly/lang/Launder.h:53:6: note: candidate: void folly::launder(const volatile void*) <deleted>
 void launder(void const volatile*) = delete;
      ^~~~~~~
In file included from /usr/include/c++/7/bits/exception_ptr.h:40:0,
                 from /usr/include/c++/7/exception:142,
                 from /usr/include/c++/7/utility:79,
                 from /usr/include/c++/7/algorithm:60,
                 from /usr/include/boost/heap/d_ary_heap.hpp:12,
[...]

PR #703, opened 2017-11-02
Patch set for bug #702

asierguti:
It looks like a new implementation of libstdc++ or GCC 7.2.1 has the exact same signature for launder. This means that the compiler can't figure out which version to use, folly's or the native one.

### DIFF
--- a/folly/Optional.h
+++ b/folly/Optional.h
@@ -353,7 +353,7 @@ class Optional {
     void clear() {
       if (hasValue_) {
         hasValue_ = false;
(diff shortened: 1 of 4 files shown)


Orvid:
The issue here is that `folly::launder` isn't guarded on whether `std::launder` exists, and `std::launder` is part of the overload set because of (IIUC) ADL. 

The correct solution is to alias `folly::launder` to `std::launder` when `std::launder` is available.

asierguti:
This seems to be fixed in a recent update, hence closing the pull request.

PR #710, opened 2017-11-11
Added addFuture function , which accepts function, expiry time, expir…

badarihp:
Added new function addFuture which can accept a function, expiry time, and expiry callback function.


badarihp:
Hi 
 Any updates on this patch ?

Thanks

yfeldblum:
Executors in folly do not have overloads of `add` which take expirations.

PR #716, opened 2017-11-20
Fix conflict with std::launder in C++17

blackthunder0812:
Fix conflict with std::launder in C++17


yfeldblum:
Is this not working right? From `folly/lang/Launder.h`:

```
#if __cpp_lib_launder >= 201606
namespace folly {
using std::launder;
} // namespace folly
#else
```

blackthunder0812:
Ah, it's updated

PR #733, opened 2018-01-10
C++17 compatibility in SharedMutex.h

### LINKED ISSUE #731: Link error with clang++ and c++17
When I try to compile the following code with this command: `clang++ -o test -std=c++17 -O3 test.cpp -lfolly`

```c++
#include <folly/Synchronized.h>
#include <iostream>
#include <map>

class Test {
public:
  void insert(int key, int value) {
    map_.withWLock([&](auto& myMap) { myMap[key] = value; });
  }

  void remove(int key) {
    map_.withWLock([&](auto& myMap) { myMap.erase(key); });
  }

  template <typename FCT> void forEach(const FCT &fct) const {
    map_.withRLock([&](auto& myMap) {
      //   for (const auto [key, value] : myMap) {
      //     fct(key, value);
      //   }
      for (const auto data : myMap) {
        fct(data.first, data.second);
      }
    });
  }

private:
  folly::Synchronized<std::map<int, int>> map_;
};

int main(int /*argc*/, char ** /*argv*/) {
  Test t;

  for (int i = 0; i < 100; ++i) {
    t.insert(i, i * 256);
  }

  for (int i = 0; i < 50; i += 2) {
    t.remove(i);
  }

  t.forEach([](const auto key, const auto value) {
    std::cerr << key << ": " << value << "\n";
  });

  return 0;
}
```

I've got the following linkage errors:
```
/tmp/test-ce6ef2.o: In function `main':
test.cpp:(.text+0x252): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kMayDefer'
test.cpp:(.text+0x258): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kHasS'
test.cpp:(.text+0x262): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kHasE'
test.cpp:(.text+0x26f): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kIncrHasS'
test.cpp:(.text+0x2fb): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kPrevDefer'
test.cpp:(.text+0x305): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kBegunE'
test.cpp:(.text+0x37f): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kPrevDefer'
test.cpp:(.text+0x389): undefined reference to `folly::SharedMutexImpl<false, void, std::atomic, false>::kBegunE'
```

It works fine with:
* :white_check_mark: `g++ -o test -std=c++17 -O3 test.cpp -lfolly`
* :white_check_mark: `g++ -o test -std=c++14 -O3 test.cpp -lfolly`
* :white_check_mark: `clang++ -o test -std=c++14 -O3 test.cpp -lfolly`
* :white_check_mark:  `clang++ -o test -std=c++17 -O0 -g test.cpp -lfolly`

with:g++ (GCC) 7.2.0 and clang version 5.0.1 (tags/RELEASE_501/final)

hauntsaninja:
Fixes #731

The issue arises because of the following change:
> If a static data member is declared constexpr, it is implicitly inline and does not need to be redeclared at namespace scope. This redeclaration without an initializer (formerly required as shown above) is still permitted, but is deprecated.
(from http://en.cppreference.com/w/cpp/language/static)

### DIFF
--- a/folly/SharedMutex.h
+++ b/folly/SharedMutex.h
@@ -243,7 +243,7 @@ template <
     bool BlockImmediately = false>
 class SharedMutexImpl {
  public:
-  static constexpr bool kReaderPriority = ReaderPriority;
+  static const bool kReaderPriority = ReaderPriority;
   typedef Tag_ Tag;
 
   typedef SharedMutexToken Token;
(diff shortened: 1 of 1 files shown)


yfeldblum:
> formerly required

Folly still needs to work with C++14, and these constants should still be `constexpr`.

hauntsaninja:
This will work perfectly fine with C++14 (we don't need to define the variables outside the class since they're not odr-used). Clearly, however, they're no longer `constexpr`, so I'll close this pull request. 

Also, reading through the spec for inline variables more carefully, I think I had some incorrect ideas about inline's implications for linkage, so probably just a clang bug. Anyhow, thanks for your time.

PR #743, opened 2018-01-30
Add missing override for adjacent_tokens_only

JoeLoser:
Summary:
The UNIX builds treat all warnings as errors, and this missing `override` keyword causes a warning which is then a hard error in the build. In Boost 1.59 and later, `Boost.ProgramOptions` provides a virtual method `adjacent_tokens_only()` which the class `GFlagValueSemanticBase` wants to override. As such, mark it using `override`.


yfeldblum:
This caused build failures internally. I would need to investigate.

yfeldblum:
Please re-submit this as a new PR? For internal tooling reasons.

JoeLoser:
Sure, I can resubmit a new PR.

PR #744, opened 2018-01-31
Add nullability specifiers to LoggerDB

### LINKED ISSUE #740: experimental/logging/LoggerDB.cpp fails to compile due to warnings of FOLLY_N
Compiler errors due to warnings of FOLLY_NULLABLE and FOLLY_NONNULL in experimental/logging/LoggerDB.h

JoeLoser:
Raw pointers are used in `LoggerDB.h` and are missing specifiers for nullability behavior. Clang warns about missing nullability specifiers. Since the Folly build treats warnings as an error, this results in a hard error
for users using Clang to build Folly.

I would say we can close #740 that I created the other day.

There is only one missing specifier that I think requires some discussion on a solution.  In `LoggerDB.h`, we have
```
 template <typename... Args>
  static void internalWarning(
      folly::StringPiece file,
      int lineNumber,
      Args&&... args) noexcept {
    internalWarningImpl(
        file, lineNumber, folly::to<std::string>(std::forward<Args>(args)...));
  }
```
We cannot always annotate  `args` with something like `FOLLY_NULLABLE_UNSPECIFIED` where `FOLLY_NULLABLE_UNSPECIFIED` would be defined as `_Null_unspecified` where appropriate. This is due to the fact that C-style arrays cannot be annotated with nullability specifiers. For example, replacing the above definition with:
```
template <typename... Args>
  static void internalWarning(
      folly::StringPiece file,
      int lineNumber,
      Args FOLLY_NULLABLE_UNSPECIFIED &&... args) noexcept {
    internalWarningImpl(
        file, lineNumber, folly::to<std::string>(std::forward<Args>(args)...));
  }
```
results in
```
candidate template ignored: substitution failure [with Args = <char const[27], int, char const[22], folly::basic_fbstring<char, std::__1::char_traits<char>,
      std::__1::allocator<char>, folly::fbstring_core<char> >>]: nullability specifier '_Null_unspecified' cannot be applied to non-pointer type 'char const[27]'
```

A possible solution would be to create two definitions of `internalWarningImpl` -- one that SFINAEs on the parameter pack. If any of the types are pointer types, we use `FOLLY_NULLABLE_UNSPECIFIED`, otherwise do not annotate the definition when the parameter pack is all of non-pointer types (i.e. where we *cannot* apply any specifier). Is this acceptable/wanted?

Perhaps we can just pragma push/pop and ignore this warning for this small section. Saw there are some portable macros in place for achieving these things inside Folly: `FOLLY_GCC_DISABLE_WARNING`. Any preference?

### DIFF
--- a/folly/experimental/logging/LoggerDB.h
+++ b/folly/experimental/logging/LoggerDB.h
@@ -42,7 +42,7 @@ class LoggerDB {
   /**
    * Get the main LoggerDB singleton.
    */
-  static LoggerDB* get();
+  static LoggerDB* FOLLY_NONNULL get();
 
   ~LoggerDB();
 
@@ -52,7 +52,7 @@ class LoggerDB {
    * This creates the LogCategory for the specified name if it does not exist
    * already.
    */
-  LogCategory* getCategory(folly::StringPiece name);
+  LogCategory* FOLLY_NONNULL getCategory(folly::StringPiece name);
 
   /**
    * Get the LogCategory for the specified name, if it already exists.
@@ -75,7 +75,7 @@ class LoggerDB {
    * use a less verbose level than its parent categories.)
    */
   void setLevel(folly::StringPiece name, LogLevel level, bool inherit = true);
-  void setLevel(LogCategory* category, LogLevel level, bool inherit = true);
+  void setLevel(LogCategory* FOLLY_NONNULL category, LogLevel level, bool inherit = true);
 
   /**
    * Get a LogConfig object describing the current state of the LoggerDB.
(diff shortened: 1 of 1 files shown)


simpkins:
I'm okay with using FOLLY_GCC_DISABLE_WARNING to disable the warnings for internalWarning()

However, I'm curious why you see this warning for LoggerDB.h but not in other parts of folly that have similar code.  For example, folly/Exception.h has functions that forward arguments to folly::to() in similar ways, and does not have nullability annotations.  Do you not get warnings from this file?

JoeLoser:
Just rebased with adding the change of push/popping to disable the warning.

I'm still trying to figure out myself why I'm **only** getting warnings in `LoggerDB.h` about `-Wnullability-completeness` warning that Clang variants support (both AppleClang and Clang). I don't know if it's a function of being able to prove that every code path in other cases never passes a `nullptr` so no annotations are needed, whereas here there actually were some cases that some raw pointers were `nullptr` and other times they were never `nullptr`. If I find more info about this, I'll let you know as I am still curious myself. 

Good news is that I'm inching closer to actually being able to build every TU on Mac OS. Wonder if I'll hit other cases once I am able to define `-DBUILD_TESTS=ON` -- we'll see.

simpkins:
I'm just going to land a change to always compile with `-Wno-nullability-completeness` when using clang.

I don't think we can really fix these warnings correctly without breaking support for some clang versions.  Recent clang versions complain about missing nullabilty specifiers for `std::unique_ptr` arguments.  However, adding specifiers breaks the build for slightly older clang versions, which only understand nullability specifiers on raw pointer arguments.

JoeLoser:
Fair enough.  I think that's probably the more appropriate thing to do here, in addition to using references when things are nonnull. I saw the disabling of warning commit come through in addition to 9656bacc814ce9efd24e178ed5e127c5c365a248, but perhaps we should just transition over to references for other things I annotated with `FOLLY_NONNULL` in this PR?

simpkins:
This warning was disabled in commit d7dfcff981b81569b782169f018421c7185c9cf0.

PR #745, opened 2018-01-31
Add missing override for adjacent_tokens_only

JoeLoser:
Resubmitting as https://github.com/facebook/folly/pull/743 failed internal builds according to @yfeldblum 

Edit: as pointed out in the conversation below, this check needs to be gated for Boost versions [1.59, 1.64] as Boost 1.65 removed this method.

### DIFF
--- a/folly/experimental/ProgramOptions.cpp
+++ b/folly/experimental/ProgramOptions.cpp
@@ -83,9 +83,16 @@ class GFlagValueSemanticBase : public po::value_semantic {
 
   std::string name() const override { return "arg"; }
 #if BOOST_VERSION >= 105900
+  FOLLY_PUSH_WARNING
+  // Unfortunately, internal builds use a patched Boost with this method
+  // removed, so applying `override` would break those builds.
+  // So, the simple case is to just ignore the warning here.
+  // Note that `adjacent_tokens_only` method is removed as of Boost 1.65
+  FOLLY_GCC_DISABLE_WARNING("-Winconsistent-missing-override")
   bool adjacent_tokens_only() const {
     return false;
   }
+  FOLLY_POP_WARNING
 #endif
   bool is_composing() const override { return false; }
   bool is_required() const override { return false; }


simpkins:
adjacent_tokens_only() was removed in boost 1.65.0, so this probably also needs to be gated to less than that version:
https://github.com/boostorg/program_options/commit/7729850bb7eb25b42552b25eff41f141dfa3729f

JoeLoser:
Excellent observation. I'll add the extra check for < Boost 1.65.

JoeLoser:
@yfeldblum did the restriction of defining the member function only for versions using Boost 1.59 through Boost 1.64 seem to make the internal builds happy?

simpkins:
Unfortunately it still caused some issues for us.  Internally some of our builds use a version of boost prior to 1.65 that was explicitly patched to remove this method, so the version check isn't correct for this internal boost library.

I think the easiest thing to do at the moment is to just disable `-Winconsistent-missing-override` around this section of code.

It looks like our internal build scripts always use `-Wno-inconsistent-missing-override` when building with clang.  I'll look at trying to make our compiler warning flags more consistent between our internal builds and our CMake builds.

cc @igorsugak

JoeLoser:
Who knew a one-word change could cause so many internal failures? ;)

Makes sense. I'll disable `-Winconsistent-missing-override` around this section and push it up here shortly. I haven't caught any other missing overrides from my local `Clang` builds, so I'd be curious to know if there are other reasons why you disable it for all targets.

Consistency with internal compiler flags used and the `CMake` flags would be greatly appreciated! 👍

Orvid:
@simpkins The `inconsistent-missing-override` flag is only there for distcc builds, where there is no context of which sources came from system headers and which ones didn't. The primary codebase (including Folly) should be clean for the warnings :)

simpkins:
Just an update here, I believe @yfeldblum and @igorsugak are planning to remove our older internal version of boost that has removed adjacent_tokens_only().  Only a small handful of things were still using it.  Once that is done we can import this pull request with the simple boost version check.

@yfeldblum / @igorsugak: is that still the plan, or should we land this version with the warning suppression in the meantime?

PR #747, opened 2018-01-31
Fix some unused parameters

JoeLoser:
Summary:
Fix a few unused parameters. One was unconditional; the other two are conditional based on some configuration values set during CMake configure time.

### DIFF
--- a/folly/io/async/AsyncServerSocket.cpp
+++ b/folly/io/async/AsyncServerSocket.cpp
@@ -736,7 +736,11 @@ int AsyncServerSocket::createSocket(int family) {
   return fd;
 }
 
+#ifndef TCP_NOPUSH
 void AsyncServerSocket::setupSocket(int fd, int family) {
+#else
+void AsyncServerSocket::setupSocket(int fd, int) {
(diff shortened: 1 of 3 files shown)


folly/io/async/AsyncServerSocket.cpp
```diff
@@ -736,7 +736,11 @@ int AsyncServerSocket::createSocket(int family) {
   return fd;
 }
 
+#ifndef TCP_NOPUSH
 void AsyncServerSocket::setupSocket(int fd, int family) {
+#else
+void AsyncServerSocket::setupSocket(int fd, int) {
+#endif
   // Put the socket in non-blocking mode
```
yfeldblum: The less conditional compilation, the better.

Instead, just stick this as the first line:

```
(void)family;
```

JoeLoser: Ah good point. Just saw this after my similar change this morning. I'll just stick it as the first line for those two functions who have conditional unused parameters. I'll send a change up tonight.

PR #749, opened 2018-01-31
Simple typo correction on ProducerConsumerQueue.md

ll-h:
"iff queue" replaced by "if the queue" and another "iff" replaced by "if".


yfeldblum:
"iff" is a contraction of "if and only if" and its usage here is not wrong.

ll-h:
Ok I did not know that. Thanks for the reply. I'm closing the request.

Issue #752, opened 2018-02-04
DynamicConverter missing specialization for vector<bool>


JoeLoser:
Summary:
`DynamicConvertTest` test case `construct` tests that `toDynamic` is well-formed for several containers and ranges. Unfortunately, there's a bit of an oddity when it comes to `vector<bool>` is often regarded as "not an STL container" or rather it is special.  It appears this snippet is ill-formed according to `Clang 5.0.1`.

```
 vector<bool> vb{true, false};
 dynamic d = dynamic::array(true, false);
 EXPECT_EQ(d, toDynamic(vb));
```
Snippet of compiler error:
```
../folly/DynamicConverter.h:312:12: error: no matching conversion for functional-style cast from 'const std::__1::__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > >' to 'folly::dynamic'
    return dynamic(x);
           ^~~~~~~~~
../folly/DynamicConverter.h:381:64: note: in instantiation of member function 'folly::DynamicConstructor<std::__1::__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > >, void>::construct' requested here
  return DynamicConstructor<typename std::remove_cv<T>::type>::construct(x);
                                                               ^
../folly/DynamicConverter.h:354:19: note: in instantiation of function template specialization 'folly::toDynamic<std::__1::__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > > >' requested here
      d.push_back(toDynamic(item));
                  ^
../folly/DynamicConverter.h:381:64: note: in instantiation of member function 'folly::DynamicConstructor<std::__1::vector<bool, std::__1::allocator<bool> >, void>::construct' requested here
  return DynamicConstructor<typename std::remove_cv<T>::type>::construct(x);
                                                               ^
../folly/test/DynamicConverterTest.cpp:389:18: note: in instantiation of function template specialization 'folly::toDynamic<std::__1::vector<bool, std::__1::allocator<bool> > >' requested here
    EXPECT_EQ(d, toDynamic(vb));
```

The solution should be to specialize `construct` with the appropriate SFINAE machinery to make it work with `vector<bool>`. Something like this may work:
```
template<class C>
struct DynamicConstructor<
    C,
    typename std::enable_if<
        std::is_same<C, std::vector<bool>>::value
    >::type>
{
    static dynamic construct(const std::vector<bool>& x)
    {
        dynamic d = dynamic::array;
        for(const auto& item : x)
        {
[...]

PR #756, opened 2018-02-06
Disable part of DynamicConverterTest which fails for Clang

JoeLoser:
Summary:
`folly::dynamic` is missing a template specialization for construction from
`std::vector<bool>`. See issue #752 for more info. This issue definitely arises
using Clang 5.0.1 on Mac, but likely comes up in earlier Clang versions too.

### DIFF
--- a/folly/test/DynamicConverterTest.cpp
+++ b/folly/test/DynamicConverterTest.cpp
@@ -383,11 +383,14 @@ TEST(DynamicConverter, construct) {
     EXPECT_EQ(d, toDynamic(c));
   }
 
+// See https://github.com/facebook/folly/issues/752
+#if !defined(__clang__)
   {
     vector<bool> vb{true, false};
     dynamic d = dynamic::array(true, false);
     EXPECT_EQ(d, toDynamic(vb));
   }
+#endif
 }
 
 TEST(DynamicConverter, errors) {


JoeLoser:
@simpkins @yfeldblum 
Does this test case not break internal builds using Clang as well?

JoeLoser:
Bump

yfeldblum:
Is this an issue with clang or with libc++ (the stdlib impl shipping with clang++)?

JoeLoser:
I am using `libc++` (verified by checking `_LIBCPP_VERSION != 0`). Can you help me understand which Clang variant(s) (and whether it is using `libstdc++` or `libc++`) that Folly is tested against internally? This will help us bissect the problem a bit to figure out whether it's a Folly bug in missing specialization for `vector<bool>`, or whether it's a standard-library implementation issue.

JoeLoser:
Since I have an old Clang 3.8.1 laying around (which also uses `libc++`), I tried this test with that. It still failed. Given that this test passes on GCC internal builds and Travis CI builds, I suspect the problem of this piece of code has been reduced down to one of the following:

1. Ill-formed code in Folly in handling `vector<bool>` for the `dynamic` type
2. Something in `libc++` implementation not doing quite what we want.

I'm leaning towards the former, but I'd need to be careful in examining all of the SFINAE done in `DynamicConvert.h` to verify that claim.

JoeLoser:
Closing this PR as we have solved the problem rather than just disabling this part of the test. Landed in 2b5eff826a7736a9e3e129b678b985df7e90d611

PR #758, opened 2018-02-06
[CMake] Add support for compiling on Mac

JoeLoser:
Summary:
There are some extra warnings that need to get marked intentionally
to not be errors for the Apple builds.

Note that while this will open the door to being able to build all of
Folly using Mac OS, there are still several (almost 30) failing tests
that I still need to look into.

Ideally, we would harmonize internal compiler flags used for CI
with various OSes and compilers to match that of the open source
project.

### DIFF
--- a/folly/compression/Counters.cpp
+++ b/folly/compression/Counters.cpp
@@ -15,7 +15,7 @@
  */
 
 #include <folly/compression/Counters.h>
-#include <folly/folly-config.h>
+#include <folly/portability/Config.h>
 
 namespace folly {
 #if FOLLY_HAVE_WEAK_SYMBOLS


CMake/FollyCompilerUnix.cmake
```diff
@@ -26,6 +24,20 @@ function(apply_folly_compile_options_to_target THETARGET)
...
       -Wunused-result
       -Wnon-virtual-dtor
-      ${FOLLY_CXX_FLAGS}
-  )
+      ${FOLLY_CXX_FLAGS})
+
+  if (APPLE)
+    target_compile_options(${THETARGET}
+      PRIVATE
+        -Wno-nullability-completeness # See https://github.com/facebook/folly/issues/740
```
simpkins: Commit d7dfcff981b8 already added this flag.

CMake/FollyCompilerUnix.cmake
```diff
@@ -26,6 +24,20 @@ function(apply_folly_compile_options_to_target THETARGET)
...
+        -Wno-inconsistent-missing-override # See https://github.com/facebook/folly/pull/745
```
simpkins: I would vote for adding this flag the same way I did with -Wno-nullability-completeness in d7dfcff981b8.
We should use this flag whenever we are compiling with clang, not just on Mac platforms.

CMake/FollyCompilerUnix.cmake
```diff
@@ -26,6 +24,20 @@ function(apply_folly_compile_options_to_target THETARGET)
...
+        -Wno-format # See https://github.com/facebook/folly/issues/757
```
simpkins: I think we should get the string format fixes in from pull request 765 instead.
If we land those, can we drop this flag?

CMake/FollyCompilerUnix.cmake
JoeLoser: Very valid point -- I should make it for Clang and not just Apple builds. Can easily add the compiler flag using `CHECK_CXX_COMPILER_FLAG` and will do so shortly. Thanks!

CMake/FollyCompilerUnix.cmake
JoeLoser: Yep, we can drop the `-Wno-format` if #765 PR goes in first. I'll operate on the assumption that 765 will go in first and just remove it on this branch.

Issue #757, opened 2018-02-06
StringTest warning with Clang on printf specifiers


JoeLoser:
Summary:
There are some formatting issues with specifiers to `stringPrintf` and its argument passed. In particular, the test case `NumericFormats` in `StringTest.cpp` is the culprit:
```
  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffff));
  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffUL));
```

This shows up with Clang 5.0.1 on Mac.

Output errors:
```
error: format specifies type 'long long' but the argument has type 'unsigned long' [-Werror,-Wformat]
  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffff));

error: format specifies type 'long long' but the argument has type 'unsigned long' [-Werror,-Wformat]
  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffUL));
```

JoeLoser:
Closed with https://github.com/facebook/folly/pull/765

PR #765, opened 2018-02-08
Fix format warning in StringTest.cpp

JoeLoser:
Summary:
Clang warns about formatting when we were using `stringPrintf` to print
the result of overflowing an unsigned 64-bit value.

Closes issue #757.

### DIFF
--- a/folly/test/StringTest.cpp
+++ b/folly/test/StringTest.cpp
@@ -48,8 +48,8 @@ TEST(StringPrintf, NumericFormats) {
   EXPECT_EQ("5000000000", stringPrintf("%lld", 5000000000LL));
   EXPECT_EQ("-5000000000", stringPrintf("%lld", -5000000000LL));
   EXPECT_EQ("-1", stringPrintf("%d", 0xffffffff));
-  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffff));
-  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffUL));
+  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffLL));
+  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffULL));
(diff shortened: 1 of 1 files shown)


folly/test/StringTest.cpp
```diff
@@ -48,8 +48,8 @@ TEST(StringPrintf, NumericFormats) {
   EXPECT_EQ("5000000000", stringPrintf("%lld", 5000000000LL));
   EXPECT_EQ("-5000000000", stringPrintf("%lld", -5000000000LL));
   EXPECT_EQ("-1", stringPrintf("%d", 0xffffffff));
-  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffff));
-  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffUL));
+  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffLL));
+  EXPECT_EQ("-1", stringPrintf("%" PRId64, 0xffffffffffffffffULL));
```
simpkins: Given that the types of these literals are long long and unsigned long long, it seems more correct to use "%lld".

If we also want to explicitly test with 64-bit types we can also add some checks where we explicitly
cast the arguments to 64-bit types.

  EXPECT_EQ("-1", stringPrintf("%" PRId64, static_cast<int64_t>(0xffffffffffffffffLL)));
  EXPECT_EQ("-1", stringPrintf("%" PRId64, static_cast<uint64_t>(0xffffffffffffffffULL)));

yfeldblum: This looks like it is intended to test the `"%" PRId64` format, so, best to cast the literals.

JoeLoser: I'm with @yfeldblum here.  On my system, `PRId64` expands to `lld` since I am on a 64-bit machine. I believe the intent of this part of the test is to use the cross-platform specifiers in `inttypes.h`. As such, I'll just cast the literals as you both suggest.

PR #770, opened 2018-02-13
Fix using Atom with MSVC

KindDragon:


folly/detail/Futex.h
```diff
@@ -44,7 +44,7 @@ enum class FutexResult {
  */
 template <template <typename> class Atom = std::atomic>
 struct Futex : Atom<uint32_t> {
-  using Atom<uint32_t>::Atom;
+  explicit constexpr Futex(uint32_t init = 0) : Atom<uint32_t>(init) {}
```
yfeldblum: Define both ctors listed here, with no default arguments.

http://en.cppreference.com/w/cpp/atomic/atomic/atomic

KindDragon:
Fixed

KindDragon:
MSVC bug https://developercommunity.visualstudio.com/content/problem/60185/cannot-compile-template-class-inheriting-construct.html

yfeldblum:
Can also add a link to this PR into that MSVC bug report.

PR #768, opened 2018-02-09
[CMake] Closer to compiling with Clang

JoeLoser:
Summary:
This brings us closer to being able to compiling the open
source version of Folly with Clang. There are a few other
open pull requests that need to be addressed to allow us to
compile on Clang without any warnings that aren't already
explicitly disabled. See pull requests `745`  and `756`.

PR #772, opened 2018-02-18
[CMake] Closer to compiling with Clang

JoeLoser:
Summary:
This brings us closer to being able to compile the open
source version of Folly with Clang. There are a few other
open pull requests that need to be addressed to allow us to
compile on Clang without any warnings that aren't already
explicitly disabled. See pull requests `745` and `756`.

@yfeldblum Sorry for the duplicate of https://github.com/facebook/folly/pull/768 -- I had rebased and force-pushed the reverting of the original changeset to then apply the new changeset as requested from the internal code review feedback. The result was that the PR got closed due to showing zero commits on the branch and then it wouldn't reopen despite force pushing to the branch with this changeset.  Apologize for that.

### DIFF
--- a/CMake/FollyCompilerUnix.cmake
+++ b/CMake/FollyCompilerUnix.cmake
@@ -11,7 +11,7 @@ function(apply_folly_compile_options_to_target THETARGET)
       "FOLLY_XLOG_STRIP_PREFIXES=\"${FOLLY_DIR_PREFIXES}\""
   )
   target_compile_options(${THETARGET}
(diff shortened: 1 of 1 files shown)


JoeLoser:
After some thought, perhaps we just need to really address #745 and then this PR isn't needed. The only thing holding that up I believe is removing the internal patched versions of Boost which removed `adjacent_tokens_only()`. I think you and @igorsugak were working on that last I knew?

PR #773, opened 2018-02-18
Fix problem when FOLLY_HAVE_VLA is undefined

KindDragon:


KindDragon:
I think this is the last remaining error

yfeldblum:
@KindDragon Is this solved for your case?

JoeLoser:
Can you elaborate on what the problem was exactly? I only see `FOLLY_HAVE_VLA` used in the code once and it's wrapped around `BOOST_PP_IF` -- shouldn't it handle the case just fine when `FOLLY_HAVE_VLA` is undefined?

yfeldblum:
Docs say [different](http://www.boost.org/doc/libs/1_66_0/libs/preprocessor/doc/ref/if.html):

> Valid values range from `0` to `BOOST_PP_LIMIT_MAG`.

KindDragon:
Yes 👍

KindDragon:
Currently, I only have problems because of missing Find modules for zstd and lz4. For Snappy you can use `find_package(Snappy)` and target `Snappy::SnappyTargets` 
https://github.com/facebook/folly/blob/1657dcf2c336a98a0167d4f10747957a912952d6/CMake/folly-deps.cmake#L101-L114

PR #778, opened 2018-02-22
[CMake] Add FindLZ4 module

JoeLoser:
Summary:
Add a CMake find_package script for LZ4

@simpkins

### DIFF
--- a/CMake/FindLZ4.cmake
+++ b/CMake/FindLZ4.cmake
@@ -1,0 +1,22 @@
+# Finds liblz4.
+#
+# This module defines:
+# LZ4_FOUND
+# LZ4_INCLUDE_DIR
+# LZ4_LIBRARY
+#
+
+find_path(LZ4_INCLUDE_DIR NAMES lz4.h)
+find_library(LZ4_LIBRARY NAMES lz4)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(
+    LZ4 DEFAULT_MSG
+    LZ4_LIBRARY LZ4_INCLUDE_DIR
+)
+
+if (LZ4_FOUND)
+    message(STATUS "Found LZ4: ${LZ4_LIBRARY}")
+endif()
+
+mark_as_advanced(LZ4_INCLUDE_DIR LZ4_LIBRARY)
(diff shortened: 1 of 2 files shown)


CMake/FindLZ4.cmake
```diff
@@ -0,0 +1,35 @@
...
+# This module defines:
+# LZ4_FOUND
+# LZ4_INCLUDE_DIR
+# LZ4_LIBRARY
+#
+
+find_path(LZ4_INCLUDE_DIR NAMES lz4.h)
+find_library(LZ4_LIBRARY NAMES lz4)
+
+# We require LZ4_compress_default() which was added in v1.7.0
```
yfeldblum: Maybe just check for that symbol?

JoeLoser: I'm not entirely certain we just require that symbol, though, actually.  First, here are all the undefined LZ4 symbols we are expected to link with. This can be generated by

```
nm <build_dir>/*.a | grep -rniI lz4 | c++filt | grep 'U' | grep -v "::" | uniq | sort  | awk '{print $4}'                                                                                         
```


```
_LZ4F_compressFrame
_LZ4F_compressFrameBound
_LZ4F_createDecompressionContext
_LZ4F_decompress
_LZ4F_freeDecompressionContext
_LZ4F_getErrorName
_LZ4F_isError
_LZ4_compressBound
_LZ4_compress_HC
_LZ4_compress_default
_LZ4_decompress_safe
_LZ4F_compressFrame
_LZ4F_compressFrameBound
_LZ4F_createDecompressionContext
_LZ4F_decompress
_LZ4F_freeDecompressionContext
_LZ4F_getErrorName
_LZ4F_isError
_LZ4_compressBound
_LZ4_compress_HC
_LZ4_compress_default
_LZ4_decompress_safe
```

So, after reading through some of the code actually, in `Compression.cpp`, we handle different versioning in addition to users having/not having LZ4 available to them already. Thus, perhaps, it makes the most sense to actually not impose versioning requirements in this find module (or perhaps just version 1.3.0?). In fact, forcing version `1.7.0` would actually break users if `Compression.cpp` has some code paths which work just fine for `1.3.0` through `1.7.0`.

Thoughts?

yfeldblum: We can start by not having any requirements around versioning. They can be added later if they are actually needed.

JoeLoser: Sounds good -- just pushed up the change incorporating no minimum version requirements. I actually remembered to squash the commits first this time too 👍

PR #781, opened 2018-02-24
Fix DynamicConverter to work with vector<bool>

JoeLoser:
Introduce a specialization for `std::vector<bool>` as the existing
specialization for "other ranges" is not sufficient for binding the
proxy iterator (`bit_const_reference`) to an lvalue when iterating over
the `vector<bool>`.

I do not think you can simply change the type from `const auto&` to `auto&&` when iterating over the container for the "other ranges" specialization.  This is because we cannot bind an rvalue to an lvalue reference (which is the function argument type for our `construct` specializations).

See #752 and https://github.com/facebook/folly/pull/756 for more info.

### DIFF
--- a/folly/DynamicConverter.h
+++ b/folly/DynamicConverter.h
@@ -368,6 +368,24 @@ struct DynamicConstructor<std::pair<A, B>, void> {
   }
 };
 
+// vector<bool>
+template<>
+struct DynamicConstructor<std::vector<bool>, void>
+{
+    static dynamic construct(const std::vector<bool>& x)
+    {
+        dynamic d = dynamic::array;
+        // Intentionally specifying the type as bool here.
+        // std::vector<bool>'s iterators return a proxy which is a prvalue
+        // and hence cannot bind to an lvalue reference such as auto&
+        for(bool item : x)
+        {
+            d.push_back(toDynamic(item));
+        }
+        return d;
+    }
+};
+
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


folly/DynamicConverter.h
```diff
@@ -368,6 +369,25 @@ struct DynamicConstructor<std::pair<A, B>, void> {
   }
 };
 
+// vector<bool>
+template<class C>
+struct DynamicConstructor<C, typename std::enable_if<
+    std::is_same<C, std::vector<bool>>::value>::type>
```
yfeldblum: Could this just pattern-match against `vector<bool>` directly?

```lang=c++
template <>
struct DynamicConstructor<std::vector<bool>, void> {
```

Would that compile?

JoeLoser: Good point -- thanks for mentioning it! Yep, we can indeed use explicit template specialization here, which is cleaner than SFINAE. Allows us to remove the SFINAE from the "other ranges" specialization too.

JoeLoser:
Just added a simple unit test and squashed the commits. Feel free to import it internally again -- let me know if there is anything else I can do to get this one across the finish line. Now, things will build out of the box using Clang and `libc++`

JoeLoser:
Looks like the two Travis builds using GCC 4.9 are failing due to your recent `to_signed` change in 5ec7c712c65bed04c28193d970cb50bbe31c2dea; works fine in GCC 5.

PR #784, opened 2018-02-24
Run clang-tidy performance checks

JoeLoser:
Runs the several clang-tidy `performance-*` checks on the entire codebase and applies the auto fix-its

### DIFF
--- a/folly/compression/Compression.cpp
+++ b/folly/compression/Compression.cpp
@@ -60,6 +60,7 @@
 #include <folly/stop_watch.h>
 #include <algorithm>
 #include <unordered_set>
+#include <utility>
 
 using folly::io::compression::detail::dataStartsWithLE;
 using folly::io::compression::detail::prefixToStringLE;
@@ -69,7 +70,7 @@ namespace io {
 
 Codec::Codec(
     CodecType type,
-    Optional<int> level,
+    const Optional<int>& level,
     StringPiece name,
     bool counters)
     : type_(type) {
@@ -278,7 +279,7 @@ std::string Codec::doUncompressString(
     const StringPiece data,
     Optional<uint64_t> uncompressedLength) {
   const IOBuf inputBuffer{IOBuf::WRAP_BUFFER, data};
-  auto outputBuffer = doUncompress(&inputBuffer, uncompressedLength);
+  auto outputBuffer = doUncompress(&inputBuffer, std::move(uncompressedLength));
   std::string output;
   output.reserve(outputBuffer->computeChainDataLength());
   for (auto range : *outputBuffer) {
@@ -293,7 +294,7 @@ uint64_t Codec::maxCompressedLength(uint64_t uncompressedLength) const {
 
 Optional<uint64_t> Codec::getUncompressedLength(
     const folly::IOBuf* data,
-    Optional<uint64_t> uncompressedLength) const {
+    const Optional<uint64_t>& uncompressedLength) const {
   auto const compressedLength = data->computeChainDataLength();
   if (compressedLength == 0) {
     if (uncompressedLength.value_or(0) != 0) {
--- a/folly/executors/ThreadPoolExecutor.cpp
+++ b/folly/executors/ThreadPoolExecutor.cpp
@@ -178,7 +178,7 @@ ThreadPoolExecutor::PoolStats ThreadPoolExecutor::getPoolStats() {
   RWSpinLock::ReadHolder r{&threadListLock_};
   ThreadPoolExecutor::PoolStats stats;
   stats.threadCount = threadList_.get().size();
-  for (auto thread : threadList_.get()) {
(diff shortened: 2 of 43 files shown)


folly/compression/Compression.cpp
```diff
@@ -69,7 +70,7 @@ namespace io {
 
 Codec::Codec(
     CodecType type,
-    Optional<int> level,
+    const Optional<int>& level,
```
yfeldblum: This does not seem very helpful.

folly/executors/ThreadPoolExecutor.cpp
```diff
@@ -232,15 +232,15 @@ size_t ThreadPoolExecutor::StoppedThreadQueue::size() {
   return queue_.size();
 }
 
-void ThreadPoolExecutor::addObserver(std::shared_ptr<Observer> o) {
+void ThreadPoolExecutor::addObserver(const std::shared_ptr<Observer>& o) {
```
yfeldblum: This is wrong.

folly/synchronization/test/RWSpinLockTest.cpp
```diff
@@ -152,7 +152,8 @@ TYPED_TEST(RWSpinLockTest, ConcurrentTests) {
   srand(time(nullptr));
 
   std::vector<std::thread> threads;
-  for (int i = 0; i < FLAGS_num_threads; ++i) {
+  threads.reserve(FLAGS_num_threads);
+for (int i = 0; i < FLAGS_num_threads; ++i) {
```
yfeldblum: Bad formatting. And why bother here?

JoeLoser: I was mixed on the vector reserve checks. It is nearly no speed-up given the few numbers of `push_back` call in unit tests, typically. Happy to revert.

As for the formatting -- is it possible to get a `.clang-format` in the open source repo?

folly/test/TryTest.cpp
```diff
@@ -132,7 +132,7 @@ TEST(Try, ValueOverloads) {
 // Make sure we can copy Trys for copyable types
 TEST(Try, copy) {
   Try<int> t;
-  auto t2 = t;
+  const auto& t2 = t;
```
Orvid: This is wrong and changes what the test does.

folly/synchronization/test/RWSpinLockTest.cpp
Orvid: I intentionally didn't apply the vector-reserve or pass-by-const-ref passes when I ran clang-tidy over Folly a while back. The vast majority of the places where the vector-reserve pass changed things were in unit tests, where the added complexity (and in many cases, added duplication of constants), are not desired. I chose not to apply the pass-by-const-ref pass in the vast majority of places because it tried to change multiple places that I knew would break things.

folly/test/AHMIntStressTest.cpp
```diff
@@ -107,7 +107,8 @@ TEST(AHMIntStressTest, Test) {
   SCOPE_EXIT { delete objs; };
 
   std::vector<std::thread> threads;
-  for (int threadId = 0; threadId < 64; ++threadId) {
+  threads.reserve(64);
+for (int threadId = 0; threadId < 64; ++threadId) {
```
Orvid: This is an example of what I mean by the vector-reserve pass duplicating constants that really shouldn't be.

folly/compression/Compression.cpp
JoeLoser: Fair point. In general, it's just checking if a type is trivially copyable. If it is not, it will pass by `const&`. In this case, `Optional` has its own copy constructor. Probably best to just disable that check entirely as it will probably have more false positives than we would like.

folly/synchronization/test/RWSpinLockTest.cpp
JoeLoser: Thanks for the insights, @orvid. I wasn't aware someone had already run clang-tidy over the entire codebase.

I agree on both fronts with the vector-reserve pass and the pass-by-const-ref pass after looking at things a bit closer.

folly/test/AHMIntStressTest.cpp
JoeLoser: Definitely not desirable, agreed.

folly/test/TryTest.cpp
JoeLoser: Very valid. Seems like, in general, we wouldn't care much about clang-tidy over unit test code as higher risk for less value-add.

JoeLoser:
Going to close this PR.  After looking a bit closer at the changes, I think it does more harm than good. Thanks for sharing some insights, @Orvid and @yfeldblum

JoeLoser:
Going to close this PR. After looking a bit closer at the changes, I think it does more harm than good. Thanks for sharing some insights, @Orvid and @yfeldblum

Issue #753, opened 2018-02-05
cmake module for gflags on fc27


dotnwat:
on fedora 27 folly cmake build doesn't find the installed version of `gflags-devel` because the install cmake module doesn't create a target called `gflags`, but rather `gflags-shared`.

```diff
diff --git a/CMake/folly-deps.cmake b/CMake/folly-deps.cmake
index 0e687b98..70346531 100755
--- a/CMake/folly-deps.cmake
+++ b/CMake/folly-deps.cmake
@@ -26,7 +26,7 @@ find_package(gflags CONFIG QUIET)
 if (gflags_FOUND)
   message(STATUS "Found gflags from package config")
   set(FOLLY_HAVE_LIBGFLAGS ON)
-  set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags)
+  set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags-shared)
   list(APPEND CMAKE_REQUIRED_LIBRARIES ${GFLAGS_LIBRARIES})
   list(APPEND CMAKE_REQUIRED_INCLUDES ${GFLAGS_INCLUDE_DIR})
 else()
```

JoeLoser:
Perhaps it would make more sense to have a CMake `option` on a per-third-party-lib basis (where appropriate, e.g. not for header-only libs), which would tell the associated `FIND_LIBRARY` call to look for static or dynamic libs?

kornrunner:
I believe this is now sorted via #789 ?

yfeldblum:
Closing this issue because it is too old.

PR #789, opened 2018-02-28
CMake: folly-deps: gflags is gflags-shared on Fedora #753

kornrunner:
This targets Fedora specifically. More details are available in #753

I personally do not think this is the best way to handle this, but it's something (and works). Still, there might be a chance for this to get merged in? Thank you!

/cc @yfeldblum


CMake/folly-deps.cmake
```diff
@@ -26,7 +26,11 @@ find_package(gflags CONFIG QUIET)
 if (gflags_FOUND)
   message(STATUS "Found gflags from package config")
   set(FOLLY_HAVE_LIBGFLAGS ON)
-  set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags)
+  if(EXISTS "/etc/fedora-release")
```
JoeLoser: Is it like this for every Fedora version or only after some threshold?

CMake/folly-deps.cmake
```diff
@@ -26,7 +26,11 @@ find_package(gflags CONFIG QUIET)
...
+    set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags-shared)
```
JoeLoser: Is it called `gflags-shared` for every Fedora version or only after some threshold?

kornrunner: I've checked latest gflags packages for:
Fedora Core 27: gflags-2.1.2-5
Fedora Core 26: gflags-2.1.2-2
Fedora Core 25: gflags-2.1.2-1
In all of them it's called `gflags-shared`.

I haven't checked previous versions of packages or distributions. Current stable is FC27. Hope this helps. Thank you!

Orvid: `gflags-shared` is one of the targets GFlags can choose to export by default. This check can be done as 
```CMake
if (TARGET gflags-shared)
  set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags-shared)
elseif (TARGET gflags)
  set(FOLLY_SHINY_DEPENDENCIES ${FOLLY_SHINY_DEPENDENCIES} gflags)
else()
  message(FATAL_ERROR "Unable to determine the target name for the GFlags package.")
endif()
```

kornrunner: Awesome! That kind of check was something that I was actually looking for.

yfeldblum: Please do that.

kornrunner:
Updated PR

kornrunner:
Can we perhaps consider reverting this change? This has worked for Fedora 27 and earlier, with gflags 2.1.2 (and earlier).

It seems that newer editions of Fedora have changed this again (see https://github.com/facebookincubator/fizz/issues/20)

I'm not sure if we wish to keep BC or not, but current state is a show-stopper for newer versions.

Thank you!

PR #791, opened 2018-02-28
Fix unused variable error in AsyncUDPSocket

JoeLoser:
Summary:
Some OSes do not define `IP_MTU_DISCOVER` or `IPV6_MTU_DISCOVER`. As such, this would result in the function argument parameter `df` being unused, which is a warning that is then treated as an error.

Earlier today (2806dda49f25a238852f75a52225ce9ee0cbccc1), this logic was added about setting the DF which broke Mac builds.

### DIFF
--- a/folly/io/async/AsyncUDPSocket.cpp
+++ b/folly/io/async/AsyncUDPSocket.cpp
@@ -133,6 +133,7 @@ void AsyncUDPSocket::bind(const folly::SocketAddress& address) {
 }
 
 void AsyncUDPSocket::dontFragment(bool df) {
+  (void) df; // to avoid potential unused variable warning
 #ifdef IP_MTU_DISCOVER
   if (address().getFamily() == AF_INET) {
     int v4 = df ? IP_PMTUDISC_DO : IP_PMTUDISC_WANT;


KindDragon:
Same errors after this changes:
```
folly\io\async\AsyncUDPSocket.cpp(139): error C2065: 'IP_PMTUDISC_WANT': undeclared identifier
folly\io\async\AsyncUDPSocket.cpp(150): error C2065: 'IPV6_PMTUDISC_DO': undeclared identifier
folly\io\async\AsyncUDPSocket.cpp(150): error C2065: 'IPV6_PMTUDISC_WANT': undeclared identifier
```
Because `IP_MTU_DISCOVER` is defined on Windows, but not 'IP_PMTUDISC_WANT', 'IPV6_PMTUDISC_DO', 'IPV6_PMTUDISC_WANT'

JoeLoser:
@KindDragon valid point. This patch was mainly targeted at OSes which did not even define `IP_MTU_DISCOVER`. Perhaps you can submit a patch after the`#ifdef IP_MTU_DISCOVER` before using those undefined macros such as  `IP_PMTUDISC_DO` and `IP_PMTUDISC_WANT` in the ternary expression?

KindDragon:
It supports [`IP_PMTUDISC_DO`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738586(v=vs.85).aspx) and Windows support [`IP_PMTUDISC_DO`](https://msdn.microsoft.com/ru-ru/library/windows/desktop/ms738574(v=vs.85).aspx) instead of `IPV6_PMTUDISC_DO`

PR #792, opened 2018-02-28
[CMake] Do not assume subdirectory for finding libiberty.h

JoeLoser:
Summary:
Do not assume `libiberty.h` is contained within a subdirectory called `libiberty`.

In general, it is probably best not to assume a certain directory structure as it will vary across package manager, OS, etc. While users can certainly set `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` accordingly to make things work, it's not ideal. It is nicer to just modify the find module scripts to support more users out of the box.

### DIFF
--- a/CMake/FindLibiberty.cmake
+++ b/CMake/FindLibiberty.cmake
@@ -1,4 +1,4 @@
-find_path(LIBIBERTY_INCLUDE_DIR NAMES libiberty/libiberty.h)
+find_path(LIBIBERTY_INCLUDE_DIR NAMES libiberty.h)
 mark_as_advanced(LIBIBERTY_INCLUDE_DIR)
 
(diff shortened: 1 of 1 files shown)


JoeLoser:
ping @yfeldblum

simpkins:
libiberty.h is installed at /usr/include/libiberty/libiberty.h on Ubuntu systems.  This probably should be updated to use PATH_SUFFIXES as done for dwarf.h in 48d6f3144268

JoeLoser:
@simpkins good point. I made the same mistake with `libdwarf` as I did here. 

I think it's slightly more involved than your approach in 48d6f31 since `libiberty.h` can be shipped with binutils or GCC. When I get some time, I'll work on a patch and resubmit another PR. Going to close this one for now.

PR #804, opened 2018-03-19
Missing installation of detail/Singleton.h

asierguti:
Makefile does not install detail/Singleton.h, which is needed by proxygen


yfeldblum:
Sorry about that!

PR #805, opened 2018-03-22
find_package(gflags CONFIG QUIET) didn't work in Ubuntu.

kumagi:
in Linux environment, `find_package(gflass)` searches `Findgflags.cmake`.
And it cannot find  https://github.com/facebook/folly/blob/master/CMake/FindGFlags.cmake , because linux is case-sensitive environment in filenames and cause error. 

```
CMake Error at CMake/folly-deps.cmake:34 (message):
  Unable to determine the target name for the GFlags package.
  CMakeLists.txt:65 (include)
```

For linux programmers, this line should find `FindGFlags.cmake` file which exists in the same directory.


kumagi:
https://github.com/facebook/folly/blob/master/CMake/folly-deps.cmake#L48
`FindGLog.cmake` seems to have the same problem.

kumagi:
Hi, could someone review it?

PR #807, opened 2018-03-27
Fix F14TestUtil printing

JoeLoser:
Summary:
There is a missing `#include <ostream`.  Add it.

### DIFF
--- a/folly/container/test/F14TestUtil.h
+++ b/folly/container/test/F14TestUtil.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <cstddef>
+#include <ostream>
 #include <vector>
 
 #include <folly/Demangle.h>


JoeLoser:
Looks like one test case threw an exception in the `fibers` test for GCC 4.9. Is this expected or a random failure that is noted internally to chase after? See the Travis logs for the GCC 4.9 build on this branch. 

// CC @yfeldblum

yfeldblum:
Why are these invalid overloads?

JoeLoser:
@yfeldblum I misspoke after looking into the issue some more. As expected, we wish to call the free function overloads for `basic_stream<char>` which have specialization for inserting a `const char*`. Here, the `const char[N]` will decay to a pointer and call the right overload.

The real problem comes down to just not having that overload in play for resolution because we are missing the `ostream` header. Simply adding
```
#include <ostream>
```
resolves the compiler errors that were of the form:
```
In file included from pathToFolly/folly/container/test/F14SetTest.cpp:18:
pathToFolly/folly/container/test/F14TestUtil.h:35:6: error: invalid operands to binary expression ('std::ostream' (aka 'basic_ostream<char>') and 'const char *')
  xo << "[";
  ~~ ^  ~~~
pathToFolly/folly/container/test/F14TestUtil.h:34:15: note: candidate function not viable: no known conversion from 'const char [2]' to 'const folly::f14::Histo' for 2nd argument
std::ostream& operator<<(std::ostream& xo, Histo const& histo) {
              ^
/usr/local/Cellar/llvm/6.0.0/include/c++/v1/memory:5334:1: note: candidate template ignored: could not match 'shared_ptr<type-parameter-0-2>' against 'char const[2]'
operator<<(basic_ostream<_CharT, _Traits>& __os, shared_ptr<_Yp> const& __p);
^
/usr/local/Cellar/llvm/6.0.0/include/c++/v1/string:3965:1: note: candidate template ignored: could not match 'basic_string<type-parameter-0-0, type-parameter-0-1, type-parameter-0-2>' against 'char const[2]'
operator<<(basic_ostream<_CharT, _Traits>& __os,
```

To reproduce the problem, just try building the `f14_map_test` target standalone. I ran into the problem just when building all of Folly originally rather than one specific target or test. This is a smaller, reproducible example, though. The `f14_set_test` is another similar small reproducible example of this problem.

My guess is that `<ostream>` was getting transitively included depending on your specific standard library implementation which unfortunately has fallen victim to exposing certain headers in their own standard library implementation.  In general, probably best to have a file be able to be compiled stand-alone without relying on transitive includes.

PR #808, opened 2018-03-28
Remove backport of index_sequence and integer_sequence

JoeLoser:
Summary:
`std::integer_sequence` and `std::index_sequence` were backported to
satisfy the requirement of being able to use GCC 4.8. Now that the min GCC
version is 4.9 which implements these functions, we can remove our own
implementation.

Concerns:
Based on @yfeldblum's [commit](https://github.com/facebook/folly/commit/890625b2c781b779f65a835e7c7c610c95f010c0) in July 2017, we may be worried about older GCC versions with their QOI in stamping out too many templates instantiations for these functions. Without seeing all the call sites of users in internal uses, it's hard to know how big of a worry this should be. 

Has anyone done any study using `-ftime-report` for the most-used applications for how much time is spent parsing and instantiating templates given a FB implementation vs off-the-shelf implementation? I suspect it was/is done with the `fatal` library.

Would we prefer to keep the backported divide-and-conquer implementation around until the min GCC version is 6 (have not verified that myself, but was stated in the commit message above) to avoid `O(N)` implementations of these functions?  Or, would we prefer to just use the standard library implementation and get corresponding vendors to improve the QOI for specific GCC versions (well, backport their improved implementations rather in all reality)?


yfeldblum:
What motivates this PR?

> Would we prefer to keep the backported divide-and-conquer implementation around until the min GCC version is 6 (have not verified that myself, but was stated in the commit message above) to avoid O(N) implementations of these functions?

The O(N) by itself is not the problem, because this is always used to implement some other O(N) algorithm. The real problem is the O(N) recursion, because unmemoized recursion is expensive.

Until we jump to a baseline libstdc++ with O(lg N) recursion, we can keep the one in folly.

Note that the version in folly (and fatal) performs fewer recursions and creates fewer template instantiations than the optimized version in libstdc++ for far-from-round parameters, but they are both O(lg N) recursion depth and we have not attempted to measure the in-practice compile-time difference.

PR #811, opened 2018-03-30
Fix some `CMAKE_REQUIRED_FLAGS` stuff

aw1621107:
This fixes two issues related to CMAKE_REQUIRED_FLAGS

----

CMAKE_REQUIRED_FLAGS, unlike the other CMAKE_REQUIRED_* variables, is a
string, not a list. A list with more than one item in
CMAKE_REQUIRED_FLAGS results in a semicolon showing up in the middle of
the command line for at least check_symbol_exists and check_type_size
calls, prematurely terminating the compilation command and resulting in
potentially incorrect CMake test failures.

For example, CMake fails to determine the size of __int128 on macOS
10.13.3, with Apple Clang 9 (clang-900.0.39.2), even though __int128 is
supported for this platform, as the following C program

    int main() {
        __int128 i;
    }

compiles successfully with the command "clang test.c". CMake fails to
compile a program with __int128 for an entirely different reason:

    Building CXX object CMakeFiles/cmTC_d7e3f.dir/INT128_SIZE.cpp.o
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++  -DGFLAGS_IS_A_DLL=0 -isystem /usr/local/include -I/usr/local/opt/openssl@1.1/include  -fsanitize=address,undefined -std=gnu++14;-Werror=unknown-warning-option   -o CMakeFiles/cmTC_d7e3f.dir/INT128_SIZE.cpp.o -c /Users/awang/code/CRES/scc-analysis/build-debug/CMakeFiles/CheckTypeSize/INT128_SIZE.cpp
    clang: error: no input files
    /bin/sh: -Werror=unknown-warning-option: command not found
    make[1]: *** [CMakeFiles/cmTC_d7e3f.dir/INT128_SIZE.cpp.o] Error 127
    make: *** [cmTC_d7e3f/fast] Error 2

Using set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} <new_flag>")
avoids this problem.

----

CMAKE_REQUIRED_FLAGS is given to both C and C++ compiler invocations.
The C++-specific flag -std=gnu++14 causes compilation to fail on Apple
LLVM 9.1.0 (clang-902.0.39.2) with errors along the lines of:

    Building C object CMakeFiles/cmTC_af532.dir/CheckSymbolExists.c.o
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -DGFLAGS_IS_A_DLL=0 -isystem /usr/local/include -I/usr/local/opt/openssl@1.1/include  -std=gnu++14   -o CMakeFiles/cmTC_af532.dir/CheckSymbolExists.c.o   -c /Users/awang/code/CRES/scc-analysis/build-debug/CMakeFiles/CMakeTmp/CheckSymbolExists.c
    error: invalid argument '-std=gnu++14' not allowed with 'C/ObjC'
    make[1]: *** [CMakeFiles/cmTC_af532.dir/CheckSymbolExists.c.o] Error 1
    make: *** [cmTC_af532/fast] Error 2

The affected functions are:

  - pthread_atfork
  - memrchr
  - preadv
  - pwritev
  - clock_gettime
  - cplus_demangle_v3_callback
  - malloc_usable_size

Of these, only pthread_atfork and clock_gettime were detected as being
absent when they were actually present.

The failure to correctly identify the presence of pthread_atfork, at
least, can cause Folly to fail to build on macOS because Folly requires
pthread_atfork for macOS but thinks it is absent due to the failed test.

### DIFF
--- a/CMake/FollyCompilerUnix.cmake
+++ b/CMake/FollyCompilerUnix.cmake
@@ -2,7 +2,7 @@ set(CMAKE_CXX_FLAGS_COMMON "-g -Wall -Wextra")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_COMMON}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_COMMON} -O3")
 
-list(APPEND CMAKE_REQUIRED_FLAGS -std=gnu++14)
+set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=gnu++14")
 function(apply_folly_compile_options_to_target THETARGET)
   target_compile_definitions(${THETARGET}
     PRIVATE
--- a/CMake/FollyConfigChecks.cmake
+++ b/CMake/FollyConfigChecks.cmake
@@ -19,7 +19,8 @@ if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
     -Werror=unknown-warning-option
     COMPILER_HAS_UNKNOWN_WARNING_OPTION)
   if (COMPILER_HAS_UNKNOWN_WARNING_OPTION)
-    list(APPEND CMAKE_REQUIRED_FLAGS -Werror=unknown-warning-option)
+    set(CMAKE_REQUIRED_FLAGS
+      "${CMAKE_REQUIRED_FLAGS} -Werror=unknown-warning-option")
   endif()
 
   CHECK_CXX_COMPILER_FLAG(-Wshadow-local COMPILER_HAS_W_SHADOW_LOCAL)
@@ -61,6 +62,13 @@ if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
   endif()
 endif()
 
+set(FOLLY_ORIGINAL_CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS}")
+string(REGEX REPLACE
+  "-std=(c|gnu)\\+\\+.."
+  ""
+  CMAKE_REQUIRED_FLAGS
+  "${CMAKE_REQUIRED_FLAGS}")
+
 check_symbol_exists(pthread_atfork pthread.h FOLLY_HAVE_PTHREAD_ATFORK)
 
 # Unfortunately check_symbol_exists() does not work for memrchr():
@@ -70,11 +78,14 @@ check_symbol_exists(preadv sys/uio.h FOLLY_HAVE_PREADV)
 check_symbol_exists(pwritev sys/uio.h FOLLY_HAVE_PWRITEV)
 check_symbol_exists(clock_gettime time.h FOLLY_HAVE_CLOCK_GETTIME)
 
+
 check_function_exists(
   cplus_demangle_v3_callback
   FOLLY_HAVE_CPLUS_DEMANGLE_V3_CALLBACK
 )
 check_function_exists(malloc_usable_size FOLLY_HAVE_MALLOC_USABLE_SIZE)
+
+set(CMAKE_REQUIRED_FLAGS "${FOLLY_ORIGINAL_CMAKE_REQUIRED_FLAGS}")
 
 check_cxx_source_compiles("
   #pragma GCC diagnostic error \"-Wattributes\"


CMake/FollyCompilerUnix.cmake
```diff
@@ -2,7 +2,6 @@ set(CMAKE_CXX_FLAGS_COMMON "-g -Wall -Wextra")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_COMMON}")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_COMMON} -O3")
 
-list(APPEND CMAKE_REQUIRED_FLAGS -std=gnu++14)
```
simpkins: It looks like removing this breaks a number of the compilation checks on Linux.
For instance, the FOLLY_HAVE_WEAK_SYMBOLS and FOLLY_HAVE_LINUX_VDSO checks incorrectly fail with this change.  I haven't looked too closely but I suspect they are failing because they use `nullptr` in the checks, and require C++ 11 or greater functionality to compile.

aw1621107: Whoops... Apparently I wasn't reading `CMakeError.log` closely enough, because those tests fail for the same reason on my machine.

The other option I was playing with when writing this PR was using `string(REGEX REPLACE)` right before the `check_symbol_exists` call to strip out C++ version flags. I thought that was pretty fragile, though, so I abandoned that approach. Do you think it's worth another shot? Or got any better ideas?

Also, for cases like this, do you want me to use a regular revert + new commit, or rewrite things so the bad commit never happened?

Orvid: All of the commits will get squashed, and the commit message will be the message put in for the Pull Request, so the individual commits don't actually matter.

aw1621107:
Looks like the `check_*_exists` calls are spread out into three places. I could add `string(REGEX REPLACE)` around those, or maybe create a wrapper `check_c_*_exists` function that wraps that up. Is one of those approaches preferred over the other?

leolchat:
Why not use `CXX_STANDARD_REQUIRED` ?

aw1621107:
@leolchat Doesn't that force that C++ standard for the entire project? I want to remove the `-std=c++14` flag, so setting `CXX_STANDARD_REQUIRED` wouldn't help. Or do you mean use it instead of `CMAKE_REQUIRED_FLAGS`? I don't know why `CMAKE_REQUIRED_FLAGS` was used originally, but I can try (un)setting `CMAKE_REQUIRED_FLAGS` and see what happens

leolchat:
It can be per target
https://cmake.org/cmake/help/v3.10/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

aw1621107:
@leolchat Ah, forgot about setting properties that way. It seems `-std=gnu++14` is already being passed on a per-target basis by a target_compile_options call, though:

https://github.com/facebook/folly/blob/3ce2b2177324ef8ac842995ceeddebc57a15aaff/CMake/FollyCompilerUnix.cmake#L13-L20

And I was under the impression that the `check_*_exists` calls don't generate targets. They all also share the `CMAKE_REQUIRED_FLAGS` variable, so it seems some kind of setting/resetting would be necessary

mohd-akram:
Can be fixed with this patch:

```diff
--- CMake/FollyConfigChecks.cmake.orig	2018-06-02 01:57:39.000000000 +0400
+++ CMake/FollyConfigChecks.cmake	2018-06-02 01:58:59.000000000 +0400
@@ -61,6 +61,7 @@
   endif()
 endif()
 
+list(REMOVE_ITEM CMAKE_REQUIRED_FLAGS -std=gnu++14)
 check_symbol_exists(pthread_atfork pthread.h FOLLY_HAVE_PTHREAD_ATFORK)
 
 # Unfortunately check_symbol_exists() does not work for memrchr():
@@ -75,6 +76,7 @@
   FOLLY_HAVE_CPLUS_DEMANGLE_V3_CALLBACK
 )
 check_function_exists(malloc_usable_size FOLLY_HAVE_MALLOC_USABLE_SIZE)
+list(APPEND CMAKE_REQUIRED_FLAGS -std=gnu++14)
 
 check_cxx_source_compiles("
   #pragma GCC diagnostic error \"-Wattributes\"
```

aw1621107:
@mohd-akram I'm somewhat surprised that works; as noted in the [first commit](https://github.com/facebook/folly/pull/811/commits/1d4fb7f2b77237599f309cc24ad0472b1576d87b) of this PR, `CMAKE_REQUIRED_FLAGS` is a string, not a list, so using `list(REMOVE_ITEM)` could have some interesting interactions (not sure how reliable the behavior would be, either). Could make sense if it's CMake stringly-typed shenanigans, though.

mohd-akram:
I don't know much about CMake really, I just followed the example of the existing code and found your PR afterwards. I also had to use another patch (see macports/macports-ports#1933) due to weird behavior by CMake. Maybe you can add that one to your PR too.

aw1621107:
@mohd-akram Funny; I don't remember having issues related to your other patch. Could be because I'm on macOS using `make`? No clue. In any case looks like someone already submitted a PR in #862.

aw1621107:
Alright, *finally* got around to fixing the PR. Configures and builds fine on my machine, at least.

yfeldblum:
Please squash the commits and update the PR summary (the PR summary is what actually gets used as the commit message, not the message within the git commit).

aw1621107:
@yfeldblum Ah, @Orvid's comment made me think that the squash happened automatically. I'll get things squashed.

And I think I updated the PR summary? That's the first comment, right?

PR #815, opened 2018-04-06
Fix CMakeLists.txt to not exclude everything with 'build' in pathname

leolchat:
The cmake globbing then excluding strategy implies that if the directory
in which we are compiling has a folder named `build` on its path like `/home/alice/project/build/folly/...`, everything breaks.
Note that it is the same for `tools` and `test`, I also fixed the tools one, but the test once has many files.... Using a file list would be the real solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ REMOVE_MATCHES_FROM_LISTS(files hfiles
     "^${FOLLY_DIR}/experimental/hazptr/example/"
     "^${FOLLY_DIR}/experimental/logging/example/"
     "^${FOLLY_DIR}/futures/exercises/"
-    "^${FOLLY_DIR}.*/test/"
+    "^${FOLLY_DIR}/(.*/)?test/"
     "^${FOLLY_DIR}/tools/"
     "Benchmark.cpp$"
     "Test.cpp$"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,14 +90,14 @@ auto_sources(hfiles "*.h" "RECURSE" "${FOLLY_DIR}")
 # library sources.  Test sources are listed separately below.
 REMOVE_MATCHES_FROM_LISTS(files hfiles
   MATCHES
-    "/build/"
-    "/experimental/exception_tracer/"
-    "/experimental/hazptr/bench/"
-    "/experimental/hazptr/example/"
-    "/experimental/logging/example/"
-    "/futures/exercises/"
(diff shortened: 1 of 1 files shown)


Orvid:
Where is this causing issues? The glob is specifically for building Folly's base library, and things in a `tools/` directory are binaries to run, and the build directory is the same way.

leolchat:
This will fail with an empty list of files to compile:
```
cd /tmp
mkdir build
cd build
git clone https://github.com/facebook/folly.git
cd folly
cmake .
```

Compiling out of sources would fail too as long as the folly source folder path has a folder named build.

leolchat:
So what is the decision on that? are you going to merge it? Or maybe you want to do a more complete fix of the issues related to globbing then excluding? Or you consider not a big enough issue?

Orvid:
I'd missed your response because something's up with my github notifications :( 

I think the most correct solution here (assuming CMake allows it) would be to not use the absolute paths of the files while filtering.

leolchat:
Maybe you could do match `${CMAKE_CURRENT_LIST_DIR}/build` or something like that, not sure.

leolchat:
your implementation in the end (the function `REMOVE_MATCHES_FROM_LISTS `) uses `if(<path> MATCHES test)` but MATCHES can take a regexp so `${CMAKE_CURRENT_LIST_DIR}/*/test/` would probably work.

leolchat:
Should be bulletproof now.

CMakeLists.txt
```diff
@@ -90,14 +90,14 @@ auto_sources(hfiles "*.h" "RECURSE" "${FOLLY_DIR}")
...
-    "/futures/exercises/"
-    "/test/"
-    "/tools/"
+    "^${FOLLY_DIR}/build/"
+    "^${FOLLY_DIR}/experimental/exception_tracer/"
+    "^${FOLLY_DIR}/experimental/hazptr/bench/"
+    "^${FOLLY_DIR}/experimental/hazptr/example/"
+    "^${FOLLY_DIR}/experimental/logging/example/"
+    "^${FOLLY_DIR}/futures/exercises/"
+    "^${FOLLY_DIR}.*/test/"
```
yfeldblum: Should this be `}/.*/`?

leolchat: no because there is `${FOLLY_DIR}/test/` to match

yfeldblum: Gotcha. Consider `}/(.*/)?`?

PR #835, opened 2018-05-03
Remove redundant static_cast

JoeLoser:
Summary: remove redundant `static_cast`.

All failing builds are just the cause of timeout from `tdigest` tests. Seems to be a common theme of exceeding 120 sec timeouts for the open-source Travis runs.


yfeldblum:
Not redundant - we want to forward the arguments.

PR #840, opened 2018-05-05
 Replace std::unique_lock with std::lock_guard.

SimonCqk:
std::lock_guard has a better performance than std::unique_lock in scoped-lock.


yfeldblum:
It looks like the latest compilers (well, gcc6+) generate identical code for the two versions. But gcc5- generates slightly smaller code for `std::lock_guard`.

PR #871, opened 2018-06-18
Added folly init and folly logging to cmake

orishu:
The two libraries libfollyinit and libfollylogging were not properly defined in
CMakeLists.txt as libraries that get built and installed.  The issue caused
build failures for mcrouter opensource project.  Patching folly before
building it fixed the issue.
Related mcrouter packaging pull request: https://github.com/facebook/mcrouter/pull/260

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,7 +268,24 @@ source_group("" FILES ${CMAKE_CURRENT_BINARY_DIR}/folly_dep.cpp)
 
 target_link_libraries(folly PUBLIC folly_deps)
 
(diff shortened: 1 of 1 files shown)


danobi:
I was under the impression the cmake build implicitly links libfollyinit and libfollylogging inside libfolly. The autotools build, however, does separate libs. I've since put up a PR to remedy this: #872

orishu:
@danobi I think you're right. I'm working on packaging facebook/mcrouter which statically links to folly, it incorrectly had "-lfollyinit" and "-lfollylogging" flags, so I should make the fix there. Closing.

Issue #882, opened 2018-06-29
Compression.cpp fails to compile because of Windows' min/max macros


kyllingstad:
I am trying to compile Folly using Visual Studio 2017 (15.7.4), and am getting the following warnings and error messages from `folly/compression/Compression.cpp`:
```
1>Compression.cpp
1>c:\[redacted]\folly\conv.h(752): warning C4002: too many actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1227): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1228): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1229): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1236): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\conv.h(1255): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1256): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1276): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1277): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\conv.h(1284): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\conv.h(1293): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\random.h(58): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\random.h(59): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\random.h(61): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\random.h(62): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\random.h(83): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\random.h(84): warning C4003: not enough actual parameters for macro 'min'
1>c:\[redacted]\folly\random.h(87): warning C4003: not enough actual parameters for macro 'max'
1>c:\[redacted]\folly\random.h(88): warning C4003: not enough actual parameters for macro 'max'
[...]

PR #884, opened 2018-07-09
Prevent bzlib.h from pulling in min and max macros

kyllingstad:
This fixes issue #882. Ping @terrelln.

### DIFF
--- a/folly/compression/Compression.cpp
+++ b/folly/compression/Compression.cpp
@@ -44,6 +44,7 @@
 #endif
 
 #if FOLLY_HAVE_LIBBZ2
+#include <folly/portability/Windows.h>
 #include <bzlib.h>
 #endif
 


folly/compression/Compression.cpp
```diff
@@ -44,6 +44,9 @@
 #endif
 
 #if FOLLY_HAVE_LIBBZ2
+#if defined(_WIN32) && !defined(NOMINMAX)
+#define NOMINMAX 1
+#endif
 #include <bzlib.h>
```
Orvid: There are more things than just `NOMINMAX` that need to be handled with `Windows.h`, so just include `<folly/portability/Windows.h>` instead. It also means you don't have to guard it with `#ifdef _WIN32` anymore either :)

PR #885, opened 2018-07-11
fix zstd macro

foreverhy:
detect FOLLY_HAVE_LIBZSTD by true/false


terrelln:
Thanks for the fix @foreverhy!

PR #888, opened 2018-07-26
throw std::bad_alloc if memory can't be allocated

karasikov:
Bug fix: call `reserve()` trying to allocate more memory than available


yfeldblum:
Hi, I can import this fix once the CLA has been signed.

PR #899, opened 2018-08-01
IP_RECVERR is not defined on Mac

JoeLoser:
Problem:
- `IP_RECVERR` symbol is not defined on Mac OS.
- A test expects this symbol to be found. But, in the case on Mac OS, the symbol is not found and hence will cause a hard compiler error.

Solution:
- `ifdef` the portion of the test that is relying on this symbol.


folly/io/async/test/AsyncUDPSocketTest.cpp
```diff
@@ -530,6 +530,7 @@ TEST_F(AsyncUDPSocketTest, TestErrToNonExistentServer) {
   socket_->setErrMessageCallback(&err);
   folly::SocketAddress addr("127.0.0.1", 10000);
   bool errRecvd = false;
+#ifdef IP_RECVERR
```
yfeldblum: Guard this with `FOLLY_HAVE_MSG_ERRQUEUE` instead?

yfeldblum:
Thanks!

PR #900, opened 2018-08-01
AtomicSharedPtr-detail.h does not work with libc++

JoeLoser:
Problem:
- There are several tests enabled which rely on using `shared_ptr_internals` found in `folly/concurrency/detail/AtomicSharedPtr-detail.h` and the implementation only works for libstdc++, not libc++.
- This results in a hard error when compiling with Clang with libc++.

Solution:
- Only run the tests when not using libc++ to prevent users from getting
  a hard error if they are using libc++.

Note:
- A longer-term solution would be to implement the similar `shared_ptr_internals` with the libc++ types.

### DIFF
--- a/CMake/FollyConfigChecks.cmake
+++ b/CMake/FollyConfigChecks.cmake
@@ -184,6 +184,15 @@ check_cxx_source_compiles("
   FOLLY_USE_LIBCPP
 )
 
+check_cxx_source_compiles("
+  #include <type_traits>
+  #if !__GLIBCXX__
+  #error No libstdc++
+  #endif
+  int main() { return 0; }"
+  FOLLY_USE_LIBSTDCPP
+)
+
 check_cxx_source_runs("
   #include <string.h>
   #include <errno.h>
--- a/folly/concurrency/test/AtomicSharedPtrPerformance.cpp
+++ b/folly/concurrency/test/AtomicSharedPtrPerformance.cpp
@@ -13,6 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// AtomicSharedPtr-detail.h only works with libstdc++, so skip these tests for
+// other vendors
+#ifdef FOLLY_USE_LIBSTDCPP
+
 #include <folly/concurrency/AtomicSharedPtr.h>
 
 #include <sys/time.h>
@@ -225,3 +230,11 @@ int main(int, char**) {
   runSuite<folly::atomic_shared_ptr<int>>();
   return 0;
 }
+
+#else // #ifdef FOLLY_USE_LIBSTDCPP
+
+int main(int, char**) {
+  return 1;
+}
+
+#endif // #ifdef FOLLY_USE_LIBSTDCPP
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -372,6 +372,12 @@ constexpr auto kIsLibcpp = true;
 constexpr auto kIsLibcpp = false;
 #endif
 
+#if FOLLY_USE_LIBSTDCPP
+constexpr auto kIsLibstdcpp = true;
+#else
+constexpr auto kIsLibstdcpp = false;
+#endif
+
 #if _MSC_VER
 constexpr auto kMscVer = _MSC_VER;
 #else
--- a/folly/concurrency/test/AtomicSharedPtrTest.cpp
+++ b/folly/concurrency/test/AtomicSharedPtrTest.cpp
@@ -13,6 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// AtomicSharedPtr-detail.h only works with libstdc++, so skip these tests for
+// other vendors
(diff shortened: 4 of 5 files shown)


folly/concurrency/test/AtomicSharedPtrPerformance.cpp
```diff
@@ -13,6 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// AtomicSharedPtr-detail.h does not work with libc++, so skip these tests for libc++
+#include <folly/Portability.h>
+#if not defined(FOLLY_USE_LIBCPP)
```
yfeldblum: `if !defined` or `ifndef`?

JoeLoser: Switched to `#ifndef` -- sorry about that.

JoeLoser:
@yfeldblum @djwatson 

I considered using generator expressions in the top-level `CMakeLists.txt` to conditionally add the tests based on if someone using `libc++`, but I was worried how this would tie into the internal Buck builds. However, it's worth noting that this problem does not exist just for tests. 

The existing problem is exposed anytime someone simply has an unconditional include of `AtomicSharedPtr-detail.h` since the libstdc++ type names won't be found if someone is using libc++ (such as https://github.com/facebook/folly/blob/master/folly/concurrency/detail/AtomicSharedPtr-detail.h#L39). So, as you know, the approach I took here isn't scalable -- it's just a short-term workaround.

I briefly looked at both vendors implementations last weekend and it didn't seem like it would be *too* bad to implement the `shared_ptr_internals` to be compatible with libc++. If I recall correctly, I think the tricky part will be how `libc++` templates on the `deleter` all the way down whereas `libstdc++` provides you with `std::_Sp_make_shared_tag` as used [here](https://github.com/facebook/folly/blob/master/folly/concurrency/detail/AtomicSharedPtr-detail.h#L81)

yfeldblum:
Note that is also cpplib (the stdlib used with msvc) to consider.

djwatson:
@yfeldblum This seems fine as a quick fix to me.  It's header only, so for the build itself, we only need to worry about not building tests.

It would be nice to provide a fallback lock + shared_ptr (AtomicSharedPtr doesn't guarantee lock-freedom already).   Happy to have libc or cpplib also if someone wants to provide them.

JoeLoser:
@djwatson good point about the header-only.  If we're going to proceed with this short-term fix, I should probably take @yfeldblum's advice and guard against cpplib too. I'll do that tonight.

Agreed on the fallback/more vendor implementations.

JoeLoser:
@yfeldblum @djwatson just pushed up a change to account for other vendors.

- Add CMake configure check for using `libstdc++`. Call it `FOLLY_USE_LIBSTDCPP`. This is set to `1` if using `libstdc++`, 0 otherwise.
- Define an appropriate `constexpr` variable to capture this macro value in `folly/Portability.h`. There are no uses of this quite yet, but I can imagine some in the future.
- Use `FOLLY_USE_LIBSTDCPP` to conditionally enable these tests to account for other vendors (MSVC, Clang with libc++, ICC, etc) not being supported in `shared_ptr_internals.`

JoeLoser:
Ping @yfeldblum @djwatson. Anything else I can do to help get this landed?

folly/concurrency/test/AtomicSharedPtrPerformance.cpp
```diff
@@ -13,6 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// AtomicSharedPtr-detail.h only works with libstdc++, so skip these tests for other vendors
```
yfeldblum: Use an 80-column ruler.

(Saves me the step of auto-formatting this after import.)

CMake/FollyConfigChecks.cmake
```diff
@@ -184,6 +184,15 @@ check_cxx_source_compiles("
   FOLLY_USE_LIBCPP
 )
 
+check_cxx_source_compiles("
+  #include <type_traits>
+  #if __GLIBCXX__
```
yfeldblum: Is this test backwards?

folly/concurrency/test/AtomicSharedPtrPerformance.cpp
```diff
@@ -225,3 +229,4 @@ int main(int, char**) {
   runSuite<folly::atomic_shared_ptr<int>>();
   return 0;
 }
+#endif // #ifdef FOLLY_USE_LIBSTDCPP
```
yfeldblum:
```
  // ...
}

#else // #ifdef FOLLY_USE_LIBSTDCPP

int main(int, char**) {
  return 1;
}

#endif // #ifdef FOLLY_USE_LIBSTDCPP
```

JoeLoser:
@yfeldblum thanks for the careful review with the mistake in the CMake check. I've fixed the code changes requested, so should be ready to re-review. Thanks!

PR #904, opened 2018-08-05
Remove unnecessary executable bit on files

JoeLoser:
Summary:
- Several files have the executable bit set and do not need to.


Orvid:
> Several files

More specifically, the files that were originally created from Windows (by me) have the executable bit set :)

Issue #897, opened 2018-08-01
[bug][regression] Future InterruptHandler lost when using `then`


leolchat:
I have a function `cancelable` returning a future whose corresponding promise has an interrup handler attached (which put the exception in the promise), so that the basic test works:
```
  folly::Future<folly::Unit> f = cancelable();
  f.cancel();
  ASSERT_TRUE(f.hasException());
```

While this version stopped working (It was in an earlier version) :
```
  folly::Future<folly::Unit> f = folly::makeFuture().then([&]{return cancelable();});
  f.cancel();
  ASSERT_TRUE(f.hasException());
}
```

leolchat:
A way of testing this behavior directly is proposed in #906.

PR #906, opened 2018-08-07
Add test to check propagation of interrupt

leolchat:
This kind of code used to work and does not anymore (See #897 )
It seems that the implementation of `thenImplementation` using `stealPromise` is the root of that issue.

This test also raises the question of the actual semantics of `cancel` (and related interrupt handling) which is really poorly described in the documentation and code.


leolchat:
Not sure what imported pull request means.. Any human feedback here?

PR #909, opened 2018-08-12
Use #ifdef OPENSSL_IS_BORINGSSL instead of #if

OleksiyA:
In some cases OPENSSL_IS_BORINGSSL can be defined as simple as:

#define OPENSSL_IS_BORINGSSL

or in copts:

-DOPENSSL_IS_BORINGSSL

this causing folly to fail compilation.

### DIFF
--- a/folly/portability/OpenSSL.cpp
+++ b/folly/portability/OpenSSL.cpp
@@ -22,7 +22,7 @@ namespace folly {
 namespace portability {
 namespace ssl {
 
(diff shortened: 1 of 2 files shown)


yfeldblum:
Hi @OleksiyA, I can import this once the CLA situation is resolved.

stevezhou6:
Which version of boringss to use? Thanks for the help.

PR #916, opened 2018-08-25
maxReadsPerEvent_ SHOULD BE initialized to 0

hidva:
AsyncSocket.setMaxReadsPerEvent() says:
> The default is **unlimited**, but callers can use this method to limit the amount of data read from the socket per event loop iteration.

### DIFF
--- a/folly/io/async/AsyncSocket.cpp
+++ b/folly/io/async/AsyncSocket.cpp
@@ -343,7 +343,7 @@ void AsyncSocket::init() {
   eventFlags_ = EventHandler::NONE;
   fd_ = -1;
   sendTimeout_ = 0;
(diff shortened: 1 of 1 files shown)


djwatson:
16 is a much better default than unlimited.  Let's change the documentation instead?

hidva:
I think it's better to follow the documentation. After all, if the user needs 16 or other values, he might have explicitly called `setMaxReadsPerEvent()`.

djwatson:
Unlimited has the danger of blocking the event loop indefinitely. 

> After all, if the user needs 16 or other values, he might have explicitly called setMaxReadsPerEvent().

This isn't a real argument, since the user can always set it to anything they want, overriding the default.

Closing, please send a pull request to change the documentation instead

PR #926, opened 2018-09-10
Add missing `detail/Futex-inl.h` header

paulrosania:
4d234b991662b796d8e43bb585880b2bc15368f0 introduces `folly/detail/Futex-inl.h`. This patch adds it to `Makefile.am` so it gets installed when building using autotools.


Orvid:
I've just published v2018.09.10.01 which includes this fix :)

PR #939, opened 2018-09-28
Remove multiple definitions about dynamic::TypeInfo

jcooky:
constexpr static memeber (ODR-used) is implicitly inline.  
Then it does not need to be redeclared in .cpp file.

And the redeclaration of constexpr static members is deprecated since C++17

## References
- [cppreference](https://en.cppreference.com/w/cpp/language/static)

> If a static data member is declared constexpr, it is implicitly inline and does not need to be redeclared at namespace scope. This redeclaration without an initializer (formerly required as shown above) is still permitted, but is deprecated. (since C++ 17)

### DIFF
--- a/folly/dynamic.cpp
+++ b/folly/dynamic.cpp
@@ -26,22 +26,6 @@
 namespace folly {
 
 //////////////////////////////////////////////////////////////////////
-
-#define FOLLY_DYNAMIC_DEF_TYPEINFO(T)                 \
-  constexpr const char* dynamic::TypeInfo<T>::name;   \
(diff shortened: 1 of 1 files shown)


Orvid:
Unfortunately, Folly still needs to support pre-C++17 compilers, so the out of line definitions are still needed :(

jcooky:
So... How about adding C++ 17 defensive code like `#ifndef ~ #endif`.

For example, GCC is supported `__cpp_inline_variables` as a definition variable.

jcooky:
We resolve this issue to use `-std=gnu++1z` instead of `-std=c++17`.

But, I still hope folly to support c++ 17 features.

PR #945, opened 2018-10-04
Fix LLVM 7 self-assignment warnings in test code

JoeLoser:
Summary:
- LLVM 7 emits a warning about self-assignment.
- Silence the warning by using a static_cast.

### DIFF
--- a/folly/experimental/test/StringKeyedTest.cpp
+++ b/folly/experimental/test/StringKeyedTest.cpp
@@ -144,7 +144,7 @@ TEST(StringKeyedUnorderedMapTest, sanity) {
 
   EXPECT_EQ(map.size(), 2);
 
-  map = map;
+  map = static_cast<decltype(map)&>(map); // suppress self-assign warning
 
   EXPECT_EQ(map.find("hello")->second, 1);
   EXPECT_EQ(map.find("lo")->second, 3);
@@ -227,7 +227,7 @@ TEST(StringKeyedSetTest, sanity) {
 
   EXPECT_EQ(set.size(), 2);
 
-  set = set;
+  set = static_cast<decltype(set)&>(set); // suppress self-assign warning
 
   EXPECT_NE(set.find(StringPiece("hello")), set.end());
   EXPECT_NE(set.find("lo"), set.end());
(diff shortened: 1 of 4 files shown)


folly/experimental/test/StringKeyedTest.cpp
```diff
@@ -129,6 +129,11 @@ using LeakCheckedUnorderedSet = BasicStringKeyedUnorderedSet<
     std::equal_to<folly::StringPiece>,
     MemoryLeakCheckerAllocator<std::allocator<std::string>>>;
 
+#if __CLANG_PREREQ(7, 0)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wself-assign-overloaded"
```
yfeldblum: Might be able to use pure-C++ tricks to evade the warning.

folly/experimental/test/StringKeyedTest.cpp
```diff
@@ -158,6 +163,10 @@ TEST(StringKeyedUnorderedMapTest, sanity) {
   }
 }
 
+#if __CLANG_PREREQ(7, 0)
```
yfeldblum: If the push is unconditional, then the pop can also be unconditional.

folly/experimental/test/StringKeyedTest.cpp
JoeLoser: Yep, static cast with a decltype to itself silences the warning. Motivation from https://reviews.llvm.org/D44883

JoeLoser: @yfeldblum just reworded the commit.  If you would prefer that I put a comment of why we're doing 

```
my_var = static_cast<decltype(my_var)>(my_var);
```

instead of 

```
my_var = my_var;
```

I'm happy to do so.

yfeldblum:
> If you would prefer that I put a comment

Yes please - something short like `// suppress self-assign warning` would be good. Don't even need to say which compiler emits the warning because, while currently there is only one, theoretically more compilers could start doing it.

folly/experimental/test/StringKeyedTest.cpp
```diff
@@ -144,7 +144,7 @@ TEST(StringKeyedUnorderedMapTest, sanity) {
 
   EXPECT_EQ(map.size(), 2);
 
-  map = map;
+  map = static_cast<decltype(map)>(map);
```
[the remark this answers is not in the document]
JoeLoser: Good catch. Another option is to use `*&` which is even shorter (https://wandbox.org/permlink/lR4uC8IucJWvFWj2). I personally found
`static_cast<decltype(map)&>(map)` the easiest to read and understand, keeping in mind future readers of these tests. While the extra parenthesis certainly creates a reference, as you said, it's pretty subtle. A cast to a `T&` is explicit and easy to follow in my opinion.

PR #946, opened 2018-10-09
Simplify suppressing self-move warning

JoeLoser:
- A couple of tests pragma push/pop to avoid `-Wself-move` warning in
  code that does a move-assignment with itself.
- Use `static_cast<T&&>` instead of `std::move` which will suppress
  the warning still without the need for pragma push/pop.


yfeldblum:
Thanks!

PR #951, opened 2018-10-12
Reduce size of underlying buffer of AtomicBitSet

khvorov:
Reduce size of the underlying buffer used by AtomicBitSet. For a bitset of 65536 bits the size of the buffer will be reduced from 512KB to 8KB which is good for the systems with a small stack


Orvid:
This eventually got done, and AtomicBitSet is now known as ConcurrentBitSet.

PR #957, opened 2018-11-02
CMake: Use correct variable name for 'GFlags_FOUND'

daverigby:
When using find_package(GFlags) in CONFIG mode, it results in a variable named 'GFlags_FOUND' (note the case) being set if the package is found - see https://cmake.org/cmake/help/latest/command/find_package.html.

However, folly-deps.cmake is incorrectly using the variable 'gflags_FOUND' (lowercase). As a result CONFIG mode always fails.

Fix by using the correct variable name.

### DIFF
--- a/CMake/folly-deps.cmake
+++ b/CMake/folly-deps.cmake
@@ -23,7 +23,7 @@ list(APPEND FOLLY_INCLUDE_DIRECTORIES ${DOUBLE_CONVERSION_INCLUDE_DIR})
 
 set(FOLLY_HAVE_LIBGFLAGS OFF)
 find_package(GFlags CONFIG QUIET)
-if (gflags_FOUND)
+if (GFlags_FOUND)
   message(STATUS "Found gflags from package config")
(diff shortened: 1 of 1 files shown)


simpkins:
Hmm, it looks like this behavior was probably broken by pull request #805, which updated the case in the find_package() call but not the _FOUND check.

It seems to me that reverting the change from #805 might be the right behavior: At least on Ubuntu 18.04 the gflags cmake config file is installed as "gflags-config.cmake", all lowercase.
It looks like #805 was probably just a misunderstanding of how the cmake code works: I think they were expecting this line to use the CMake/FindGFlags.cmake file shipped with folly, where as this line is expected to find the "gflags-config.cmake" file installed by gflags.  There is a seprate find_package(GFlags) call below if this one fails that uses the module shipped with folly.

simpkins:
Note that either fix works in practice: with either `find_package(gflags CONFIG)` or `find_package(GFlags CONFIG)` CMake still correctly finds the all-lowercase `gflags-config.cmake` file installed by Ubuntu's libgflags-dev package.

Issue #847, opened 2018-05-21
getting error while configuring folly


krustagi:
Hi,

I am running below command to configure folly on linux based environment. 
cmake configure .

**_I am getting below error:
CMake Error at CMakeLists.txt:6 (apply_folly_compile_options_to_target):
  Unknown CMake command "apply_folly_compile_options_to_target"._**

CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.6)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring incomplete, errors occurred!
See also "/home/krastogi/git/folly/folly/CMakeFiles/CMakeOutput.log".

Please help me in resolving this error.

Orvid:
This is the result of running CMake from `folly/`. Run CMake in the root of the repo instead.

krustagi:
Tried same but same issue.

sungwonida:
> This is the result of running CMake from `folly/`. Run CMake in the root of the repo instead.

That was my case. The issue has been solved. Thanks!

PR #984, opened 2018-12-10
Correct Ubuntu build instructions

CodeWithOz:
Associated issue: #847 

The `cmake` and `make` commands given in the [last section](https://github.com/facebook/folly#ubuntu-1604-lts) of the Ubuntu build instructions all need to be run from inside the `folly` folder rather than the `_build` folder, as indicated by @Orvid in #847. Also, the final step (`make install`) may need to be `sudo make install` to allow access to the `/usr` folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,11 +141,21 @@ sudo apt-get install \
 
 In the folly directory, run:
 ```
-  mkdir _build && cd _build
+  mkdir _build
   cmake ..
   make -j $(nproc)
   make install
 ```
+
+If you receive an error like:
+
+> CMake Error at cmake_install.cmake:41 (file):
+> file INSTALL cannot copy file
+> "/home/[yourname]/[folly-version]/folly/libfolly.a" to
+> "/usr/local/lib/libfolly.a".
+
+then you [need to use `sudo make install`](https://askubuntu.com/a/954877)
+as the last step instead of just `make install`.
 
 #### OS X (Homebrew)
 
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ sudo apt-get install \
     libdwarf-dev
 ```
 
-In the folly directory, run:
+In the folly directory (e.g. the checkout root or the archive unpack root), run:
 ```
-  mkdir _build
+  mkdir _build && cd _build
   cmake ..
   make -j $(nproc)
   make install # with either sudo or DESTDIR as necessary
--- a/README.md
+++ b/README.md
@@ -144,18 +144,8 @@ In the folly directory, run:
   mkdir _build
   cmake ..
   make -j $(nproc)
-  make install
+  make install # with either sudo or DESTDIR as necessary
 ```
-
-If you receive an error like:
-
-> CMake Error at cmake_install.cmake:41 (file):
-> file INSTALL cannot copy file
-> "/home/[yourname]/[folly-version]/folly/libfolly.a" to
-> "/usr/local/lib/libfolly.a".
-
-then you [need to use `sudo make install`](https://askubuntu.com/a/954877)
-as the last step instead of just `make install`.
 
 #### OS X (Homebrew)
 


README.md
```diff
@@ -141,12 +141,22 @@ sudo apt-get install \
 
 In the folly directory, run:
 ```
-  mkdir _build && cd _build
+  mkdir _build
```
yfeldblum: The line below is `cmake ..`; that makes this change incorrect.

The reason for building from within the `_build` dir is to avoid cmake polluting the top-level dir with build artifacts, which it otherwise would do.

README.md
```diff
@@ -141,12 +141,22 @@ sudo apt-get install \
...
 ```
 
+If you receive an error like:
+
+> CMake Error at cmake_install.cmake:41 (file):
+> file INSTALL cannot copy file
+> "/home/[yourname]/[folly-version]/folly/libfolly.a" to
+> "/usr/local/lib/libfolly.a".
+
+then you [need to use `sudo make install`](https://askubuntu.com/a/954877)
```
yfeldblum: This link does not need to exist - this random answer to a random question is not any kind of canonical or authoritative source text.

README.md
```diff
@@ -141,12 +141,22 @@ sudo apt-get install \
...
+as the last step instead of just `make install`.
```
yfeldblum: Not sold on this whole comment needing to exist. Installing things into the system is well-known to require admin privileges or the root user, and sudo is the well-known way to get those on Ubuntu.

Of course there are many people who don't know this, but it's not really about folly and it's more about getting familiar with the operating system.

At best, we can have an in-line comment suggesting either to use sudo or to use DESTDIR but without offering any explanation. E.g.:

```
  make install # with either sudo or DESTDIR as necessary
```

README.md
CodeWithOz: Hmm do you mean the top level directory that contains the `.md` files and the `build`, `folly` and `CMake` folders? No build artifacts were added there. It's within the `folly` folder that a few new files and folders were created. More so, the commands just didn't work inside `[folly version]/folly/_build`. #847 contains the error I got from running `cmake ..` inside it, but both `make -j  $(nproc)` and `make install` gave errors that indicated necessary files were missing, and then they worked when I went back from `/[folly version]/folly/_build/` to `[folly version]/folly/` as @Orvid suggested.

README.md
CodeWithOz: Okay I will remove it.

README.md
CodeWithOz: Alright, I will make the change.

README.md
yfeldblum: Right, top-level or `folly/` subdir - we want to avoid build artifact pollution.

Was your shell in `[folly-checkout]/folly/_build` or in `[folly-checkout]/_build` when you ran `cmake ..`?

CodeWithOz: It was in `[folly-checkout]/folly/_build` when I ran it and got the errors. Should the `_build` folder have been made in `[folly-checkout]/` rather than `[folly-checkout]/folly/`? The README says
> In the folly directory

so I interpreted that as `[folly-checkout]/folly`.

yfeldblum: The shell should be in `[folly-checkout]/_build`, and `cmake ..` should be run from that directory.

CodeWithOz: Okay, thanks for clarifying. So what do you think of these changes:

In the root of the folly directory, run:
```
mkdir _build && cd _build # puts the shell in [folly-checkout]/_build
```

yfeldblum: I don't see that as a helpful comment because it merely restates the code into English.

yfeldblum: "The X directory" and "the root of the X directory" mean the same thing. However you might say

> In the folly directory (e.g. the checkout root or the archive unpack root), run:

CodeWithOz: Okay, thank you. I have updated the code.

PR #989, opened 2019-01-05
folly/synchronization/LifoSem: fixed race condition between tryRemoveNode and shutdown by checking the lock in shutdown

d4l3k:
Summary:
Original post: https://fb.workplace.com/groups/560979627394613/permalink/1370192126473355/

There was a bug where we weren't checking the isLocked bit when setting the isShutdown bit. Thus, tryRemoveNode would acquire the lock, shutdown would set the shutdown bit, and then tryRemoveNode would release the lock via a direct store which accidentally cleared the shutdown bit.

The new code wait loops in shutdown until the lock is cleared.

Differential Revision: D13586264

### DIFF
--- a/folly/synchronization/LifoSem.h
+++ b/folly/synchronization/LifoSem.h
@@ -400,6 +400,12 @@ struct LifoSemBase {
     // first set the shutdown bit
     auto h = head_->load(std::memory_order_acquire);
     while (!h.isShutdown()) {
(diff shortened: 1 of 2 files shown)


yfeldblum:
Exporting pull requests from internal changes to folly is not a required part of the workflow. Internal changes will automatically be synced to github once they are landed.

PR #995, opened 2019-01-12
Remove folly::is_trivially_copyable

JoeLoser:
Summary:
- There is no need for `folly::is_trivially_copyable` in
  `folly/Traits.h` since `std::is_trivially_copyable` ships with the
  standard library for GCC 5 and later.
- Previously, for GCC 4.9 and below, the symbol
  `std::is_trivially_copyable` was not present.
- Remove the CMake check for detecting presence of the symbol.


yfeldblum:
While Folly no longer supports gcc49 as a whole and is documented (thanks!) that way, there are some internal cases which are still on gcc49 and which are using some undocumented subsets of Folly. So this PR can't be merged quite yet.

yfeldblum:
I can't give specifics, but progress is being made. I want to bump to gcc5 also.

JoeLoser:
Just rebased, fixed the recent `bitCast` utility to use `std::is_trivially_copyable`, and also removed the `#ifdef` for the `TriviallyCopyable` test in ExpectedTest. With this change, we expect any compiler we claim support for should have `std::is_trivially_copyable`, so we should enforce that with this test rather than only testing on Clang/GCC.

JoeLoser:
I'll rebase again tonight as a703ff288c18f797aebd815d0cc7144bab473c21 landed with some usages of `folly::is_trivially_copyable` and we also have conflicts now in `Traits.h`.

JoeLoser:
Rebased. Corresponding Thrift PR is here: https://github.com/facebook/fbthrift/pull/327

yfeldblum:
Not as rapid progress as one might hope, unfortunately. I couldn't land this internally yet.

JoeLoser:
Thanks for the update. I know you'll be happy only having to support gcc5 and later soon.

JoeLoser:
Rebased and fixed conflicts. Any update on the progress of moving all of the internal projects off of GCC 4.9 so we can land this?

JoeLoser:
Rebased with top of tree. Now that folly only supports GCC 5+, we should be able to land this safely, right?

yfeldblum:
I hope so!

yfeldblum:
Progress, but it looks like there are still cases internally of using clang with older libstdc++ which does not define `std::is_trivially_copyable`.

JoeLoser:
Thanks for checking. We'll get there one day.

yfeldblum:
This has now been done.

PR #996, opened 2019-01-12
Remove workaround for lack of std::atomic_init

JoeLoser:
Summary:
- Since GCC 5 and later has `std::atomic_init`, remove the workaround
  present in `Tearable.h` to default initialize atomic variables.
- Default initialization of atomics do not work as you would expect. See
  http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf
  for the explanation why.
- To get around the default initialization issue, we just call
  `std::atomic_init` for each element in the array of atomics.

### DIFF
--- a/folly/synchronization/Tearable.h
+++ b/folly/synchronization/Tearable.h
@@ -53,7 +53,13 @@ class Tearable {
       is_trivially_copyable<T>::value,
       "Tearable types must be trivially copyable.");
 
-  Tearable() = default;
+  Tearable() noexcept
+  {
+      for (std::size_t i = 0; i < kNumDataWords; ++i)
+      {
+          std::atomic_init(&data_[i], RawWord{});
+      }
+  }
 
   Tearable(const T& val) : Tearable() {
     store(val);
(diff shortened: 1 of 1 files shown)


yfeldblum:
From internal review, pedantically, this leaves the atomic in uninitialized state.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0883r0.pdf

So `Tearable` would need either this intermediate type or to call `std::atomic_init` in a loop.

JoeLoser:
Thanks for the paper link; I was not aware of this gotcha. It's a bit unfortunate the default initialization for atomics is broken IMO. I like Herb's quote of: "I have never seen the current behavior of deliberately failing to initialize an atomic<T> (to the
obvious default value of T{}) to be anything but a source of user surprise and a bug farm."

In any case, I decided to switch to using an array of atomics and explicitly call `std::atomic_init` for each element. I think this new behavior is a bit more explicit in our intent to default initialize the elements rather than through the intermediate type. What do you think?

folly/synchronization/Tearable.h
```diff
@@ -63,17 +70,17 @@ class Tearable {
   // anything* with the result may trigger undefined behavior unless you've
   // verified that the data you read was consistent.
   void load(T& dst) const {
-    RawWord newDst[kNumDataWords];
+    std::array<RawWord, kNumDataWords> newDst;
```
yfeldblum: This is an unrelated change. Let's stick to one thesis here.

folly/synchronization/Tearable.h
```diff
@@ -53,7 +53,14 @@ class Tearable {
       is_trivially_copyable<T>::value,
       "Tearable types must be trivially copyable.");
 
-  Tearable() = default;
+  Tearable() noexcept
```
yfeldblum: (This will be reformatted before merging.)

folly/synchronization/Tearable.h
JoeLoser: Fair enough, I just changed it to support the minimal change of removing the intermediate type in favor of calling `std::atomic_init` in a loop. I will save the `std::array` change for another PR.

PR #998, opened 2019-01-14
Reserve the vector to save reallocation cost.

shahzadlone:
Even though this is a test, we are still reallocating a vector (resizing until re-sized/reallocated to 32), which in theory calls for allocation 5 times. But we can save this by just reserving the vector pre-hand (we know the size).


shahzadlone:
@facebook-github-bot Is there anything else you need?

shahzadlone:
@yfeldblum does imported mean that this was merged?

yfeldblum:
It means that a copy of this PR exists in the internal code review tool.

shahzadlone:
I see @yfeldblum, so then you will close this one once the internal one merges?

yfeldblum:
Yes, the PR will be automatically closed.

shahzadlone:
Thanks for the merge! @facebook-github-bot

PR #999, opened 2019-01-17
Remove disabling -Warray-bounds for GCC in FBString.h

JoeLoser:
Summary:
- Existing code disables `-Warray-bounds` warning for GCC due to a bug in
  GCC 4.9.
- As seen in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59124, the bug
  has been resolved for GCC 5.1.
- Since the bug has been resolved, we no longer need to disable the
  warning.

@yfeldblum I suspect that we have to hold off on this PR until the internal use cases relying on GCC 4.9 upgrade, but I figured it is better to have some of these cleanup opportunities ready for when the time arises.

### DIFF
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -78,10 +78,6 @@
 FOLLY_PUSH_WARNING
 // Ignore shadowing warnings within this file, so includers can use -Wshadow.
 FOLLY_GNU_DISABLE_WARNING("-Wshadow")
-// GCC 4.9 has a false positive in setSmallSize (probably
-// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59124), disable
-// compile-time array bound checking.
-FOLLY_GNU_DISABLE_WARNING("-Warray-bounds")
 
 // FBString cannot use throw when replacing std::string, though it may still
 // use folly::throw_exception


JoeLoser:
@yfeldblum does this PR also have to wait until the remaining internal uses of GCC 4.9 move on to GCC 5+?

JoeLoser:
Ping @lbrandy in case this is blocked in the internal review as well. Would appreciate the review.

Last I heard from @yfeldblum, there were a few internal things relying on GCC 4.9, so we may not quite be able to land this PR until they move over to GCC 5+. Let me know if that is the case so we can proceed accordingly.

yfeldblum:
Hey @JoeLoser, there is some internal build configuration which I think is gcc5 but which doesn't agree and is showing build-time failures that I don't quite understand it yet.

```
folly/FixedString.h:145:13: error: array subscript is above array bounds [-Werror=array-bounds]
       ? left[i]
             ^
```

yfeldblum:
Also fwiw, I don't see anything internally atm which still uses gcc49 depending on this.

JoeLoser:
Thanks for the update and looking into this a bit @yfeldblum. After rereading some of the comments on the GCC bug report (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59124), my takeaway was that that bug is "mostly fixed" in what made the GCC 5 branch, but there are some lingering issues with it that remain even in GCC 8 (but were fine in GCC 6/7).

If we're hitting one of those cases that have not yet been fixed, it seems like we should keep the disabling of the warning in, but not bound it to GCC versions 5 or less (https://github.com/facebook/folly/pull/999/files#diff-614a0f21886a05769b4e7c183635055aL105)

yfeldblum:
Yeah that's fine. At least keep the suppression in FixedString.h but we can try removing the suppression only in FBString.h. Want to update the PR?

JoeLoser:
Sure, I kept `FixedString.h` as is for now. The comments on the separate bug report (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61971) than the one in `FBString.h` show that this one was fully backported and fixed in GCC 5. But, until we get it to not trip on the internal failure, it seems best to leave `FixedString.h` untouched.

PR #1001, opened 2019-01-17
Use boost::has_less instead of rolling our own

JoeLoser:
Summary:
- An internal type trait, `HasLess` is used to detect if there exists an
  `operator<` such that `T < U` is well-formed.
- This type trait is not needed with recent versions of Boost which has
  this exact type trait: `boost::has_less`.
- Since we rely on Boost 1.55 or later which does contain this symbol,
  just use that.

Note:
- A future commit will change `notThereYet` to use the `#if 0`
  implementation now that we are on Boost 1.46 or later.

### DIFF
--- a/folly/container/Foreach.h
+++ b/folly/container/Foreach.h
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <boost/type_traits/has_less.hpp>
+
 #include <folly/Portability.h>
 #include <folly/Preprocessor.h>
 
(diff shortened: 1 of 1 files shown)


JoeLoser:
@yfeldblum anything I can do to help get this landed?

lbrandy:
This was blocked on internal review. I just accepted it...

JoeLoser:
Thanks for the review, @lbrandy 👍

yfeldblum:
Thanks for flagging this.

Note that I just removed `HasLess` in ff5f718ecab5dd2de75045097f06aa61aad513a3 and f2c0cfdd576c08f1e0217079cba79bc0e4c12312.

PR #1019, opened 2019-02-15
CMake: folly-deps: gflags is no longer gflags-shared on Fedora

kornrunner:
This undoes #789 

Previous patch worked fine for Fedora 25, 26 and 27. Current release is Fedora 29, and `gflags` package in it is again identified as `gflags` (vs previous `gflags-shared`).

This might be BC break, but it's current and likely forward compatible.
Roots for this change are actually within https://github.com/facebookincubator/fizz/issues/20

Sorry for bothering you again with this. Thank you!

/cc @yfeldblum


dbussink:
Looks like the change in #789 also broke building on Debian 9 (Stretch).

PR #1020, opened 2019-02-16
Cut constexpr function comp objects in ConstexprMath.h

JoeLoser:
Summary:
- We have several function objects that define constexpr call operators
  for `operator<`, `operator<=`, etc. These were needed prior to
  C++14 in order for them to be `constexpr`, as vendor implementations did not
  mark them `constexpr`.
- Since Folly requires C++14, callers who are relying on these should
  migrate to `std::less<T>`, and other function objects in the standard
  library, which are now marked constexpr.

### DIFF
--- a/folly/ConstexprMath.h
+++ b/folly/ConstexprMath.h
@@ -17,49 +17,11 @@
 #pragma once
 
 #include <cstdint>
+#include <functional>
 #include <limits>
 #include <type_traits>
 
 namespace folly {
-
-// TODO: Replace with std::equal_to, etc., after upgrading to C++14.
-template <typename T>
-struct constexpr_equal_to {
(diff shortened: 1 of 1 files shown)


yfeldblum:
We're not *quiiiite* there yet internally. But it is the direction we're going!

JoeLoser:
Not yet because GCC 4.9 didn't mark these `constexpr`?

yfeldblum:
Right.

JoeLoser:
Rebased with top of tree. Since folly requires C++14 now, this should be easier to land, @yfeldblum

yfeldblum:
I hope so! There is still a chance that the change to enforce against gcc < 5 will need to be reverted, but hopefully not.

PR #1021, opened 2019-02-16
Cut backport of std::exchange

JoeLoser:
Summary:
- Backporting `std:exchange` is no longer needed as Folly requires C++14
  support and C++14-compliant vendors have `std::exchange` in their
  standard library, notably including GCC 4.9.
- Cut backport of `std::exchange` and fix call sites to explicitly use
  `std::exchange` where they were previously relying on `exchange` being
  found in the `folly` namespace.

### DIFF
--- a/folly/Executor.h
+++ b/folly/Executor.h
@@ -18,6 +18,7 @@
 
 #include <cassert>
 #include <climits>
+#include <utility>
 
 #include <folly/Function.h>
 #include <folly/Utility.h>
@@ -59,7 +60,7 @@ class Executor {
     }
 
     KeepAlive(KeepAlive&& other) noexcept
-        : executorAndDummyFlag_(exchange(other.executorAndDummyFlag_, 0)) {}
+        : executorAndDummyFlag_(std::exchange(other.executorAndDummyFlag_, 0)) {}
 
     KeepAlive(const KeepAlive& other) noexcept
         : KeepAlive(getKeepAliveToken(other.get())) {}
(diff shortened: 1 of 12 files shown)


JoeLoser:
@yfeldblum this shouldn't break anything internally as even GCC 4.9 supports `std::exchange`.

I don't have strong opinions here, but if we want a `constexpr` exchange, we should keep `folly::exchange`,  mark it `constexpr`, and possibly migrate all call sites to use `folly::exchange`. In the C++20 standard library, `std::exchange` is marked `constexpr`.

yfeldblum:
Okay so just as a bit of background on what's happening internally, we have some code using folly with gcc49, and we also have some code using folly with gcc5+ in c++11 (v.s. c++14) mode.

I think I was not clear in another PR, which was about a warning where gcc49 vs gcc5+ mattered. This PR is about a library backport, where c++11 vs c++14 also matters.

I haven't imported this and run internal CI yet, so I don't know yet whether or not that will be a problem for *this* PR. This is just some background since it might be a blocker for other similar PRs.

JoeLoser:
Thanks for the insight. I wasn't aware we were also still supporting C++11 in some cases.  I was under the impression it was C++14 across the board with those falling into two groups: GCC 4.9 and GCC 5.1 or later.

Sorry to bother with the removal of C++14 backports then in the case that it breaks the C++11 use cases.

yfeldblum:
Sorry, I guess I wasn't clear again!

Folly is (currently) a C++14 library, supporting versions of gcc >= 5 (plus some other compilers). So we should remove legacy bits. Thanks!

Internally, we've got some drag, since sometimes we find something which was on C++11 and which was using some piece of folly which just happened to work, but which would break on a change like this. So it slows down this type of work.

yfeldblum:
Missing at least one call. CI shows `folly/synchronization/detail/ProxyLockable-inl.h`.

JoeLoser:
Thanks for the explanation and noting a missing call.  I just pushed up a fix to have `ProxyLockable-Inl.h` use `std::exchange` as well.

PR #1022, opened 2019-02-16
Cut backport of std::index_sequence and friends

JoeLoser:
Summary:
- Backporting `std::index_sequence`, `std::integer_sequence`, and other
  associated template aliases such as `index_sequence_for` and
  `make_index_sequence` is not needed as Folly requires C++14. These
  functions and template aliases are available in C++14 standard libraries,
  including GCC 4.9.

### DIFF
--- a/folly/Fingerprint.cpp
+++ b/folly/Fingerprint.cpp
@@ -20,6 +20,8 @@
 #include <folly/Utility.h>
 #include <folly/detail/FingerprintPolynomial.h>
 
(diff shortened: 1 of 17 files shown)


JoeLoser:
@yfeldblum similar to https://github.com/facebook/folly/pull/1021, this shouldn't break anything internally as GCC 4.9 implements `std::integer_sequence`, `std::index_sequence`, etc.

PR #1024, opened 2019-02-20
Replace [[no_discard]] with FOLLY_NODISCARD

JoeLoser:
Summay:
- Some call sites use `[[no_discard]]` which will be ignored by some
  older C++14 compilers, such as GCC 4.9 through GCC 6.X. This can
  result in a warning emitted to the effect of "nodiscard attribute
  directive ignored".
- Change the call sites to use `FOLLY_NODISCARD` which will only expand
  to `[[no_discard]]` if and only if the attribute exists (by detection of
  feature test macro for it).
- This will prevent warnings about ignored attribute directives and
  provide consistency with other call sites using `FOLLY_NODISCARD`.

### DIFF
--- a/folly/experimental/coro/Baton.h
+++ b/folly/experimental/coro/Baton.h
@@ -18,6 +18,8 @@
 #include <atomic>
 #include <experimental/coroutine>
 
+#include <folly/Portability.h>
+
 namespace folly {
 namespace coro {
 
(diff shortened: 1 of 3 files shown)


yfeldblum:
All of these changes are in the coroutines library code. GCC does not yet support coroutines at all, so these headers should not be included with that compiler, and so the `[[no_discard]]` issue should not be relevant to any currently released version of GCC.

PR #1025, opened 2019-02-20
Remove trailing return type for to_signed and to_unsigned

JoeLoser:
Summary:
- `to_signed` and `to_unsigned` specify their return type via trailing
  return type syntax. This was used to help older compilers out with
  `auto` return type.
- Cut the trailing return type syntax as compilers are sufficiently
  smart enough to deduce the return type in these cases.

### DIFF
--- a/folly/Utility.h
+++ b/folly/Utility.h
@@ -369,7 +369,7 @@ class MoveOnly {
 using MoveOnly = moveonly_::MoveOnly;
 
 template <typename T>
-constexpr auto to_signed(T const& t) -> typename std::make_signed<T>::type {
+constexpr auto to_signed(T const& t) {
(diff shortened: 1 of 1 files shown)


yfeldblum:
I prefer keeping explicit return types in public and exported functions as a rule. They don't need to be trailing.

JoeLoser:
Thanks for the rationale. I saw we had a lot of trailing return type, so this was just to see what your opinion on the matter was as I was in this neighborhood of code when writing `to_underlying_type`.

Since a lot of the uses are in public-facing headers, I agree with you that we should be explicit about our return types. So, I'll go ahead and close this PR and not worry about the other trailing return types.

yfeldblum:
👍

PR #1028, opened 2019-02-21
Add to_underlying_type to cast enum to underlying

JoeLoser:
Summary:
- It is verbose at the call sites to cast an enum or enum class to its
  underlying type.
- Introduce `to_underlying_type`, which, given an enum or enum class,
  returns the value from static casting to the underlying type.

### DIFF
--- a/folly/Utility.h
+++ b/folly/Utility.h
@@ -22,6 +22,7 @@
 #include <utility>
 
 #include <folly/CPortability.h>
+#include <folly/Portability.h>
 #include <folly/Traits.h>
 
 namespace folly {
@@ -383,5 +384,12 @@ constexpr auto to_unsigned(T const& t) -> typename std::make_unsigned<T>::type {
   using U = typename std::make_unsigned<T>::type;
   return static_cast<U>(t);
 }
+
+template <class E>
+constexpr std::underlying_type_t<E> to_underlying_type(E e) noexcept {
+    static_assert(std::is_enum<E>::value, "not an enum type");
+    return static_cast<std::underlying_type_t<E>>(e);
+}
+
 
 } // namespace folly
--- a/folly/Conv.h
+++ b/folly/Conv.h
@@ -43,6 +43,7 @@
 #include <folly/Range.h>
 #include <folly/Traits.h>
 #include <folly/Unit.h>
+#include <folly/Utility.h>
 #include <folly/lang/Exception.h>
 #include <folly/lang/Pretty.h>
 #include <folly/portability/Math.h>
@@ -648,15 +649,13 @@ template <class Tgt, class Src>
 typename std::enable_if<
     std::is_enum<Src>::value && IsSomeString<Tgt>::value>::type
 toAppend(Src value, Tgt* result) {
-  toAppend(
-      static_cast<typename std::underlying_type<Src>::type>(value), result);
+  toAppend(to_underlying_type(value), result);
 }
 
 template <class Src>
 typename std::enable_if<std::is_enum<Src>::value, size_t>::type
 estimateSpaceNeeded(Src value) {
-  return estimateSpaceNeeded(
-      static_cast<typename std::underlying_type<Src>::type>(value));
+  return estimateSpaceNeeded(to_underlying_type(value));
 }
 
 /*******************************************************************************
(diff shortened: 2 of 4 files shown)


folly/Utility.h
```diff
@@ -384,4 +385,13 @@ constexpr auto to_unsigned(T const& t) -> typename std::make_unsigned<T>::type {
   return static_cast<U>(t);
 }
 
+template <class E>
+FOLLY_NODISCARD constexpr auto to_underlying_type(E e) noexcept
+{
+    static_assert(std::is_enum<E>::value, "to_underlying_type requires the"
```
yfeldblum: Shorter message: `"not an enum type"`.

folly/Utility.h
```diff
@@ -384,4 +385,13 @@ constexpr auto to_unsigned(T const& t) -> typename std::make_unsigned<T>::type {
   return static_cast<U>(t);
 }
 
+template <class E>
+FOLLY_NODISCARD constexpr auto to_underlying_type(E e) noexcept
+{
```
yfeldblum: Note: I would auto-reformat this anyway after importing, but `{` would go on the previous line.

folly/Utility.h
```diff
@@ -384,4 +385,13 @@ constexpr auto to_unsigned(T const& t) -> typename std::make_unsigned<T>::type {
...
+    static_assert(std::is_enum<E>::value, "to_underlying_type requires the"
+            "underlying type to be an enum or enum class");
```
yfeldblum: This line is not even correct! Anyway the message can be super short, as above.

folly/Utility.h
```diff
@@ -384,4 +385,13 @@ constexpr auto to_unsigned(T const& t) -> typename std::make_unsigned<T>::type {
   return static_cast<U>(t);
 }
 
+template <class E>
+FOLLY_NODISCARD constexpr auto to_underlying_type(E e) noexcept
```
yfeldblum: I tend to prefer writing return types explicitly for public and exported functions.

The `[[no_discard]]` does not seem useful because this function does not have
* any status / error condition which are signaled by the return value
* any side-effects which are managed by the return value

Using `[[no_discard]]` on pure functions such as this does not seem to offer value to me.

folly/Utility.h
JoeLoser: Reformatted to put the brace on the previous line.

folly/Utility.h
JoeLoser: Fair point. I like that over multiline literals anyway. 👍

folly/Utility.h
JoeLoser: Thanks for your explanation. We've gone back and forth about our use of `[[no_discard]]` in our codebases. We began by sprinkling it on all free functions and member functions (shame we can't use it on lambdas...), but after a while of use, we came to the same reasoning you have: only to use it where it is most likely a true error if a caller ignores the return values (such as an `empty()` or `size()` member function on a container).

We're in agreement and I've removed the `[[no_discard]]` attribute.

JoeLoser:
Thanks for the review, @yfeldblum. I've fixed up your comments.

What do you think about extending `to_underlying_type` return the value if a user passes in an integral type? This would make https://github.com/facebook/folly/blob/master/folly/DynamicConverter.h#L213 also eligible for `to_underlying_type`. Something like

```
template <class E>
constexpr std::underlying_type_t<E> to_underlying_type(E e) noexcept {
   static_assert(std::is_enum<E>::value, "not an enum type");
   if (std::is_enum<E>::value)
   {
       return static_cast<std::underlying_type_t<E>>(e);
   }
   else if (std::is_integral<E>::value)
    {
        return e;
    }
    else
    {
         static_assert(always_false<E>::value, "not an enum or integral type");
    }
 }
```

This would let callers use this utility in more generic programming contexts where you know you are working a regular value type, but not sure whether it is wrapped in an enum or enum class. It eliminates the need to do SFINAE machinery (such as checking if the type is an enum) to not force a hard error in a caller using `to_underlying_type`.

yfeldblum:
Maybe, but that sounds like a completely different function.

E.g. `to_signed` just wraps `static_cast<make_signed_t<...>>` (effectively).

Then you run into questions like, why not work on arithmetic but non-integral types, etc.?

`std::underlying_type` is only defined over complete enumeration type until C++20, and is only defined over complete types and only has a member type alias `type` for complete enumeration types since C++20. I'd just go with `to_underlying_type` being a wrapper for `static_cast<underlying_type_t<...>>` with param type deduction.

PR #1029, opened 2019-02-21
Cut FOLLY_USE_CPP14_CONSTEXPR and FOLLY_CPP14_CONSTEXPR

JoeLoser:
Summary:
- `FOLLY_USE_CPP14_CONSTEXPR` macro is not needed as Folly requires a
  recent enough version of MSVC where its constexpr support is "good
  enough". For Clang and GCC, the min compiler versions supported would
  both evaluate the prior implementation of this macro to true in both
  cases. This is potentially slightly behavior changing since
  `FOLLY_USE_CPP14_CONSTEXPR` would be `inline` for ICC and I am not
  sure if its constexpr support is "good enough" for a min version of
  ICC we claim support for.
- Replace `FOLLY_CPP14_CONSTEXPR` with `constexpr` in all call sites and
  remove the `FOLLY_CPP14_CONSTEXPR` macro.
- Simplify how we define `FOLLY_STORAGE_CONSTEXPR` and
  `FOLLY_STORAGE_CPP14_CONSTEXPR` after cutting
  `FOLLY_USE_CPP14_CONSTEXPR`.

### DIFF
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -440,24 +440,6 @@ constexpr auto kCpplibVer = 0;
 #endif
 } // namespace folly
 
-// Define FOLLY_USE_CPP14_CONSTEXPR to be true if the compiler's C++14
-// constexpr support is "good enough".
-#ifndef FOLLY_USE_CPP14_CONSTEXPR
-#if defined(__clang__)
-#define FOLLY_USE_CPP14_CONSTEXPR __cplusplus >= 201300L
-#elif defined(__GNUC__)
-#define FOLLY_USE_CPP14_CONSTEXPR __cplusplus >= 201304L
-#else
-#define FOLLY_USE_CPP14_CONSTEXPR 0 // MSVC?
-#endif
-#endif
-
-#if FOLLY_USE_CPP14_CONSTEXPR
-#define FOLLY_CPP14_CONSTEXPR constexpr
-#else
-#define FOLLY_CPP14_CONSTEXPR inline
-#endif
-
 //  MSVC does not permit:
 //
 //    extern int const num;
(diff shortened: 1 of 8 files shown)


folly/Portability.h
```diff
@@ -440,18 +440,6 @@ constexpr auto kCpplibVer = 0;
...
-#if defined(__clang__)
-#define FOLLY_USE_CPP14_CONSTEXPR __cplusplus >= 201300L
-#elif defined(__GNUC__)
-#define FOLLY_USE_CPP14_CONSTEXPR __cplusplus >= 201304L
-#else
-#define FOLLY_USE_CPP14_CONSTEXPR 0 // MSVC?
-#endif
-#endif
-
 #if FOLLY_USE_CPP14_CONSTEXPR
```
yfeldblum: Need to address this too in the same PR.

JoeLoser: Ah, good point. I'll fix up `FOLLY_CPP14_CONSTEXPR` and also `FOLLY_STORAGE_CPP14_CONSTEXPR`. 

`FOLLY_STORAGE_CPP14_CONSTEXPR` may break ICC. Do we make any guarantees in supporting a min version of ICC? See a8b6dd79.

JoeLoser: I've fixed this up now. Please take a peek.

JoeLoser:
This is potentially behavior-changing for ICC since `FOLLY_USE_CPP14_CONSTEXPR` may define to `0` for some versions of ICC we may still support? I don't know if Folly makes any guarantees of min compiler versions of ICC supported. But, given a8b6dd79, you may want to run CI internally with some projects relying on ICC to determine breakage earlier rather than later.

folly/Portability.h
```diff
@@ -472,17 +454,11 @@ constexpr auto kCpplibVer = 0;
 #if _MSC_VER
 #define FOLLY_STORAGE_CONSTEXPR
 #define FOLLY_STORAGE_CPP14_CONSTEXPR
```
yfeldblum: Let's remove `FOLLY_STORAGE_CONSTEXPR` and let's rename `FOLLY_STORAGE_CPP14_CONSTEXPR` to `FOLLY_STORAGE_CONSTEXPR`.

Looking at this code, it seems that the intention was that icc with c++14 would be okay with `constexpr` storage. AFAICT this just needs to distinguish between gcc/clang (and icc?) v.s. msvc.

PR #1030, opened 2019-02-21
Replace _LIBCPP_VERSION with FOLLY_USE_LIBCPP in some cases

JoeLoser:
Problem:
- When using `_LIBCPP_VERSION` without a specific version check, it is
  hard to reason about intent.

Solution:
- To make intent clear at the the point of use, use `FOLLY_USE_LIBCPP`
  when we only care if the user is using `libc++` but we do not care about
  the specific version.
- In the case we check against a particular version, still use
  `_LIBCPP_VERSION`.

### DIFF
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -2771,7 +2771,7 @@ operator<<(
         typename basic_fbstring<E, T, A, S>::value_type,
         typename basic_fbstring<E, T, A, S>::traits_type>& os,
     const basic_fbstring<E, T, A, S>& str) {
-#if _LIBCPP_VERSION
+#if FOLLY_USE_LIBCPP
   typedef std::basic_ostream<
       typename basic_fbstring<E, T, A, S>::value_type,
       typename basic_fbstring<E, T, A, S>::traits_type>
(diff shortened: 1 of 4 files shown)


yfeldblum:
> When using `_LIBCPP_VERSION` without a specific version check, it is hard to reason about intent.

Intent is to check whether any supported version of libc++ is being used. Why do you say this is hard to reason about?

JoeLoser:
Right, I understand the intent. The commit message should probably not say that it is "hard to reason about". 

From my perspective, it is a bit weird to not use `FOLLY_USE_LIBCPP` when my understanding of the purpose of that macro is exactly as you described: to cover the case of any libc++ version being used by the user. In the end, this commit only provides a small readability improvement (in my opinion, YMMV). I would expect `libc++` checks to fall into two groups:

1. Use `FOLLY_HAS_LIBCPP` when the specific version of `libc++` is not relevant
2. Use `_LIBCPP_VERSION` when doing specific `libc++` version checks

I do not have strong opinions about such a small change, so feel free to close the PR if you think it's something that doesn't improve the overall readability.

yfeldblum:
AFAICT `FOLLY_HAS_LIBCPP` is just used in [one guard](https://github.com/facebook/folly/blob/v2019.02.18.00/folly/Portability.h#L221-L225) which may be found in a context in which no previous c++ stdlib headers have yet been included.

PR #1032, opened 2019-02-25
Surround code snippets with tildes 

xuning97:
Angled brackets inside angled brackets is not displayed.
This makes the document misleading.


shixiao:
@xuning97, thanks for the PR!

This is an oversight on our part. Instead of escaping, we should make all the code snippets in F14.md `inline code spans` by adding surrounding them with tildes (`). Would you mind updating your PR accordingly? Thanks again!

xuning97:
@shixiao updated the PR, please check.

PR #1034, opened 2019-02-27
Cut noexcept calculations in dynamic-inl.h

JoeLoser:
Summary:
- GCC 4.9 previously required marking `TypeError` special member
  functions with conditional noexcept.
- GCC 5.1 does not have this issue. So, remove the conditional noexcept
  as they always evalaute to true in the existing code.

### DIFF
--- a/folly/dynamic-inl.h
+++ b/folly/dynamic-inl.h
@@ -134,17 +134,6 @@ struct FOLLY_EXPORT TypeError : std::runtime_error {
       const std::string& expected,
       dynamic::Type actual1,
       dynamic::Type actual2);
-  // TODO: noexcept calculation required through gcc-v4.9; remove once upgrading
-  // to gcc-v5.
-  TypeError(const TypeError&) noexcept(
-      std::is_nothrow_copy_constructible<std::runtime_error>::value);
-  TypeError& operator=(const TypeError&) noexcept(
-      std::is_nothrow_copy_assignable<std::runtime_error>::value);
-  TypeError(TypeError&&) noexcept(
-      std::is_nothrow_move_constructible<std::runtime_error>::value);
-  TypeError& operator=(TypeError&&) noexcept(
-      std::is_nothrow_move_assignable<std::runtime_error>::value);
-  ~TypeError() override;
 };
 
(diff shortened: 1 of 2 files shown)


JoeLoser:
Thanks, just fixed the corresponding `.cpp` file as well. Any particular reason all five special member functions aren't just `=default`ed in `dynamic-inl.h`?

yfeldblum:
There was a reason which appears no longer to be relevant. This was to outline these functions to shrink the code size of `throw TypeError(...);` statements.

We are now using `throw_exception<TypeError>(...);` statements which already takes care of shrinking the code size, so the reason is no longer applicable.

So there should be no problem with declaring all of these as defaulted within the class body, which would automatically take care of finding the proper `noexcept` specification.

JoeLoser:
Thanks for the explanation -- makes sense. I went ahead and just defaulted them in `dynamic-inl.h`.

folly/dynamic-inl.h
```diff
@@ -134,17 +134,11 @@ struct FOLLY_EXPORT TypeError : std::runtime_error {
...
-  TypeError(TypeError&&) noexcept(
-      std::is_nothrow_move_constructible<std::runtime_error>::value);
-  TypeError& operator=(TypeError&&) noexcept(
-      std::is_nothrow_move_assignable<std::runtime_error>::value);
-  ~TypeError() override;
+  TypeError(const TypeError&) = default;
+  TypeError& operator=(const TypeError&) = default;
+  TypeError(TypeError&&) = default;;
+  TypeError& operator=(TypeError&&) = default;
+  ~TypeError() override = default;
```
yfeldblum: Actually can we just omit all five of these and let them all be implicitly defaulted?

JoeLoser: Yep, even better. The special member functions will be implicitly generated and be noexcept-correct. 👍 

Just pushed up a change to just omit them all.

yfeldblum:
I am internally rewriting the commit message on this one before landing to reflect that this is now about de-outlining the special member functions.

JoeLoser:
Nice reword of the commit. Thanks for that and helping it land!

PR #1035, opened 2019-02-27
Allow trivially move constructible types to be relocatable

JoeLoser:
Summary:
- If a type is not marked explicitly with `IsRelocatable`
  as `std::true_type`, we infer its relocability only based on the
  property of whether the type is trivially copyable.
- Extend the default relocability to also be true for trivially move
  constructible types.

### DIFF
--- a/folly/test/TraitsTest.cpp
+++ b/folly/test/TraitsTest.cpp
@@ -21,6 +21,7 @@
 #include <type_traits>
 #include <utility>
 
+#include <folly/Portability.h>
 #include <folly/ScopeGuard.h>
 #include <folly/portability/GTest.h>
 
@@ -96,6 +97,19 @@ TEST(Traits, typedefd) {
 TEST(Traits, unset) {
   EXPECT_TRUE(IsRelocatable<F1>::value);
   EXPECT_TRUE(IsRelocatable<F4>::value);
+}
+
+TEST(Traits, triviallyMoveConstructible) {
+FOLLY_PUSH_WARNING
+FOLLY_CLANG_DISABLE_WARNING("-Wunneeded-member-function")
+  struct TriviallyMoveConstructible {
+    TriviallyMoveConstructible(const TriviallyMoveConstructible&) = delete;
+    TriviallyMoveConstructible& operator=(const TriviallyMoveConstructible&) = delete;
+    TriviallyMoveConstructible(TriviallyMoveConstructible&&) = default;
+    TriviallyMoveConstructible& operator=(TriviallyMoveConstructible&&) = delete;
+  };
+  EXPECT_TRUE(IsRelocatable<TriviallyMoveConstructible>::value);
+FOLLY_POP_WARNING
 }
 
 TEST(Traits, bitAndInit) {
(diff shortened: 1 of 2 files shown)


folly/test/TraitsTest.cpp
```diff
@@ -98,6 +98,14 @@ TEST(Traits, unset) {
   EXPECT_TRUE(IsRelocatable<F4>::value);
 }
 
+TEST(Traits, triviallyMoveConstructible) {
+  struct TriviallyMoveConstructible {
+    TriviallyMoveConstructible(const TriviallyMoveConstructible&) = delete;
+    TriviallyMoveConstructible(TriviallyMoveConstructible&&) = default;
```
yfeldblum: This function being defined but never called at runtime is a violation of `-Wunneeded-member-function`, which is being shown internally.

JoeLoser: That's unfortunate. Is this just for Clang? Do you suggest just pragma push/popping it?

We should also add this flag to `FollyCompilerUNIX.cmake`. Then, there is a chance that Travis would fail on this error too unless it is Clang only (as we only use GCC for Travis).

folly/test/TraitsTest.cpp
```diff
@@ -98,6 +98,14 @@ TEST(Traits, unset) {
   EXPECT_TRUE(IsRelocatable<F4>::value);
 }
 
+TEST(Traits, triviallyMoveConstructible) {
+  struct TriviallyMoveConstructible {
+    TriviallyMoveConstructible(const TriviallyMoveConstructible&) = delete;
```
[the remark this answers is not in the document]
JoeLoser: Good lint rule to have. I just fixed them up.

folly/test/TraitsTest.cpp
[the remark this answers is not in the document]
JoeLoser: Fixed 👍

yfeldblum: This is a [clang warning](https://releases.llvm.org/7.0.0/tools/clang/docs/DiagnosticsReference.html#wunneeded-member-function) but the build does not fail because of it. This warning violation is (currently) being surfaced as if it were a lint rule violation.

folly/test/TraitsTest.cpp
```diff
@@ -98,6 +98,19 @@ TEST(Traits, unset) {
   EXPECT_TRUE(IsRelocatable<F4>::value);
 }
 
+TEST(Traits, triviallyMoveConstructible) {
+#pragma GCC diagnostic push
```
yfeldblum: Please use our wrapper macros in `folly/Portability.h` - this would otherwise fail MSVC builds.

An alternative option would be to make a new header `folly/test/TraitsTestHelpers.h` with this type, but that is higher-ceremony than this.

JoeLoser: Thanks, I just switched to using `FOLLY_CLANG_DISABLE_WARNING` in this test case only. I agree with you that this doesn't warrant the creation of a new header file.

yfeldblum:
We had to revert this since some places with gcc49 broke.

JoeLoser:
When we get everyone off of GCC 4.9, think we can try landing this again?

yfeldblum:
That's the plan!

JoeLoser:
If you think the recent change to keep GCC5 as min compiler, feel free to try landing this again. 🤞

yfeldblum:
Per the other PR, I think the issue remains with clang using old libstdc++ in some Android NDKs.

PR #1037, opened 2019-02-28
Customize growth strategy of FBVector

JoeLoser:
Summary:
- Allow users to customize the growth stratrgy of FBVector by means of a
  new member function in the allocator type.
- If the allocator type passed to `fbvector` class template has a member
  function of the signature `size_t calculate_capcity(size_t)`, then this
  function will be called every time `fbvector` needs to increase its
  internal capacity.
- The semantics of `calculate_capacity` are that it takes the old
  capacity as its input and returns the new size.
- If an allocator type does not implement `calculate_capacity` member
  function, the same capacity calculations are done as in the present day
  implementation.

### DIFF
--- a/folly/FBVector.h
+++ b/folly/FBVector.h
@@ -1194,7 +1194,7 @@ class fbvector {
   // std::vector implements a similar function with a different growth
   //  strategy: empty() ? 1 : capacity() * 2.
   //
-  // fbvector grows differently on two counts:
+  // fbvector grows differently on three counts:
   //
   // (1) initial size
   //     Instead of growing to size 1 from empty, fbvector allocates at least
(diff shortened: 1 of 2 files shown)


yfeldblum:
What needs this?

Why does this behavior need to be on the allocator?

JoeLoser:
After some more thought, the use cases are very limited it seems. The comment suggested there was a desire for growth as a customization point to users -- but I just now realized the comment is from 2012, so it's quite outdated.

I could *maybe* imagine a use case where someone wants a quick implementation of a fixed capacity `fbvector` which would throw on `calculate_capacity()`. But, for something like that, it would be a poor container design as it would still have public member functions like `push_back` to get you in that situation which can be easily avoided. In practice, fixed containers like that would just have `operator[]` but not expose `push_back`. Most people using a container like that also disable exceptions, so it's a weak argument.

Like you pointed out, if we did want to implement growth as a customization point, it would be less obscure to just make it a policy on the class template itself, rather than on the allocator. Otherwise, most users would just inherit from `std::allocator` and implement `calculate_capacity()` and not change anything else. That leads one to easily conclude it doesn't belong in the allocator.

As such, I'm going to close the PR. It's a very niche customization point that I don't have any immediate use cases for.

nbronson:
It would actually be useful to adjust the capacity to minimize fragmentation waste (make the allocation and malloc_usable_size as close as possible). I don't think this PR is the cleanest way to do this, but it should be something we consider.

PR #1041, opened 2019-03-02
Fix bootstrap-osx-homebrew.sh to work with CMake builds

### LINKED ISSUE #1038: Mac OS build, help needed.
Hi.

Trying to build folly on Mac OS. I would like to contribute, so I need to be able to build tests and benchmarks - I cannot just take the brew installation.

bootstrap-osx-homebrew.sh fails with "autoreconf: 'configure.ac' or 'configure.in' is required"
I googled - seems like running it is no longer relevant.

CMAKE fails due to not being able to find dependencies:
```
-- Could NOT find DOUBLE_CONVERSION (missing: DOUBLE_CONVERSION_LIBRARY DOUBLE_CONVERSION_INCLUDE_DIR) 
-- Using third-party bundled double-conversion
-- Could NOT find LIBGFLAGS (missing: LIBGFLAGS_LIBRARY LIBGFLAGS_INCLUDE_DIR) 
-- Could NOT find LIBGLOG (missing: LIBGLOG_LIBRARY LIBGLOG_INCLUDE_DIR) 
-- Found libevent: /usr/local/lib/libevent.dylib
-- Found OpenSSL: /usr/lib/libcrypto.dylib (found version "1.0.2p") 
-- Looking for ASN1_TIME_diff
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIBGFLAGS_INCLUDE_DIR
   used as include directory in directory /space/folly/_build/CMakeFiles/CMakeTmp
   used as include directory in directory /space/folly/_build/CMakeFiles/CMakeTmp
   used as include directory in directory /space/folly/_build/CMakeFiles/CMakeTmp
LIBGFLAGS_LIBRARY
    linked by target "cmTC_e2f29" in directory /space/folly/_build/CMakeFiles/CMakeTmp

CMake Error at /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/CheckFunctionExists.cmake:72 (try_compile):
  Failed to configure test project build system.
  CMake/folly-deps.cmake:75 (check_function_exists)
  CMakeLists.txt:85 (include)


```

Could you please give instructions on how to build folly: from clear download to running tests and benchmarks?

JoeLoser:
Summary:
- The `bootstrap-osx-homebrew.sh` script is out of date. It previously
  worked when Folly used Autotools, but was not updated when CMake
  replaced Autotools for the build system.
- Update dependencies to install.
- Do not export environment variables when it is not needed; simply pass the
  defines needed directly to the `cmake` invocation.
- Update `README.md` to specify correct dependencies with CMake build
  for both Homebrew and Macports.

Closes #1038

### DIFF
--- a/folly/build/bootstrap-osx-homebrew.sh
+++ b/folly/build/bootstrap-osx-homebrew.sh
@@ -5,8 +5,7 @@
 # fail fast
 set -e
 
-BASE_DIR="$(cd "$(dirname -- "$0")"/.. ; pwd)"  # folly/folly
-cd "$BASE_DIR"
+BUILD_DIR=${BUILD_DIR:-_build}
 
 # brew install alias
 brew_install() {
@@ -15,45 +14,52 @@ brew_install() {
 
 # install deps
 install_deps() {
-	# folly deps
-	dependencies=(autoconf automake libtool pkg-config double-conversion glog gflags boost libevent xz snappy lz4 jemalloc openssl)
+    # folly deps
+    dependencies=(
+        boost
+        cmake
+        double-conversion
+        gflags
+        glog
+        jemalloc
+        libevent
+        lz4
+        openssl
+        pkg-config
+        snappy
+        xz
+    )
 
-	# fetch deps
-	for dependency in "${dependencies[@]}"; do
-		brew_install "${dependency}"
-	done
+    # fetch deps
+    for dependency in "${dependencies[@]}"; do
+        brew_install "${dependency}"
+    done
 }
 
-# set env flags
-export_flags() {
-	# fetch opt dirs
-	OPT_GFLAGS=$(brew --prefix gflags)
-	OPT_OPENSSL=$(brew --prefix openssl)
-
-	# export flags
-	export LDFLAGS=-L${OPT_OPENSSL}/lib
-	export OPENSSL_INCLUDES=-I${OPT_OPENSSL}/include
-	export GFLAGS_LIBS=-L${OPT_GFLAGS}/lib
-	export GFLAGS_CFLAGS=-I${OPT_GFLAGS}/include
-}
-
-# now the fun part
 install_deps
-export_flags
-autoreconf -ivf
-./configure --disable-silent-rules --disable-dependency-tracking
+
+# Allows this script to be invoked from anywhere in the source tree but the
+# BUILD_DIR we create will always be in the top level folly directory
+TOP_LEVEL_DIR="$(cd "$(dirname -- "$0")"/../.. ; pwd)"  # folly
+cd "$TOP_LEVEL_DIR"
+mkdir -p "${BUILD_DIR}"
+cd "${BUILD_DIR}"
+
+OPENSSL_INCLUDES=$(brew --prefix openssl)/include
+cmake \
+    -DOPENSSL_INCLUDE_DIR="${OPENSSL_INCLUDES}" \
+    "$@" \
+    ..
 
 # fetch googletest, if doesn't exist
-pushd test
 GTEST_VER=1.8.0
 GTEST_DIR=gtest-${GTEST_VER}
 if [ ! -d ${GTEST_DIR} ]; then
-	mkdir ${GTEST_DIR}
+    mkdir ${GTEST_DIR}
     curl -SL \
     	https://github.com/google/googletest/archive/release-${GTEST_VER}.tar.gz | \
     	tar -xvzf - --strip-components=1 -C ${GTEST_DIR}
 fi
-popd
 
 # make, test, install
 make
--- a/README.md
+++ b/README.md
@@ -154,9 +154,10 @@ folly is available as a Formula and releases may be built via `brew install foll
 You may also use `folly/build/bootstrap-osx-homebrew.sh` to build against `master`:
 
 ```
-  cd folly
-  ./build/bootstrap-osx-homebrew.sh
+  ./folly/build/bootstrap-osx-homebrew.sh
 ```
+
+This will create a build directory `_build` in the top-level.
 
 #### OS X (MacPorts)
 
@@ -164,9 +165,8 @@ Install the required packages from MacPorts:
 
 ```
   sudo port install \
-    autoconf \
-    automake \
     boost \
+    cmake \
     gflags \
     git \
     google-glog \
@@ -174,8 +174,9 @@ Install the required packages from MacPorts:
     libtool \
     lz4 \
     lzma \
-    scons \
+    openssl \
     snappy \
+    xz \
     zlib
 ```
 
@@ -193,9 +194,10 @@ Download and install folly with the parameters listed below:
 
 ```
   git clone https://github.com/facebook/folly.git
-  cd folly/folly
-  autoreconf -ivf
-  ./configure CPPFLAGS="-I/opt/local/include" LDFLAGS="-L/opt/local/lib"
+  cd folly
+  mkdir _build
+  cd _build
+  cmake ..
   make
   sudo make install
 ```


yfeldblum:
Thanks!

JoeLoser:
Please reimport as I just pushed up a change to explicitly specify the gflags lib path to the `cmake` invocation as based on discussion from https://github.com/facebook/folly/issues/1038

JoeLoser:
Just pushed up a commit to modify the commit message to indicate I also fixed the `README.md` to correctly specify the deps for Homebrew/Macports build. There was also an issue in the script which wasn't calling `install_deps` which is fixed.

JoeLoser:
Fixed the `shellcheck` warnings here. Similar story as in https://github.com/facebook/folly/pull/1042 where I had to fix some existing code which was flagged by `shellcheck`.

calebmarchent:
@JoeLoser; had a look at the diff today, looks quite sensible; but I want to test it before final approval; I expect that'll be daytime UK time tomorrow.

JoeLoser:
@calebmarchent sounds good. Thanks for the review!

calebmarchent:
Update, on first pass got:

```
Error: openssl 1.0.2p is already installed
To upgrade to 1.0.2q, run `brew upgrade openssl`
```

I did the upgrade manually and got, the following error on CMake:

```
CMake Error at CMake/FindLibEvent.cmake:29 (message):
  Could NOT find libevent.
```

JoeLoser:
@calebmarchent yes, you need to explicitly `brew link libevent` as it stated. Do you think we should do this to make it easier for users? I just noticed now that the `Fizz` build explicitly links in their scripts: https://github.com/facebookincubator/fizz/blob/master/fizz/mac-build.sh#L55

As for the `unary_function`, yeah, it is deprecated and fully removed in C++17. Folly uses `-std=gnu++1z` so it shouldn't be an issue.

folly/build/bootstrap-osx-homebrew.sh
```diff
@@ -10,41 +10,52 @@ cd "$BASE_DIR"
...
 install_deps
-export_flags
-autoreconf -ivf
-./configure --disable-silent-rules --disable-dependency-tracking
+
+(
+    cd ..
+    mkdir -p _build
+    cd _build
+)
```
calebmarchent: If we are going to build in a subdir, then the steps below need to happen within that subdirectory. 
I think this closing bracket needs to move to the bottom of the file.

calebmarchent:
Hi @JoeLoser, I've spent a little time trying to build and test folly with this change, as yet without success. It would be helpful for me if you could provide some detail of what testing you have done with this diff and the results you found?

Despite attempting to configure cmake within the `folly/_build` directory. I see files being generated within the `folly/` tree, which makes me think the CMake configuration is not correct; I even tried breaking up the steps of the install and calling them individually, but observed similar issues. 
I tried building a different cmake project to check my build environment: (https://github.com/no1msd/mstch); which built fine without contaminating the source tree?

For reference, I am building with Mojave 10.14.3 (18D109); XCode is installed, Homebrew 2.0.3.

folly/build/bootstrap-osx-homebrew.sh
```diff
@@ -10,41 +10,52 @@ cd "$BASE_DIR"
...
+
+OPENSSL_INCLUDES=$(brew --prefix openssl)/include
+GFLAGS_ROOT=$(brew --prefix gflags)
+GFLAGS_INCLUDES=${GFLAGS_ROOT}/include
+GFLAGS_LIB=${GFLAGS_ROOT}/lib
+cmake \
+    -DOPENSSL_INCLUDE_DIR="${OPENSSL_INCLUDES}" \
+    -DLIBGFLAGS_INCLUDE_DIR="${GFLAGS_INCLUDES}" \
+    -DLIBGFLAGS_LIBRARY="${GFLAGS_LIB}" \
+    ..
```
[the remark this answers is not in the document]
JoeLoser: Agreed, with the exception of OpenSSL.

folly/build/bootstrap-osx-homebrew.sh
JoeLoser: Good catch. I pushed up the shellcheck fixes without fully testing them unfortunately as I was running off to a flight. I apologize for that.

You are certainly right that the lifetime of that bash subprocess would need to be for both the configure and make invocations.

I've refactored the script a bit to not require the `cd ..` and hence the 

```
(
cd ..
// ...
)
```

isn't needed. We now create a build directory in the top level dir, rather than using the previous `BASE_DIR` (folly/folly) and then doing `cd ..` to get us to `folly` top level.

JoeLoser:
Hi @calebmarchent 

I was manually running the script on Mac OS X 10.13.x/10.14.y.  I also manually uninstalled some dependencies via Homebrew to make sure things weren't completely broken. 

As mentioned in a previous comment, I broke the script when I fixed the shellcheck warnings as the lifetime of the subshell was only for the `cmake/configure` but not for `make`. This was a bug and has been addressed in a better manner to not even require the subshell.

The source tree `(folly/folly)` should not be contaminated with the build files as we build in a subdirectory. It was possible this was the case when I broke the script with the subshell only running for `cmake` and not `make`, but that is fixed now. Please try again -- sorry for the inconvenience.

JoeLoser:
@calebmarchent have you had a chance to look at the most recent changes? Would you recommend doing a brew link for each dependency like Fizz does?

calebmarchent:
> @calebmarchent have you had a chance to look at the most recent changes? Would you recommend doing a brew link for each dependency like Fizz does?

Been out, but will try and have a look today. At the moment it is not a case that it "just works"; I need to find out why not. Currently I suspect its down to the original state of the Mac am testing on.

calebmarchent:
Further update... I got it to compile; but with the following change in CMake config:

```
cmake -DOPENSSL_INCLUDE_DIR=$(brew --prefix openssl)/include -DFOLLY_HAVE_WEAK_SYMBOLS=ON ..
```
It looks like on my system weak symbols support detection is not working, see the CMake log:

```
-- Performing Test FOLLY_HAVE_WEAK_SYMBOLS
-- Performing Test FOLLY_HAVE_WEAK_SYMBOLS - Failed
```
I need to look a little further into why this is failing on my system..

folly/build/bootstrap-osx-homebrew.sh
```diff
@@ -5,46 +5,53 @@
...
+BUILD_DIR=${BUILD_DIR:-_build}
 
 # brew install alias
 brew_install() {
-    brew install $@ || brew upgrade $@
+    brew install "$@" || brew upgrade "$@"
 }
 
 # install deps
 install_deps() {
```
yfeldblum: The indentation in this function has gotten inconsistent.

JoeLoser:
Thanks for taking another look and playing with it, @calebmarchent.

I think the weak symbol detection in `FollyConfigChecks.cmake` doesn't work as intended for Mac OS X. Whether I use `AppleClang 10.0.0.10001145` or Clang 7.0.1, I do get the failed to have weak symbols during the configure step. Regardless, I can properly compile and link folly (building as static lib, which is the default). You may get some warnings as expected from `ranlib` as expected though since some object files will have no code due to platform-dependent `#ifdef`. In these cases, the resulting object file should have no code, and hence no symbols. For example, when using AppleClang (the default for this script as we're not setting `CMAKE_CXX_COMPILER`), I get:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Poly.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Zstd.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Baton.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Mutex.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SharedMutex.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(ChecksumDetail.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SocketFileDescriptorMap.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Builtins.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Dirent.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Fcntl.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(PThread.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Sched.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Semaphore.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Sockets.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Stdio.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SysFile.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SysMman.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SysResource.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SysStat.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(SysTime.cpp.o) has no symbols
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: /usr/local/lib/libfolly.a(Unistd.cpp.o) has no symbols
```

If you are getting hard link errors, you may have `binutils` linked via `homebrew`. I have run into this problem before. It results in mismatched versions of `ar` when doing the archiving.  This can be seen by running `lip info <path_to_folly_static_lib>`. You can fix this by doing `brew unlink binutils` and it will find the right "system default" `ar` in your PATH which is what CMake is expecting to use, unless you explicitly set `CMAKE_AR` and other associated options, which Folly does not do.

In summary:

- Weak symbol detection check failing at configure time is nothing new for this script, i.e. I did not introduce new behavior.
- The lack of support of weak symbols should not result in a failure to link or compile when building folly as a static library.

I can dig into the weak symbol detection outside of this PR and see what is going on.

Also, if we wish to split the documentation changes to a separate PR, I am happy to do so.

calebmarchent:
Hi @JoeLoser thanks for all the additional info; I am happy with the state of the code now; so just waiting for the stamp I need to get your change imported. 

_I do agree the target should build with/without weak symbol support; furthermore, I checked the detection code that had been in the old autotools version; which was the same mechanism. To get a more reliable check-point; we really need a CI build for Mac; to ensure the config of the system in which the build is tested is known and controlled. Once your PR is merged. we should give us a good foundation to set that up._

PR #1042, opened 2019-03-02
Update Ubuntu build scripts to work with CMake build

JoeLoser:
Summary:
- Update Ubuntu 14.04 build script to use CMake instead of Autotools.
- Rename Ubuntu 14.04 build script to use hyphens rather than underscores to match POSIX standard for bash scripts.
- The Ubuntu 14.04 build script is now meant to be run from the top-level rather than from inside the `folly` source tree.
- Remove the Ubuntu 12.04 build scripts. It was previously only needed for Travis CI, but Travis CI isn't using it anymore.


yfeldblum:
We run [shellcheck](https://github.com/koalaman/shellcheck) internally and it is reporting lint warnings.

JoeLoser:
Just fixed up the shellcheck lint warnings; glad you guys run it 👍 

It seems that when `shellcheck` lint was introduced internally, we didn't fix existing scripts? For example, it flagged the `which fpm` line which exists in top of tree. Obviously not a biggie here, but wanted to bring it to your attention in case it's worth running and fixing the existing shell scripts internally.

Issue #1011, opened 2019-02-04
Compile error on Linux 2.6.32-754.6.3.el6.x86_64 /gcc 7.4/NetOps.cpp


smahapatra1:
In function ‘int folly::netops::sendmmsg(folly::NetworkSocket, mmsghdr*, unsigned int, int)’:
.../third_party/folly@master/folly/net/NetOps.cpp:375:36: error: ‘::sendmmsg’ has not been declared
   return wrapSocketFunction<int>(::sendmmsg, socket, msgvec, vlen, flags);
                                    ^~~~~~~~
.../third_party/folly@master/folly/net/NetOps.cpp:375:36: note: suggested alternative:
.../third_party/folly@master/folly/net/NetOps.cpp:369:5: note:   ‘folly::netops::sendmmsg’
 int sendmmsg(

The problem is that sendmmsg is defined in Linux 3.0. Either it should be found automatically or at least define have:
#ifndef FOLLY_HAVE_SENDMMSG
#define FOLLY_HAVE_SENDMMSG 1
#endif

yfeldblum:
We are not testing on Linux 2.6 anymore.

I might consider a PR that updated the preprocessor check: https://github.com/facebook/folly/blob/v2019.02.04.00/folly/net/NetOps.h#L74

PR #1043, opened 2019-03-04
Compile error on Linux 2.6.32-754.6.3.el6.x86_64 /gcc 7.4/NetOps.cpp

deryck1031:
Compile error on Linux 2.6.32-754.6.3.el6.x86_64 /gcc 7.4/NetOps.cpp (issue #1011)

added below in NetOps.h

#ifndef FOLLY_HAVE_SENDMMSG
#define FOLLY_HAVE_SENDMMSG 1
#endif

### DIFF
--- a/FollyConfigChecks.cmake
+++ b/FollyConfigChecks.cmake
@@ -1,0 +1,245 @@
+include(CheckCXXSourceCompiles)
+include(CheckCXXSourceRuns)
+include(CheckFunctionExists)
+include(CheckIncludeFileCXX)
+include(CheckSymbolExists)
+include(CheckTypeSize)
+include(CheckCXXCompilerFlag)
+
+CHECK_INCLUDE_FILE_CXX(jemalloc/jemalloc.h FOLLY_USE_JEMALLOC)
+
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
+  # clang only rejects unknown warning flags if -Werror=unknown-warning-option
+  # is also specified.
+  CHECK_CXX_COMPILER_FLAG(
+    -Werror=unknown-warning-option
+    COMPILER_HAS_UNKNOWN_WARNING_OPTION)
+  if (COMPILER_HAS_UNKNOWN_WARNING_OPTION)
+    set(CMAKE_REQUIRED_FLAGS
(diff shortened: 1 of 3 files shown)


yfeldblum:
Why is this a fix?

deryck1031:
This is to support older versions of Linux (2.6 in this case) that does not have sendmmsg.

folly/net/NetOps.h
```diff
@@ -77,8 +77,10 @@ struct mmsghdr {
   unsigned int msg_len;
 };
 #else
+#ifndef FOLLY_HAVE_SENDMMSG
```
yfeldblum: But why is *this* the fix?

V.s. updating the `#if` line just above.

deryck1031: struct mmsghdr is already defined in /usr/include/bits/socket.h.  If update #if statement, we will get 'redefinition of struct mmsghdr' compilation error.

yfeldblum: What did you try?

deryck1031: I tried below:

#if (!__linux__) || (defined(__ANDROID__) && (__ANDROID_API__ < 21)) || (defined(HAS_NO_SENDMMSG))

yfeldblum: We may need to move this to a config check in `folly/CMake/FollyConfigChecks.cmake`.

deryck1031: Make sense, I think we need to make below changes:

1. remove define in NetOps.h
-#else
-#define FOLLY_HAVE_SENDMMSG 1

2. add  FOLLY_HAVE_SENDMMSG  to  CMake/folly-config.h.cmake
+#cmakedefine FOLLY_HAVE_SENDMMSG 1

3. update CMake/FollyConfigChecks.cmake
+check_symbol_exists(sendmmsg sys/socket.h FOLLY_HAVE_SENDMMSG)

I can create a PR if above changes look good to you.

yfeldblum: Yes something like this. Can force-update this PR if desired.

deryck1031: there was something wrong with my old forked repo, so I have created a new PR (1053).
https://github.com/facebook/folly/pull/1053

PR #1048, opened 2019-03-08
Cut remove_cvref_t from pushmi/traits.h

JoeLoser:
Summary:
- Remove `remove_cvref_t` alias template from `pusmhi/traits.h` in favor of
  `folly::remove_cvref_t` from `folly/Traits.h` which is the same
  definition.

### DIFF
--- a/folly/experimental/pushmi/traits.h
+++ b/folly/experimental/pushmi/traits.h
@@ -75,9 +75,6 @@ PUSHMI_INLINE_VAR constexpr int sum_v = detail::sum_impl<Is...>();
 template <class...>
 struct typelist;
 
-template <class T>
-using remove_cvref_t = std::remove_cv_t<std::remove_reference_t<T>>;
-
 PUSHMI_CONCEPT_DEF(
   template(class... Args)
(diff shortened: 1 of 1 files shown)


ericniebler:
Thanks, @JoeLoser !

JoeLoser:
@ericniebler didn't realize `pushmi` was getting worked on at https://github.com/facebookresearch/pushmi and then just synced with folly. This sort of PR makes the syncing annoying as I suspect `pushmi` doesn't want to depend on folly for such a little thing like this.

What do you think?

ericniebler:
@JoeLoser we're sync'ing pushmi manually to facebookresearch/pushmi to facilitate collaboration with some external teams that don't want to (or can't) take a dependence on all of Folly. This is just a temporary state of affairs. Don't worry about `folly::remove_cvref_t`. It's a drop in the bucket.

(In this case, yes, some improvements appeared in the facebookresearch/pushmi repo before appearing here. That will not be typical, though. It was a huge diff that was taking a while to pass internal code-review, and we wanted our external collaborators to have the improvements early.)

PR #1053, opened 2019-03-13
moved FOLLY_HAVE_SENDMMSG to FollyConfigChecks.cmake for issue #1011

deryck1031:
Compile error on Linux 2.6.32-754.6.3.el6.x86_64 /gcc 7.4/NetOps.cpp (issue #1011)

### DIFF
--- a/folly/net/NetOps.h
+++ b/folly/net/NetOps.h
@@ -18,6 +18,7 @@
 
 #include <cstdint>
 
+#include <folly/Portability.h>
 #include <folly/net/NetworkSocket.h>
 #include <folly/portability/IOVec.h>
 #include <folly/portability/SysTypes.h>
@@ -71,7 +72,7 @@
 #define UDP_MAX_SEGMENTS (1 << 6UL)
 #endif
 
-#if (!__linux__) || (defined(__ANDROID__) && (__ANDROID_API__ < 21))
+#if !(FOLLY_HAVE_RECVMMSG || FOLLY_HAVE_SENDMMSG)
 struct mmsghdr {
   struct msghdr msg_hdr;
   unsigned int msg_len;
--- a/folly/net/NetOps.h
+++ b/folly/net/NetOps.h
@@ -76,8 +76,6 @@ struct mmsghdr {
   struct msghdr msg_hdr;
   unsigned int msg_len;
 };
-#else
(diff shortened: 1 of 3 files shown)


folly/net/NetOps.h
```diff
@@ -76,8 +76,6 @@ struct mmsghdr {
   struct msghdr msg_hdr;
   unsigned int msg_len;
 };
-#else
```
yfeldblum: I think we should change the test above to just `#if !FOLLY_HAVE_SENDMMSG`.

deryck1031: This will cause issue for older version of Linux that doesnt have sendmmsg.  Code wont compile due to multiple define of mmsghdr:

net/NetOps.h:76:8: error: redefinition of 'struct mmsghdr'
 struct mmsghdr {
        ^~~~~~~
In file included from /usr/include/sys/socket.h:40:0,
                 from /usr/include/netinet/in.h:25,

yfeldblum: Do some versions of Linux have `mmsghdr` but not `sendmmsg`?

deryck1031: The issue is that mmsghdr is used by both sendmmsg and recvmmsg.  recvmmsg was added to Linux 2.6.33 but sendmmsg was added to Linux 3.0.  Between those 2 versiosn, there is mmsghdr but no sendmmsg.

yfeldblum: Alright, let's do a cmake check for `recvmmsg` and a check for `sendmmsg`.

We will use the result of the check for `recvmmsg` in the test above conditionally defining `mmsghdr`, and the result of the check for `sendmmsg` for guarding calls to `sendmmsg`.

deryck1031: sounds good, will add FOLLY_HAVE_RECVMMSG and update the if statement

deryck1031: I have committed the changes.  Another thing that I have to add is to include Portability.h so folly-config.h can be included.

deryck1031: Build failed.  But not sure if is related to any changes from this PR after went through the build log

yfeldblum: Build failure appears unrelated.

PR #1055, opened 2019-03-14
Use common StrictAllOf in Expected.h and ExceptionWrapper.h

JoeLoser:
Summary:
- Both `Expected.h` and `ExceptionWrapper.h` define a similar type trait
  to check if multiple types satisfy a type trait. This is done through
  a wrapper around `StrictConjunction`.
- Move this common definition of the type trait: `StrictAllOf` into
  `folly/Traits.h` and fix up the call sites to use this.
- Add tests for `StrictAllOf`.


folly/Traits.h
```diff
@@ -500,6 +500,9 @@ template <class... Ts>
 struct StrictConjunction
     : std::is_same<Bools<Ts::value...>, Bools<(Ts::value || true)...>> {};
 
+template <template <class...> class Trait, class... Ts>
```
yfeldblum: Naming - `Trait` is odd - don't know what that's supposed to mean. Just `F` (for `Function`) would be fine.

Also why `template <class...> class` v.s. `template <class> class`?

folly/Traits.h
```diff
@@ -500,6 +500,9 @@ template <class... Ts>
...
+using StrictAllOf = StrictConjunction<Trait<Ts>...>;
```
yfeldblum: Naming - `StrictConjunctionBy` probably.

JoeLoser: Fixed. Thanks.

folly/Traits.h
JoeLoser: Both the name (`Trait`) and the signature with the template template just came from `Expected.h` as is. I suspect the intention was that our existing call sites use standard type traits as the metafunction, so naming the metafunction as `Trait` isn't too far from wrong, but this type trait is more general than that.

I agree that since this is a simple metafunction accepting only one type, we should name it as such (`F`) and make it so the template template parameter is just of one type for the class template.

I've made those changes as requested 👍

JoeLoser:
@yfeldblum should be ready to re-import again now. Thanks.

I saw a few places where this trait can be utilized which will come in a follow-up MR when this lands.

folly/ExceptionWrapper-inl.h
```diff
@@ -657,17 +657,15 @@ inline bool exception_wrapper::with_exception(Fn fn) const {
 
 template <class... CatchFns>
 inline void exception_wrapper::handle(CatchFns... fns) {
-  using AllStdEx =
-      exception_wrapper_detail::AllOf<IsStdException, arg_type<CatchFns>...>;
+  using AllStdEx = StrictConjunctionBy<IsStdException, arg_type<CatchFns>...>;
```
yfeldblum: Hmm, so thinking again, this could just be

```
using AllStdEx = StrictConjunction<IsStdException<arg_type<CatchFns>>...>;
```

JoeLoser: Good spot! I agree. Just changed it and the other call site in this file.

JoeLoser:
@yfeldblum anything else I can do with this PR? While there are only a few current call sites using `StrictConjunctionBy`, I have seen some other call sites which could be improved with it from the code I have read the last few days.

JoeLoser:
After reading [P0013](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0013r1.html) which was the paper proposing the logical type traits, I still think this is a useful utility to prevent writing a separate variadic trait using conjunction that wraps each individual trait.

Have you given any more thought/discussion to this?

JoeLoser:
@yfeldblum think it's worth splitting the `ExceptionWrapper-inl` changes if there isn't desire for this metafunction `StrictConjunctionBy` in `folly/Traits.h`? Here is a branch which does that: https://github.com/JoeLoser/folly/commit/22447f49964eb5bed4776f9d012e8e4efd7fb4a4

yfeldblum:
It seems odd to have this for only `StrictConjunction` - can we do the other 3 as well?

JoeLoser:
Sure, no problem. By "the other 3", I assume you mean `StrictDisjunctionBy`, `Conjunction`, and `Disjunction`? This would provide nice symmetry, which I'm in favor of.

yfeldblum:
Right.

JoeLoser:
Just added the symmetric flavors of logical traits with metafunctions as requested.

JoeLoser:
ping @yfeldblum

yfeldblum:
Ah thanks!

PR #1063, opened 2019-03-16
Contains with precomputed hash for F14Map and F14Set

JoeLoser:
Summary:
- Implement `contains(key_type const& key, F14HashToken const& token)` similar
  to the C++20 `contains(key_type const& key, size_t hash)` interface.

Note:
- This is slightly different than the existing interfaces on F14
  containers, as they usually accept the `F14HashToken` as the first
  function parameter. We chose it as the second parameter to be "closer"
  to the C++20 member functions which take a precomputed hash as the
  second argument.

### DIFF
--- a/folly/container/F14Map.h
+++ b/folly/container/F14Map.h
@@ -833,6 +833,18 @@ class F14BasicMap {
     return !table_.find(key).atEnd();
   }
 
+  FOLLY_ALWAYS_INLINE bool contains(
+      F14HashToken const& token,
+      key_type const& key) const {
+    return !table_.find(token, key).atEnd();
+  }
+
+  template <typename K>
+  FOLLY_ALWAYS_INLINE EnableHeterogeneousFind<K, bool> contains(
+      F14HashToken const& token, K const& key) const {
+    return !table_.find(token, key).atEnd();
+  }
+
   std::pair<iterator, iterator> equal_range(key_type const& key) {
     return equal_range(*this, key);
   }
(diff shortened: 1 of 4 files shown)


JoeLoser:
@nbronson if this is something we want, I can implement this for transparent key types too. This is essentially a follow-up addition on top of https://github.com/facebook/folly/commit/94a77ab3adee9f3ab184e23fe0c38d4cb5086635. 

How do you feel about the API of the precomputed hash being first or second? I don't feel strongly and am happy to move it to be the first parameter. In general, I would prefer it to be as similar to the standard library utilities where possible as a common use case is experimenting with different containers and having it be as close to a drop-in-replacement is handy.

yfeldblum:
Should:
* Add heterogeneous-lookup overloads.
* Be consistent with the rest of the F14 API. If we were to make the F14 API look closer to the new std unordered containers API, we will do it for all member functions at the same time.

JoeLoser:
Thanks for the feedback @yfeldblum. I'll add the heterogenous lookup overload for both `F14Map` and `F14Set` and change the parameter order to be consistent with the rest of the F14 API later tonight.

JoeLoser:
> Should:
> 
> * Add heterogeneous-lookup overloads.
> * Be consistent with the rest of the F14 API. If we were to make the F14 API look closer to the new std unordered containers API, we will do it for all member functions at the same time.

Both of those are done now; please re-review. Thanks.

yfeldblum:
Thanks!

PR #1065, opened 2019-03-17
Use feature test macro for deduction guide in Replaceable.h

JoeLoser:
Summary:
- Replace a check for deduction guide support to use a feature test macro
  for deduction guides rather than a check using `__cplusplus` directive.
- The existing check using the `__cplusplus` directive is not correct
  for GCC 5.1. With GCC 5.1, `__cplusplus > 201402L` will be true, but the check
  for `#if __cpp_deduction_guides >= 201703` will be false. So, it is
  not valid to just rely on the `__cplusplus` directive to infer
  deduction guide support.
- We extend the check to use the feature test macro except in the case
  of MSVC, which we infer the feature is supported. MSVC does not define
  feature test macros, despite having implemented the feature.

### DIFF
--- a/folly/Replaceable.h
+++ b/folly/Replaceable.h
@@ -640,8 +640,7 @@ class alignas(T) Replaceable
   aligned_storage_for_t<T> storage_[1];
 };
 
-#if __cplusplus > 201402L
-// C++17 allows us to define a deduction guide:
+#if __cpp_deduction_guides >= 201703 || _MSC_VER
 template <class T>
 Replaceable(T)->Replaceable<T>;
 #endif


folly/Replaceable.h
```diff
@@ -640,8 +640,7 @@ class alignas(T) Replaceable
   aligned_storage_for_t<T> storage_[1];
 };
 
-#if __cplusplus > 201402L
-// C++17 allows us to define a deduction guide:
+#if __cpp_deduction_guides >= 201703
```
yfeldblum: Seems like MSVC2017 does not define this name but does support deduction guides. Can guard this with `|| _MSC_VER`. (We don't support earlier versions of MSVC anymore.)

JoeLoser: That's a bit annoying from a QoI perspective. Based on the `pushmi` usages of `__cpp_deduction_guides`, we may not support things in `experimental` with MSVC?

I can use `|| _MSC_VER` here, but we may want to put this expression into a `constexpr bool` in `Portability.h` after we acquire some more uses of deduction guides.

JoeLoser:
Updated the commit message to better reflect that this is a bug as discussed in https://github.com/facebook/folly/pull/1070

Issue #755, opened 2018-02-06
[Question] Interest in .clang-format for open source project


JoeLoser:
Is there any interest in using a `.clang-format`? I suspect there is already one used internally. Is it possible to promote it to the open source tree?

JoeLoser:
@wez @calebmarchent

Is it possible to auto sync the `.clang-format` to all of the open source facebook repos using the recent `fbcode_builder` scripts now that `fbcode_builder` is more widely used?

wez:
we're talking about that internally (@yfeldblum kicked off a thread on this).  I think we can do this, but there's a bit of fiddly config code to get put in place that makes it a slightly bigger project than it sounds.

JoeLoser:
Thanks for kicking off the discussions internally, @yfeldblum. Also, thanks for the update @wez and some scope for this. 

I'd love to see this worked on when you guys get time. I'm sure it would make other open source developers working on FB repos happy too 👍

PR #1069, opened 2019-03-19
Implement erase_if for F14 containers

JoeLoser:
Summary:
- For each F14 container, implement `erase_if(F14Container&, Predicate)` similar to C++20 `erase_if(std::unordered_map&, Predicate)`.
- The semantics of `erase_if` are that it erases all elements in the container which satisfy the given predicate.

### DIFF
--- a/folly/container/F14Map.h
+++ b/folly/container/F14Map.h
@@ -1490,4 +1490,28 @@ void swap(
   lhs.swap(rhs);
 }
 
+template <typename K, typename M, typename H, typename E, typename A,
+          typename Pred>
+void erase_if(F14ValueMap<K, M, H, E, A>& c, Pred pred) {
+  f14::detail::erase_if_impl(c, pred);
+}
+
+template <typename K, typename M, typename H, typename E, typename A,
+          typename Pred>
+void erase_if(F14NodeMap<K, M, H, E, A>& c, Pred pred) {
+  f14::detail::erase_if_impl(c, pred);
+}
+
+template <typename K, typename M, typename H, typename E, typename A,
+          typename Pred>
+void erase_if(F14VectorMap<K, M, H, E, A>& c, Pred pred) {
+  f14::detail::erase_if_impl(c, pred);
+}
+
+template <typename K, typename M, typename H, typename E, typename A,
+          typename Pred>
+void erase_if(F14FastMap<K, M, H, E, A>& c, Pred pred) {
+  f14::detail::erase_if_impl(c, pred);
+}
+
 } // namespace folly
--- a/folly/container/detail/F14Table.h
+++ b/folly/container/detail/F14Table.h
@@ -187,6 +187,16 @@ struct StdNodeReplica<
 
 #endif
 
+template <class Container, class Predicate>
+void erase_if_impl(Container& c, Predicate& predicate) {
+  for (auto i = c.begin(), last = c.end(); i != last;) {
+    auto prev = i++;
+    if (predicate(*prev)) {
+      c.erase(prev);
+    }
+  }
+}
+
 } // namespace detail
 } // namespace f14
 
--- a/folly/container/test/F14MapTest.cpp
+++ b/folly/container/test/F14MapTest.cpp
@@ -1777,6 +1777,26 @@ TEST(F14Map, containsWithPrecomputedHash) {
   testContainsWithPrecomputedHash<F14FastMap>();
 }
 
+template<template<class...> class TMap>
+void testEraseIf()
+{
+  TMap<int, int> m{{1, 1}, {2, 2}, {3, 3}, {4, 4}};
+  const auto isEvenKey = [](const auto& p) {
+    return p.first % 2 == 0;
+  };
+  erase_if(m, isEvenKey);
+  ASSERT_EQ(2u, m.size());
+  EXPECT_TRUE(m.contains(1));
+  EXPECT_TRUE(m.contains(3));
+}
+
+TEST(F14Map, eraseIf) {
+  testEraseIf<F14ValueMap>();
+  testEraseIf<F14VectorMap>();
+  testEraseIf<F14NodeMap>();
+  testEraseIf<F14FastMap>();
+}
+
 ///////////////////////////////////
 #endif // FOLLY_F14_VECTOR_INTRINSICS_AVAILABLE
 ///////////////////////////////////
--- a/folly/container/F14Set.h
+++ b/folly/container/F14Set.h
@@ -1225,4 +1225,24 @@ void swap(F14FastSet<K, H, E, A>& lhs, F14FastSet<K, H, E, A>& rhs) noexcept(
     noexcept(lhs.swap(rhs))) {
   lhs.swap(rhs);
 }
+
+template <typename K, typename H, typename E, typename A, typename Pred>
+void erase_if(F14ValueSet<K, H, E, A>& c, Pred pred) {
+  f14::detail::erase_if_impl(c, pred);
+}
+
+template <typename K, typename H, typename E, typename A, typename Pred>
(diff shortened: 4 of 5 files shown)


folly/container/detail/F14Table.h
```diff
@@ -2586,6 +2586,19 @@ class F14Table : public Policy {
 
 #endif // FOLLY_F14_VECTOR_INTRINSICS_AVAILABLE
 
+namespace {
+    const auto erase_if_impl = [](auto& c, auto predicate) {
+      for (auto i = c.begin(), last = c.end(); i != last;) {
+        if (predicate(*i)) {
```
nbronson: The generated code will be more compact if it is written

    auto prev = i++;
    if (predicate(*prev)) {
      c.erase(prev);
    }

JoeLoser: Nice! I played around with this myself using GCC and Clang trunk on Godbolt and saw the same results. Thanks for the tip 👍 

I've rewritten the loop to use your suggestion.

JoeLoser:
@yfeldblum what do you think of adding `erase_if` to the various containers in Folly? The idea is from [N4009](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4009.htm) which got merged into the C++20 working draft already.

yfeldblum:
I like the idea of adding `erase_if` for all containers in line with what we know will ship with C++20.

JoeLoser:
Great. If you think it's ready, feel free to import this then and I can work on adding `erase_if` to other containers in the short-term future.

folly/container/detail/F14Table.h
```diff
@@ -2586,6 +2586,17 @@ class F14Table : public Policy {
 
 #endif // FOLLY_F14_VECTOR_INTRINSICS_AVAILABLE
 
+namespace {
+    const auto erase_if_impl = [](auto& c, auto predicate) {
```
yfeldblum: This seems like this is not properly namespaced within `f14::detail`.

JoeLoser: Just moved `erase_if_impl` into `f14::detail` namespace to match local consistency.

folly/container/detail/F14Table.h
```diff
@@ -187,6 +187,15 @@ struct StdNodeReplica<
 
 #endif
 
+const auto erase_if_impl = [](auto& c, auto predicate) {
```
yfeldblum: Is there a reason to make this a callable global object v.s. a function template?

JoeLoser: No particular reason at all. In general, I reach for common lambdas over function templates as it's less verbose syntax due to `auto` parameters rather than them being deduced from templates.

Should I prefer function templates over lambdas for common helper functions in this codebase/here in general? I don't have strong opinions, so I'm happy to adopt whichever is preferred.

yfeldblum: If there is a specific reason that a lambda here is better than a function template then go with the lambda, but otherwise function templates are the norm in folly.

JoeLoser: Just switched to a function template for `erase_if_impl` to stay consistent with other parts of the code. The lambda doesn't buy us much here.

folly/container/detail/F14Table.h
```diff
@@ -187,6 +187,16 @@ struct StdNodeReplica<
...
 
+template <class Container, class Predicate>
+void erase_if_impl(Container& c, Predicate predicate) {
+  for (auto i = c.begin(), last = c.end(); i != last;) {
+    auto prev = i++;
+    if (predicate(*prev)) {
+      c.erase(prev);
+    }
+  }
+};
```
yfeldblum: We effectively build internally with `-Werror,-Wc++98-compat-extra-semi`, so this extra `;` breaks the build.

folly/container/detail/F14Table.h
```diff
@@ -187,6 +187,16 @@ struct StdNodeReplica<
 
 #endif
 
+template <class Container, class Predicate>
+void erase_if_impl(Container& c, Predicate predicate) {
```
yfeldblum: While it follows the std for `erase_if` to take `Predicate` by value, the trouble comes when `erase_if` calls `erase_if_impl` and attempts to copy `Predicate` - the type may not be copy-constructible. We can take `Predicate&` here.

folly/container/detail/F14Table.h
JoeLoser: Fixed. One of my vim plugins added it and I didn't notice; sorry.

Any opposition to a separate MR turning on the extra semicolon warning to the compiler flags for the CMake build?

folly/container/detail/F14Table.h
JoeLoser: Right, I followed the std in API here, but I'm fine taking `Predicate&`.

The predicate could now be a `folly::Function` which isn't copyable, which may be a real use case. In practice, I would expect most predicates here to be cheaply copyable lambdas.

folly/container/detail/F14Table.h
yfeldblum: Thanks.

We build with `-Wextra-semi` in many clang builds. Haven't checked if gcc has anything similar.

folly/container/F14Map.h
```diff
@@ -1490,4 +1490,28 @@ void swap(
   lhs.swap(rhs);
 }
 
+template <typename K, typename M, typename H, typename E, typename A,
+          typename Pred>
```
yfeldblum: JFYI internal formatting will translate this to:

```
template <
    typename K,
    typename M,
    typename H,
    typename E,
    typename A,
    typename Pred>
```

folly/container/test/F14MapTest.cpp
```diff
@@ -1777,6 +1777,26 @@ TEST(F14Map, containsWithPrecomputedHash) {
   testContainsWithPrecomputedHash<F14FastMap>();
 }
 
+template<template<class...> class TMap>
+void testEraseIf()
+{
```
yfeldblum: And this to:

```
template <template <class...> class TMap>
void testEraseIf() {
```

JoeLoser:
Just updated the commit message to reflect the difference in the std's API by taking the `Predicate` by reference rather than by value.

folly/container/F14Map.h
JoeLoser: That's fine. Want me to do it or let clang-format internally do it?

It would be nice to have the `.clang-format` in the open source repo as I asked [last year](https://github.com/facebook/folly/issues/755).

folly/container/test/F14MapTest.cpp
JoeLoser: No problem. Sorry for the old style habits; I just let clang-format do its thing on my code to not worry about these things too much.

yfeldblum:
The new commit message doesn't seem right to me. Having `erase_if` take `Predicate` by value is what the code does and also follows the std.

As an internal implementation detail, `erase_if_impl` needs to take `Predicate&` by reference in case `Predicate` is non-copyable. But this does not affect the interface of `erase_if`.

JoeLoser:
Valid point, sorry I missed that. You're certainly right -- changed the commit message back to its original form.

folly/container/F14Map.h
yfeldblum: I agree that it would be nice! Given that we already do this as of recently with the fbcode_builder scripts, it might be easier to auto-sync `.clang-format` into all the repos now.

Anyway I am just pointing this out up front. No matter what you do, since our `.clang-format` is currently not exported, there would be some little whitespace nit somewhere that internal formatting would change.

JoeLoser: 👍 Thanks for the heads up. No biggie on the whitespace diffs.

I'd love to see the auto syncing with `.clang-format` 😄

PR #1070, opened 2019-03-19
Mark swap(Replaceable) conditionally noexcept

JoeLoser:
Summary:
- Mark `swap(Replaceable&)` noexcept iff `Replaceable` is
  noexcept-swappable. As the comment aludes to, this is a C++17 feature,
  but we have it backported in `folly/Traits.h`

### DIFF
--- a/folly/Replaceable.h
+++ b/folly/Replaceable.h
@@ -595,11 +595,8 @@ class alignas(T) Replaceable
 
   /**
    * `swap` just calls `swap(T&, T&)`.
-   *
-   * Should be `noexcept(std::is_nothrow_swappable<T>::value)` but we don't
-   * depend on C++17 features.
    */
-  void swap(Replaceable& other) {
+  void swap(Replaceable& other) noexcept(IsNothrowSwappable<Replaceable>{}) {
     using std::swap;
     swap(*(*this), *other);
   }
--- a/folly/test/ReplaceableTest.cpp
+++ b/folly/test/ReplaceableTest.cpp
@@ -44,7 +44,9 @@ struct HasRef final {
     ++i1;
   }
 };
-
+void swap(HasRef& lhs, HasRef& rhs) {
+  std::swap(lhs.i1, rhs.i1);
+}
 struct OddA;
 struct OddB {
   OddB() = delete;
@@ -69,6 +71,11 @@ struct OddA {
 };
 struct Indestructible {
   ~Indestructible() = delete;
+};
+
+struct HasInt {
+  HasInt(int v) : value{v} {}
+  int value{};
 };
 } // namespace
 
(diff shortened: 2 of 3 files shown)


folly/Replaceable.h
```diff
@@ -595,11 +595,8 @@ class alignas(T) Replaceable
 
   /**
    * `swap` just calls `swap(T&, T&)`.
-   *
-   * Should be `noexcept(std::is_nothrow_swappable<T>::value)` but we don't
-   * depend on C++17 features.
    */
-  void swap(Replaceable& other) {
+  void swap(Replaceable& other) noexcept(IsNothrowSwappable<Replaceable>{})
```
yfeldblum: Missing `{`.

aary:
Looks good to me.  Can we add a simple unit test for this as well?  So if there are issues with our fallback implementation, tests pick it up?

JoeLoser:
@aary sure, I can write a test tonight. I noticed this morning that the ReplaceableTest isn't even a built executable (i.e. it's not specified in the top-level `CMakeLists.txt`). I assume this is an oversight and we should want these tests running in the open source build. Is that a valid assumption?

yfeldblum:
@JoeLoser Yes - that is certainly an oversight! Internally we use Buck when iterating, and we tend to be forgetful about CMake when adding new unit-test sources.

JoeLoser:
@yfeldblum No worries, I figured as much. 👍 

@aary I wrote a simple unit test, please take a look.

JoeLoser:
Adding `ReplaceableTest` to the open source build exposed the lack-of-use of the feature test macro for deduction guides: https://travis-ci.org/facebook/folly/builds/509234406#L3675

I fixed this in https://github.com/facebook/folly/pull/1065, so this PR will have to wait until https://github.com/facebook/folly/pull/1065 lands. 

This use of `__cplusplus` directive in that PR is wrong as seen by [Wandbox example](https://wandbox.org/permlink/EqAXoP4k2ozQsDE9). I will update the commit on that PR accordingly.

yfeldblum:
Okay!

JoeLoser:
Rebased now that #1065 has landed. This should be ready now.

folly/test/ReplaceableTest.cpp
```diff
@@ -70,6 +70,14 @@ struct OddA {
...
   ~Indestructible() = delete;
 };
+
+struct HasInt {
+  HasInt(int v) : value{v} {}
+  int value{};
+};
+void swap(HasInt& lhs, HasInt& rhs) noexcept {
+  std::swap(lhs.value, rhs.value);
+}
```
yfeldblum: Reviewer suggests

> can we add 1) a namespace here as well just so we explicitly test the ADL case and 2) a swap that is not noexcept to test the failure case?

JoeLoser: Sure, I can write those tests tonight. I agree they would be good to have.

JoeLoser: Wrote both of those tests and updated the commit message. We now test both positive/negative cases where `Replaceable<T>` is noexcept in addition to a separate ADL-swap test to verify the user-provided `swap` function is called.

yfeldblum: Nice!

JoeLoser:

Issue #1061, opened 2019-03-16
Fibers test fails to link as of db3f999


JoeLoser:
In https://github.com/facebook/folly/commit/db3f999e816ed1a5994621c0dd4fc5b795ce228a, `folly/fibers/test/FibersTest.cpp` now requires linking with `folly/fibers/Semaphore.cpp` in order to get the definition for `future_wait`. This breaks the open source build, as seen by Travis: https://travis-ci.org/facebook/folly/builds/507052180#L4803

JoeLoser:
Fixed in https://github.com/facebook/folly/commit/b7e9f46e217ef05fa85bf3df51e9d943959e3ff8

Thanks @yfeldblum!

JoeLoser:
Opening this issue back up as, while `future_wait` is found now by the linker, `wait(std::unique_ptr<Baton>)` is not found:

```
Undefined symbols for architecture x86_64:
  "folly::futures::wait(std::__1::unique_ptr<folly::fibers::Baton, std::__1::default_delete<folly::fibers::Baton> >)", referenced from:
      folly::fibers::Semaphore::future_wait() in libfolly.a(Semaphore.cpp.o)
```

Any ideas? Adding in what look to be missing includes:

```
+#include <folly/futures/Future.h>
+#include <folly/fibers/Baton.h>
+#include <memory
```

does not solve the linker issue. This happens both with GNU Make and Ninja, so it doesn't seem to be a dependencies issue in the build system forming its dag. I am a bit surprised to see Travis CI pass.

JoeLoser:
Found the bug. Proposed PR here -> https://github.com/facebook/folly/pull/1072

Summary:

- `future_wait` will cause link errors as `wait(std::unique_ptr<fibers::Baton>)` will not exist for mobile or Mac OS X users. This is due to `FOLLY_FUTURE_USING_FIBER` will be set to `0`.
- Travis CI passes as `FOLLY_FUTURE_USING_FIBER` will be set to 1 and hence the symbol will be available to avoid linker errors.

PR #1072, opened 2019-03-21
Only define future_wait if FOLLY_FUTURE_USING_FIBER is set

JoeLoser:
Summary:
- In `db3f999e816ed1a5994621c0dd4fc5b795ce228a`, `future_wait` was added
  to `folly/fibers/Semaphore.cpp`. In this function, it requires being
  able to call `wait(std::unique_ptr<fibers::Baton>)`. This symbol will
  only be defined if and only if `FOLLY_FUTURE_USING_FIBER` is true. For
  example, this will break Mac OS X and mobile users, i.e. where
  `FOLLY_FUTURE_USING_FIBER` is not set.
- To avoid breaking those users, we only define `future_wait` if
  `FOLLY_FUTURE_USING_FIBER` is set.
- Change the Semaphore test in `folly/fibers/test/FibersTest.cpp` to
  only unconditionally call `wait()` in the loop rather than randomly
  choosing to call `wait` or `future_wait()` for the case where
  `FOLLY_FUTURE_USING_FIBER` is `0`. The code is the same for the case
  where `FOLLY_FUTURE_USING_FIBER` is set.

Closes https://github.com/facebook/folly/issues/1061.

### DIFF
--- a/folly/fibers/Semaphore.cpp
+++ b/folly/fibers/Semaphore.cpp
@@ -118,6 +118,7 @@ coro::Task<void> Semaphore::co_wait() {
 
 #endif
 
+#if FOLLY_FUTURE_USING_FIBER
 SemiFuture<Unit> Semaphore::future_wait() {
   auto oldVal = tokens_.load(std::memory_order_acquire);
   do {
@@ -137,6 +138,7 @@ SemiFuture<Unit> Semaphore::future_wait() {
       std::memory_order_acquire));
   return makeSemiFuture();
 }
+#endif
 
 size_t Semaphore::getCapacity() const {
   return capacity_;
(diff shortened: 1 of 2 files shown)


yfeldblum:
Thanks for looking at this. I actually have a more complete version of this fix in flight internally.

JoeLoser:
Of course. Anything that breaks my local dev usually jumps higher in the priority list ;)

Feel free to close this MR when your fix lands. Thanks for taking a look at it. 👍

yfeldblum:
f8ae709202cb0481b570d06415411a49e1b147d6 should do it.

JoeLoser:
Yep, that works. Thanks for the better fix @yfeldblum 🎉

PR #1073, opened 2019-03-21
TravisCI build for Folly on MacOS

calebmarchent:
Use homebrew bootstrap script for process; note nearly all the Travis
steps become conditional on os; so bit of shuffling was required to make
that work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,11 @@ matrix:
     - env: ['os_image=ubuntu:16.04', gcc_version=5]
       services: [docker]
 
+addons:
+  apt:
+    packages: python2.7
+
 script:
-  # Travis seems to get confused when `matrix:` is used with `language:`
-  - sudo apt-get install python2.7
   # We don't want to write the script inline because of Travis kludginess --
   # it looks like it escapes " and \ in scripts when using `matrix:`.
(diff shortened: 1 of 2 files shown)


JoeLoser:
Thanks for working on this @calebmarchent.

JoeLoser:
@calebmarchent one thing to keep in mind is that the existing bootstrap script for Mac *does not* run the tests. We should probably pass `-DBUILD_TESTS=ON` when invoking this script for CI purposes.

JoeLoser:
Any update on this, @calebmarchent? How do others internally feel about this work?

calebmarchent:
It did get held at internal review, then I got onto something else, now I need to finish some stuff this week; but next week should be clearer so will tackle at start of next week.

calebmarchent:
Rebased; because it has been a while; will revise on next step.

calebmarchent:
Update: Have implemented the first code review feedback; splitting out enable of IPv6 in docker.
But found the travis config is shared with a number of oss projects; which means I will need to split out the config before enabling OSX build; which I will do for folly only.

Orvid:
We've dropped support for Travis in favor of Github Actions, which already has MacOS support.

PR #1074, opened 2019-03-22
Use folly::MoveOnly in PolyTest.cpp

JoeLoser:
Summary:
- Use the `folly::MoveOnly` type in `PolyTest.cpp` instead of defining
  a separate type.


yfeldblum:
When building with Clang,

```
folly/test/PolyTest.cpp:608:21: error: no viable conversion from 'folly::MoveOnly' (aka 'folly::moveonly_::MoveOnly') to 'Poly<folly::poly::IMoveOnly>'
    Poly<IMoveOnly> p = MoveOnly{};
                    ^   ~~~~~~~~~~
```

JoeLoser:
Sigh, burned again by the fact that [this test isn't built](https://github.com/facebook/folly/blob/master/CMakeLists.txt#L822) by the open source build which is why I didn't catch the compiler error. I'll look into it. Thanks.

I'll create a PR in the near future to enable many missing tests from `folly/test` to be built/run by the open-source build to avoid these sort of things in the future.

yfeldblum:
When building with Clang,

```
folly/test/PolyTest.cpp:609:35: error: temporary of type 'folly::MoveOnly' (aka 'folly::moveonly_::MoveOnly') has protected destructor
    Poly<IMoveOnly> p = MoveOnly_{};
```

JoeLoser:
I couldn't recreate that compiler issue locally with Clang 7.0.1 or Clang 8.0.0. This [LLVM Code Review](https://reviews.llvm.org/D53860) may be the culprit.

I'm not as pleased with the change's benefit as the original code now, but it's a tiny better over the existing code. 

It did expose some linker errors on my end with this test which I have not yet chased down. Given that we'll have a Mac OS X CI soon, I added a new tag for tests which are disabled on Mac OS X, similar to the `WINDOWS_DISABLED`. I marked this test with the new tag.

folly/test/PolyTest.cpp
```diff
@@ -614,6 +605,9 @@ TEST(Poly, Move) {
     EXPECT_EQ(&poly_cast<int>(p), &poly_cast<int>(q));
   }
   {
+    struct MoveOnly_ : MoveOnly {
+      virtual ~MoveOnly_() = default;
```
yfeldblum: Does this need to be `virtual`?

JoeLoser: No, sorry. Fixed.

PR #1075, opened 2019-03-23
Run several more tests in open source build

JoeLoser:
Summary:
- There are a few tests which are disabled for all platforms, but only
  need to be disabled on Windows. Uncomment them so they are built and
  run as part of the open source build. Note that `EventHandlerTest.cpp`
  could be built as part of the CI, but we omit that for now. In the
  short-term future, there will be a folly Mac OS CI build and this test
  will not build for Mac OS X.
- Some tests from `folly/test` are not listed in the top-level
  `CMakeLists.txt` and hence are not getting built and run. Add these.
  Modify a comment in `FBStringTest.cpp` as the gtest auto discovery of
  tests was treating the comment as a new test case. In turn, this
  results in an error at CMake configure time as you cannot have two
  test cases with the same name as part of the same test suite.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -637,7 +637,7 @@ if (BUILD_TESTS)
 
     DIRECTORY gen/test/
       # MSVC bug can't resolve initializer_list constructor properly
-      #TEST base_test SOURCES BaseTest.cpp
+      TEST base_test WINDOWS_DISABLED SOURCES BaseTest.cpp
       TEST combine_test SOURCES CombineTest.cpp
       TEST parallel_map_test SOURCES ParallelMapTest.cpp
       TEST parallel_test SOURCES ParallelTest.cpp
@@ -677,11 +677,10 @@ if (BUILD_TESTS)
           AsyncSSLSocketTest2.cpp
           AsyncSSLSocketWriteTest.cpp
           AsyncTransportTest.cpp
-          # This is disabled because it depends on things that don't exist
-          # on Windows.
-          #EventHandlerTest.cpp
-          # The async signal handler is not supported on Windows.
-          #AsyncSignalHandlerTest.cpp
+      # This is disabled because it depends on things that don't exist
+      # on Windows.
+      # TODO: Refactor EventHandlerTest to not use eventfd so it can work on Mac OS X.
+      #TEST async_event_handler_test WINDOWS_DISABLED SOURCES EventHandlerTest.cpp
       TEST async_timeout_test SOURCES AsyncTimeoutTest.cpp
       TEST AsyncUDPSocketTest SOURCES AsyncUDPSocketTest.cpp
       TEST DelayedDestructionTest SOURCES DelayedDestructionTest.cpp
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -747,15 +746,20 @@ if (BUILD_TESTS)
       TEST ahm_int_stress_test SOURCES AHMIntStressTest.cpp
       TEST arena_smartptr_test SOURCES ArenaSmartPtrTest.cpp
       TEST ascii_check_test SOURCES AsciiCaseInsensitiveTest.cpp
+      TEST atfork_test SOURCES AtForkTest.cpp
```
yfeldblum: Windows does not have `fork`.

JoeLoser: Fixed. A caveat with this PR: I didn't spin up a Windows docker image to see which tests just don't work on Windows.

CMakeLists.txt
```diff
@@ -764,10 +768,14 @@ if (BUILD_TESTS)
...
       TEST dynamic_converter_test SOURCES DynamicConverterTest.cpp
       TEST dynamic_other_test SOURCES DynamicOtherTest.cpp
+      TEST enable_shared_from_this SOURCES EnableSharedFromThisTest.cpp
       TEST endian_test SOURCES EndianTest.cpp
+      TEST exception_string_test BROKEN SOURCES ExceptionStringTest.cpp
       TEST exception_test SOURCES ExceptionTest.cpp
       TEST exception_wrapper_test SOURCES ExceptionWrapperTest.cpp
+      TEST expected_coroutines_test SOURCES ExpectedCoroutinesTest.cpp
       TEST expected_test SOURCES ExpectedTest.cpp
+      TEST fbstring_test SOURCES FBStringTest.cpp
```
yfeldblum: Note: `fbstring` had a problem building with msvc, fixed (and also enabling in cmake) in 8c458a4d6d2b09bd95556505f68278eb93242c11. So this line will be rebased out.

JoeLoser: Yep, I saw 8c458a4 go in today -- nice fix 👍 

Rebased with top of tree.

JoeLoser: @yfeldblum when your `fbstring` fix landed earlier, it accidentally broke CMake 3.9+ users as discussed in this PR. I've made a separate PR with that test comment fix -> https://github.com/facebook/folly/pull/1080

JoeLoser:
Rebased and updated commit message since https://github.com/facebook/folly/pull/1080 landed.

CMakeLists.txt
```diff
@@ -856,8 +870,12 @@ if (BUILD_TESTS)
       TEST token_bucket_test SOURCES TokenBucketTest.cpp
       TEST traits_test SOURCES TraitsTest.cpp
       TEST try_test SOURCES TryTest.cpp
+      TEST type_list_test SOURCES TypeListTest.cpp
```
Orvid: This needs to be `WINDOWS_DISABLED`, as it doesn't currently work under MSVC due to compiler issues.

JoeLoser: Just marked it `WINDOWS_DISABLED`. Are there other tests that I've enabled in the CMake build which break Windows? I don't readily have access to a machine where I can build folly with MSVC 2017 to test.

JoeLoser:
@Orvid Do any other tests need to be marked `WINDOWS_DISABLED`?

PR #1077, opened 2019-03-23
Do not define emplace_back_aux out-of-line in FBVector

JoeLoser:
Summary:
- Older versions of GCC required defining private member function
  `emplace_back_aux` out-of-line. That is, declare the member in-class and
  then provide the definition fully out of line.
- These older versions of GCC are not supported anymore, so just define
  it inline as you would any other member function.

### DIFF
--- a/folly/FBVector.h
+++ b/folly/FBVector.h
@@ -1214,7 +1214,71 @@ class fbvector {
   }
 
   template <class... Args>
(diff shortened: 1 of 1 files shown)


JoeLoser:

yfeldblum:
Sorry, been backlogged and hadn't added any reviewers to this.

PR #1090, opened 2019-03-28
Deduction guide for fbvector

JoeLoser:
Summary:
- Add deduction guide for `fbvector` which is only available if the
  feature test macro for deduction guides is available or if we are on
  MSVC.

### DIFF
--- a/folly/FBVector.h
+++ b/folly/FBVector.h
@@ -1748,4 +1748,11 @@ void attach(fbvector<T, A>& v, T* data, size_t sz, size_t cap) {
   v.impl_.z_ = data + cap;
 }
 
+#if __cpp_deduction_guides >= 201703
+template<class InputIt,
+         class Allocator = std::allocator<typename std::iterator_traits
+         <InputIt>::value_type>>
+fbvector(InputIt, InputIt, Allocator = Allocator())
+  -> fbvector<typename std::iterator_traits<InputIt>::value_type, Allocator>;
+#endif
 } // namespace folly
--- a/folly/test/FBVectorTest.cpp
+++ b/folly/test/FBVectorTest.cpp
@@ -256,3 +256,15 @@ TEST(FBVector, zero_len) {
   fb6 = il;
   fbvector<int> fb7(fb6.begin(), fb6.end());
 }
+
+#if __cpp_deduction_guides >= 201703
+TEST(FBVector, deduction_guides) {
+  fbvector<int> v(3);
+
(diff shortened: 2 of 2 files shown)


folly/test/FBVectorTest.cpp
```diff
@@ -256,3 +256,15 @@ TEST(FBVector, zero_len) {
...
+
+#if __cpp_deduction_guides >= 201703 || _MSC_VER
+TEST(FBVector, deduction_guides) {
+  fbvector<int> v(3);
+
+  fbvector x(v.begin(), v.end());
+  static_assert(std::is_same_v<fbvector<int>, decltype(x)>);
+
+  fbvector y{v.begin(), v.end()};
+  static_assert(std::is_same_v<fbvector<fbvector<int>::iterator>, decltype(y)>);
```
yfeldblum: lol

JoeLoser: Bit extreme of a test, or laughing because of how the first phase of overload resolution for list-initialization will select the initializer-list constructor?

yfeldblum: Pay no mind - this is just fine.

folly/FBVector.h
```diff
@@ -1748,4 +1748,11 @@ void attach(fbvector<T, A>& v, T* data, size_t sz, size_t cap) {
   v.impl_.z_ = data + cap;
 }
 
+#if __cpp_deduction_guides >= 201703 || _MSC_VER
```
yfeldblum: You may hate me now, but based on the discussion in the internal diff review tool, there appears to be precisely one point release of MSVC which supports deduction guides but which also leaves the name `__cpp_deduction_guides` undefined. So we can remove `_MSC_VER` from the guard.

Would you mind also submitting a PR to `Replaceable` removing `_MSC_VER` from the guard?

JoeLoser: No worries, I don't hate you 😆 Thanks for the heads up. I lean on you guys for the MSVC nuances as I just use Clang for local dev.

Just removed `_MSC_VER` from the macro guard. [Here](https://github.com/facebook/folly/pull/1097) is a PR for removing `_MSC_VER` from the guard as well for `Replaceable`'s deduction guide.

JoeLoser:
@yfeldblum just switched these tests to use `EXPECT_EQ` per discussion from https://github.com/facebook/folly/pull/1097.

PR #1094, opened 2019-03-29
Cut FBSTRING_SANITIZE_ADDRESS in FBString.h 

JoeLoser:
Summary:
- Defining `FBSTRING_SANITIZE_ADDRESS` is more complicated than it needs
  to be and it is in fact not even needed now.
- Remove separate `FBSTRING_SANITIZE_ADDRESS` symbol and use
  `FOLLY_SANITIZE_ADDRESS` in places which previously used
  `FBSTRING_SANITIZE_ADDRESS`.

### DIFF
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -42,6 +42,7 @@
 #else // !_LIBSTDCXX_FBSTRING
 
 #include <folly/CppAttributes.h>
+#include <folly/CPortability.h>
 #include <folly/Portability.h>
 
 // libc++ doesn't provide this header, nor does msvc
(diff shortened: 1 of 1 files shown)


folly/FBString.h
```diff
@@ -98,13 +99,7 @@ FOLLY_GNU_DISABLE_WARNING("-Wshadow")
...
 FOLLY_FBSTRING_BEGIN_NAMESPACE
 
-#if defined(__clang__)
-#if __has_feature(address_sanitizer)
-#define FBSTRING_SANITIZE_ADDRESS
-#endif
-#elif defined(__GNUC__) &&                                             \
-    (((__GNUC__ == 4) && (__GNUC_MINOR__ >= 8)) || (__GNUC__ >= 5)) && \
-    __SANITIZE_ADDRESS__
+#if defined(FOLLY_SANITIZE_ADDRESS) && defined(__GNUC__)
```
yfeldblum: Don't need the `__GNUC__` test anymore, then.

Which implies that we don't need `FBSTRING_SANITIZE_ADDRESS` as a separate symbol anymore and uses may be replaced with `FOLLY_SANITIZE_ADDRESS`.

JoeLoser: Good point; even better! Just made that change. 👍

PR #1095, opened 2019-03-29
Simplify some macros testing __GNUC__ and __clang__

JoeLoser:
Summary:
- Both GCC and Clang define the macro `__GNUC__`, `__GNUG__` and
  friends. Simplify some macros whose intent is to check "if we are GCC or
  Clang" but do so via checking `__GNUC__` in addition to `__clang__`. It
  is sufficient to just check `__GNUC__`.

### DIFF
--- a/folly/CPortability.h
+++ b/folly/CPortability.h
@@ -147,7 +147,7 @@
 // noinline
 #ifdef _MSC_VER
 #define FOLLY_NOINLINE __declspec(noinline)
-#elif defined(__clang__) || defined(__GNUC__)
+#elif defined(__GNUC__)
 #define FOLLY_NOINLINE __attribute__((__noinline__))
 #else
 #define FOLLY_NOINLINE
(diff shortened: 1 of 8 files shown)


yfeldblum:
@Orvid How does this affect clang as a front-end to msvc?

Orvid:
clang-c2 is dead, so that side doesn't matter. I'm not sure what full clang on windows does with regards to these matters though, and there are some eventual plans internally to test that out, so I'd rather not add more work than needed to that effort :)

Ignoring the technical aspects, I kinda like having both clang and gnuc being checked at each location, as it makes it clearer to the average reader (who likely doesn't know that both are defined by clang) that clang will also use that set of code. Whether it's worth keeping them for that, especially when nothing enforces that they're used consistently, I'm not sure :)

JoeLoser:
Thanks for the input @Orvid. I agree that unless you've done some development in supporting multiple compilers, one may not be aware of the fact that both gcc and clang define `__GNUC__`.

Do you think it's worth seeing what others internally think about this before we close the MR? I don't feel strongly one way or another regarding the code; it was just a drive-by cleanup as I was reading code. In general, I favor more concise code, but if it harms the general readability by others, we may not want this.

Do you think most developers working inside folly are aware of this macro knowledge?

yfeldblum:
We definitely do not do both tests consistently.

JoeLoser:
Just rebased and fixed another use in `folly/experimental/pushmi/detail/concept_def.h` which landed on Friday.

PR #1097, opened 2019-03-30
Fix deduction guide macro guard for Replaceable

JoeLoser:
Summary:
- `Replaceable` defines a deduction guide if the feature test macro for
  it is met, or if we are on MSVC. However, the MSVC version that folly
  supports also defines the feature test macro for deduction guides. As
  such, remove the extra check for MSVC.
- Add `ReplaceableTest` to the  cmake build.
- Add test for deduction guide.

### DIFF
--- a/folly/test/ReplaceableTest.cpp
+++ b/folly/test/ReplaceableTest.cpp
@@ -289,3 +289,11 @@ TEST(ReplaceableTest, Conversions) {
   Replaceable<OddA> rOddA{std::move(rOddB)};
   Replaceable<OddB> rOddB2{rOddA};
 }
+
+#if __cpp_deduction_guides >= 201703
+TEST(ReplaceableTest, DeductionGuide) {
+  Basic b{};
+  Replaceable r{b};
+  EXPECT_TRUE((std::is_same_v<Replaceable<Basic>, decltype(r)>));
+}
+#endif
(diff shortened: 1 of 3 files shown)


folly/test/ReplaceableTest.cpp
```diff
@@ -289,3 +289,11 @@ TEST(ReplaceableTest, Conversions) {
   Replaceable<OddA> rOddA{std::move(rOddB)};
   Replaceable<OddB> rOddB2{rOddA};
 }
+
+#if __cpp_deduction_guides >= 201703
+TEST(ReplaceableTest, DeductionGuide) {
+  Basic b{};
+  Replaceable r{b};
+  static_assert(std::is_same_v<Replaceable<Basic>, decltype(r)>);
```
yfeldblum: There are other checks of compile-time properties, but they are done as runtime expectations with `EXPECT_EQ` etc.

JoeLoser: Is there a particular reason to use `EXPECT_EQ` over the `static_assert`? Happy to change, just looking to understand the reasoning.

yfeldblum: Builds by default tend to terminate at the first build failure, but tests by default show all expectation failures.

JoeLoser: Just switched to `EXPECT_TRUE`. Want me to do the same for `fbvector` tests over in https://github.com/facebook/folly/pull/1090?

yfeldblum: Please.

PR #1104, opened 2019-04-09
Add onError interface to Expected type

xuning97:
So that we can do some thing in the middle or chaining, for example
log the error.

While using Expected, we found that It's nice to have the chaining operations, but one downside is  there is no place to log each possible error here and there.

So I added this onError interface, expecting the usage like this

Expected<D, Error> maybeD = someFunc()
      .then([](A a){return B(a);})
      .onError([](Error e) { LOG_DEBUG("error happened while executing function A->B");})
      .then([](B b){return C(b);})
      .then([](C c){return D(c);});


xuning97:
@ericniebler may you kindly have a look?

xuning97:
@yfeldblum is this PR useful? If the purpose is useful and the implementation is not that good, I can continue update it

PR #1107, opened 2019-04-10
Fix warning 'redundant-move'

stac47:
Hello Facebook,

This small PR to fix the redundant std::move usage detected by gcc (GCC) 9.0.1 20190225 (experimental)

```
/folly/BUILD/folly-2019.04.01.00/folly/executors/task_queue/PriorityLifoSemMPMCQueue.h:92:30: error: redundant move in return statement [-Werror=redundant-move]
   92 |         return std::move(item);
      |                              ^
```

Stac

### DIFF
--- a/folly/executors/task_queue/LifoSemMPMCQueue.h
+++ b/folly/executors/task_queue/LifoSemMPMCQueue.h
@@ -57,7 +57,7 @@ class LifoSemMPMCQueue : public BlockingQueue<T> {
         return folly::none;
       }
     }
-    return std::move(item);
+    return item;
   }
 
   size_t capacity() {
--- a/folly/executors/task_queue/PriorityLifoSemMPMCQueue.h
+++ b/folly/executors/task_queue/PriorityLifoSemMPMCQueue.h
@@ -89,7 +89,7 @@ class PriorityLifoSemMPMCQueue : public BlockingQueue<T> {
     T item;
     while (true) {
       if (nonBlockingTake(item)) {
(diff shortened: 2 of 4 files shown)


folly/executors/task_queue/LifoSemMPMCQueue.h
```diff
@@ -57,7 +57,7 @@ class LifoSemMPMCQueue : public BlockingQueue<T> {
         return folly::none;
       }
     }
-    return std::move(item);
+    return item;
```
yfeldblum: The Clang warning `return-std-move` seems like it ought to fire here when omitting `std::move(...)` because the local variable type differs from the containing function's return type.

https://clang.llvm.org/docs/DiagnosticsReference.html#wreturn-std-move

stac47: I tested with the following clang version with:

- version 6.0.0-1ubuntu2
- version 7.1.0-svn353565-1~exp1~20190406090509.61
- version 8.0.0

And the described warning is not raised. Tell me if I missed something in the Travis build result.

But I agree that this can cause problems on old compilers because using move semantic on a returned local lvalue has been fixed in the standard in 2014 (See US 13 in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3903.html#US13)

I also tried the previously mentioned versions with the following options `-std=gnu++11`, `-std=gnu++14`, `-std=gnu++17`: I encountered no such warnings.

PR #1109, opened 2019-04-12
TravisCI build for Folly on MacOS

calebmarchent:
Use homebrew bootstrap script for process;

As Travis files are shared between many Facebook OSS projects; Folly .travis.yml needs to be forked to a unique file before this PR can be landed. This can only be done from within Facebook.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,15 +31,19 @@ matrix:
   include:
     - env: ['os_image=ubuntu:16.04', gcc_version=5]
       services: [docker]
+    - os: osx
 
 addons:
   apt:
     packages: python2.7
 
 script:
-  # We don't want to write the script inline because of Travis kludginess --
-  # it looks like it escapes " and \ in scripts when using `matrix:`.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,15 +31,19 @@ matrix:
...
   apt:
     packages: python2.7
 
 script:
-  # We don't want to write the script inline because of Travis kludginess --
-  # it looks like it escapes " and \ in scripts when using `matrix:`.
-  - ./build/fbcode_builder/travis_docker_build.sh
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]];
+    then
+      folly/build/bootstrap-osx-homebrew.sh;
```
JoeLoser: I think we should pass `-DBUILD_TESTS=ON` to this script so we build and run the tests. By default, the script does not build/run tests.

Orvid:
We've dropped support for Travis in favor of Github Actions, which already has MacOS support.

PR #1110, opened 2019-04-16
Update stop_watch.h

soyoo:
modify comment.

### DIFF
--- a/folly/stop_watch.h
+++ b/folly/stop_watch.h
@@ -181,7 +181,7 @@ struct custom_stop_watch {
    *
    *  do_something();
    *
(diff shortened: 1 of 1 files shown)


yfeldblum:
Thank you for the contribution but we won't be able to import it without a signed CLA. This policy is actually checked in our tooling for importing pull requests.

soyoo:
> Thank you for the contribution but we won't be able to import it without a signed CLA. This policy is actually checked in our tooling for importing pull requests.

Now, the CLA was signed.

yfeldblum:
Thanks!

PR #1113, opened 2019-04-19
Remove unused FOLLY_NONNULL macro

JoeLoser:
Summary:
- In `339c14d3`, `FOLLY_NONNULL` was added to make the internal linter
  not complain about passing around non-null pointers to functions
  instead of using a reference.
- It had one or two usages with `d81cde8f` and `a04d119c`, but it is no
  longer used.
- Since it is no longer used, remove the macro entirely.


JoeLoser:
Okay, thanks for the info. I'll close this then.

PR #1114, opened 2019-04-19
Return reference in FBVector::emplace_back

JoeLoser:
Summary:
- Since C++17, `vector::emplace_back()` returns a reference to
  the new element inserted.
- Differential Revision: D15007059 added support for these semantics in
  `small_vector`. This adds it for `FBVector` too to be symmetric.

### DIFF
--- a/folly/FBVector.h
+++ b/folly/FBVector.h
@@ -1136,13 +1136,14 @@ class fbvector {
   // modifiers (common)
  public:
   template <class... Args>
-  void emplace_back(Args&&... args) {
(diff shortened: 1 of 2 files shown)


JoeLoser:
CC: @ot

ot:
@JoeLoser Thanks! Can you add a test that verifies the address of the returned reference, instead of the value, to make sure we're correctly propagating a reference instead of copying? Similarly to what I did for `small_vector`.

JoeLoser:
@ot good point; just added that test.  👍

PR #1116, opened 2019-04-26
Add non-const data() method to FBString

JoeLoser:
Summary:
- As of C++17, `std::string` contains a non-const `data()` member function.
  This is useful when working with C APIs that have `char*` output parameters.
- Implements P0272 for `FBString`:
  http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0272r1.html

Example of workaround without this:

```
  fbstring s{};
  some_api(&s.front());
```

- The expression `&s.front()` above causes undefined behavior since the string
  is empty.

- Add non-const public member function `data()` to `fbstring` which
  delegates to `fbstring_core`.

### DIFF
--- a/folly/FBString.h
+++ b/folly/FBString.h
@@ -388,6 +388,10 @@ class fbstring_core {
     return c_str();
   }
 
+  Char* data() {
+    return c_str();
+  }
+
   Char* mutableData() {
     switch (category()) {
       case Category::isSmall:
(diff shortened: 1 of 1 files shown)


yfeldblum:
This should model `std::string` and not introduce non-`const` overload of `c_str`. This might be done by flipping what calls what.

JoeLoser:
Right. After some thought, the non-const `c_str()` is just outright dangerous, actually. We also do want 1-1 compatibility as a drop-in replacement for `std::string` without surprises. I'll look into this when I get some time.

yfeldblum:
@JoeLoser Just checking in here.

JoeLoser:
@yfeldblum sorry for the delay; was busy wrapping things up at work before CppNow and did not get a chance to look at this until now.

I've reworked it a bit. We now have a non-const public member function `data()` in `fbstring` which dispatches to public member function `data()` in `fbstring_core`. There, we hide the non-const `c_str()` as a private member function.

PR #1125, opened 2019-05-10
SFINAE-friendly to_unsigned

JoeLoser:
Summary:
- `to_unsigned` is not SFINAE friendly. It only works on types where
  `std::make_unsigned_t` is well-formed, such as integral types
  (non-bool) or enum type. This makes it less useful in generic
  programming contexts.
- Remove `to_unsigned` from `Histogram.h` in favor of the
  SFINAE-friendly `to_unsigned` in `Utility.h`.
- Add `UtilityTest` to be built and run as part of the CMake build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -860,6 +860,7 @@ if (BUILD_TESTS)
       TEST try_test SOURCES TryTest.cpp
       TEST unit_test SOURCES UnitTest.cpp
       TEST uri_test SOURCES UriTest.cpp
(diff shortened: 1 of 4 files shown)


yfeldblum:
It's intentionally not sfinae-friendly - this is just supposed to be a thin type-deducing wrapper around `make_unsigned_t`, and to have the same restrictions on when it is allowed.

The thing in Histogram has a different meaning, even though it shares a name.

PR #1127, opened 2019-05-10
Fix compilation without deprecated OpenSSL APIs

neheb:
First commit is for 1.0.2 and the second for 1.1.

### DIFF
--- a/folly/ssl/OpenSSLVersionFinder.h
+++ b/folly/ssl/OpenSSLVersionFinder.h
@@ -25,7 +25,7 @@ namespace folly {
 namespace ssl {
 inline std::string getOpenSSLLongVersion() {
 #ifdef OPENSSL_VERSION_TEXT
-  return SSLeay_version(SSLEAY_VERSION);
+  return OpenSSL_version(OPENSSL_VERSION);
 #elif defined(OPENSSL_VERSION_NUMBER)
   return folly::format("0x{:x}", OPENSSL_VERSION_NUMBER).str();
 #else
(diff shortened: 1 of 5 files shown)


folly/ssl/OpenSSLVersionFinder.h
```diff
@@ -18,14 +18,20 @@
 #include <folly/Conv.h>
 #include <folly/portability/OpenSSL.h>
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#define OPENSSL_VERSION SSLEAY_VERSION
```
yfeldblum: These symbols should be namespaced to folly.

neheb: Not sure what you mean (I don't really know C++).

yfeldblum: `#define FOLLY_DETAIL_OPENSSL_VERSION ...`.

And then `#undef` them after using them (preferred) or at the bottom of the file.

Orvid: Prefer to have this in the OpenSSL portability header so that everything is uniform.

neheb:
Force pushed. I think this is what was meant?

yfeldblum:
Can this be split into two PR, one PR for each of the commits?

yfeldblum:
Actually nah, importing as-is.

neheb:
In case compilation fails with OpenSSL 1.0.2, folly/ssl/OpenSSLCertUtils.cpp needs to include folly/portability/OpenSSL.h . Just noticed it.

PR #1128, opened 2019-05-15
Replace some NDEBUG with kIsDebug

JoeLoser:
Summary:
- Replace some calls sites from `NDEBUG` to `kIsDebug`.
- Not every call site is changed; there are a few useful cases of
  `NDEBUG` still. `xlog` macros are one of them. Another is where we
  conditionally define a variable in a local scope and then use it in an
  `assert`.

### DIFF
--- a/folly/FixedString.h
+++ b/folly/FixedString.h
@@ -79,6 +79,10 @@ constexpr std::size_t checkOverflowOrNpos(std::size_t i, std::size_t max) {
       : (i <= max ? i : (void(assertOutOfBounds()), max));
 }
 
+constexpr std::size_t checkOverflowIfDebug(std::size_t i, std::size_t size) {
+  return kIsDebug ? checkOverflow(i, size) : i;
+}
+
 // Intentionally NOT constexpr. See note above for assertOutOfBounds
 [[noreturn]] inline void assertNotNullTerminated() noexcept {
   assert(!"Non-null terminated string used to initialize a BasicFixedString");
@@ -1110,22 +1114,14 @@ class BasicFixedString : private detail::fixedstring::FixedStringBase {
    * \return `*(data() + i)`
    */
   constexpr Char& operator[](std::size_t i) noexcept {
-#ifdef NDEBUG
-    return data_[i];
-#else
-    return data_[detail::fixedstring::checkOverflow(i, size_)];
-#endif
+    return data_[detail::fixedstring::checkOverflowIfDebug(i, size_)];
   }
 
   /**
    * \overload
    */
   constexpr const Char& operator[](std::size_t i) const noexcept {
-#ifdef NDEBUG
-    return data_[i];
-#else
-    return data_[detail::fixedstring::checkOverflow(i, size_)];
-#endif
+    return data_[detail::fixedstring::checkOverflowIfDebug(i, size_)];
   }
 
   /**
@@ -1147,22 +1143,14 @@ class BasicFixedString : private detail::fixedstring::FixedStringBase {
    * \pre `!empty()`
    */
   constexpr Char& back() noexcept {
-#ifdef NDEBUG
-    return data_[size_ - 1u];
-#else
-    return data_[size_ - detail::fixedstring::checkOverflow(1u, size_)];
-#endif
+    return data_[size_ - detail::fixedstring::checkOverflowIfDebug(1u, size_)];
   }
 
   /**
    * \overload
    */
   constexpr const Char& back() const noexcept {
-#ifdef NDEBUG
-    return data_[size_ - 1u];
-#else
-    return data_[size_ - detail::fixedstring::checkOverflow(1u, size_)];
-#endif
+    return data_[size_ - detail::fixedstring::checkOverflowIfDebug(1u, size_)];
   }
 
   /**
--- a/folly/experimental/exception_tracer/StackTrace.cpp
+++ b/folly/experimental/exception_tracer/StackTrace.cpp
@@ -61,46 +61,46 @@ bool StackTraceStack::pushCurrent() {
   }
   node->frameCount = n;
 
-  node->next = top_;
-  top_ = node;
+  node->next = state_[kTopIdx];
+  state_[kTopIdx] = node;
   return true;
 }
 
 bool StackTraceStack::pop() {
   checkGuard();
-  if (!top_) {
+  if (!state_[kTopIdx]) {
     return false;
   }
 
-  auto node = top_;
-  top_ = node->next;
+  auto node = state_[kTopIdx];
+  state_[kTopIdx] = node->next;
   node->deallocate();
   return true;
 }
 
 bool StackTraceStack::moveTopFrom(StackTraceStack& other) {
   checkGuard();
-  if (!other.top_) {
+  if (!other.state_[kTopIdx]) {
     return false;
   }
 
-  auto node = other.top_;
-  other.top_ = node->next;
-  node->next = top_;
-  top_ = node;
+  auto node = other.state_[kTopIdx];
+  other.state_[kTopIdx] = node->next;
+  node->next = state_[kTopIdx];
+  state_[kTopIdx] = node;
   return true;
 }
 
 void StackTraceStack::clear() {
   checkGuard();
-  while (top_) {
+  while (state_[kTopIdx]) {
     pop();
   }
 }
 
 StackTrace* StackTraceStack::top() {
   checkGuard();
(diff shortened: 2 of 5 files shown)


folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -85,18 +98,12 @@ class StackTraceStack {
...
+      assert(guard1_ == 0 && guard2_ == 0);
+    }
   }
 
-#ifndef NDEBUG
-  uintptr_t guard1_;
-#endif
   Node* top_;
-#ifndef NDEBUG
-  uintptr_t guard2_;
```
yfeldblum: This is tricky because the intention is for the two guards to surround `top_`. If they are both in the base class, then they both precede `top_` rather than surrounding `top_`.

This may be solved by having three base classes....

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -32,9 +34,20 @@ struct StackTrace {
   uintptr_t addresses[kMaxFrames];
 };
 
+template<bool IsDebug = false>
+class StackTraceStackBase {};
```
yfeldblum: Would need to be in a detail namespace.

folly/FixedString.h
```diff
@@ -1110,22 +1110,16 @@ class BasicFixedString : private detail::fixedstring::FixedStringBase {
    * \return `*(data() + i)`
    */
   constexpr Char& operator[](std::size_t i) noexcept {
-#ifdef NDEBUG
-    return data_[i];
-#else
-    return data_[detail::fixedstring::checkOverflow(i, size_)];
-#endif
+    return kIsDebug ? data_[detail::fixedstring::checkOverflow(i, size_)] :
```
yfeldblum: C++11 constexpr rules are no longer required, so:

```
i = kIsDebug ? detail::fixedstring::checkOverflow(i, size_) : i;
return data_[i];
```

Or even just define `checkOverflowIfDebug` and:

```
return data_[detail::fixedstring::checkOverflowIfDebug(i, size_)];
```

Here and below.

folly/experimental/exception_tracer/StackTrace.h
JoeLoser: I was worried about the ordering of the NSDM when I made the base class. Thanks for the confirmation of my suspicion. Not ideal to have the three base classes, but I do prefer this over the preprocessor still. As such, made the change; I don't see many alternatives to force a well-specified ordering of the NSDM in a conditional manner.

JoeLoser:
I'll rebase tonight since 598b2103 landed this morning which has some of the changes in this PR.

JoeLoser:
Just rebased.

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -32,11 +34,37 @@ struct StackTrace {
   uintptr_t addresses[kMaxFrames];
 };
 
-// note: no constructor so this can be __thread.
-// A StackTraceStack MUST be placed in zero-initialized memory.
-class StackTraceStack {
+namespace detail {
+template<bool IsDebug = false>
```
yfeldblum: Here is how I might structure it.

```
template <size_t, bool IsDebug = false>
class StackTraceStackBaseGuard {
  uintptr_t guard() const {
    return 0;
  }
};
template <size_t I>
class StackTraceStackBaseGuard<I, true> {
 protected:
  uintptr_t guard() const {
    return guard_;
  }

 private:
  uintptr_t guard_;
};

// ...

class StackTraceStack
    : public detail::StackTraceStackBaseGuard<1, kIsDebug>,
      public detail::StackTraceStackBase,
      public detail::StackTraceStackBaseGuard<2, kIsDebug> {
  void checkGuard() const {
    assert(detail::StackTraceStackBaseGuard<1, kIsDebug>::guard() == 0);
    assert(detail::StackTraceStackBaseGuard<2, kIsDebug>::guard() == 0);
  }
};
```

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -85,18 +113,10 @@ class StackTraceStack {
...
-#endif
+    if (kIsDebug) {
+      assert(guard1_ == 0 && guard2_ == 0);
+    }
   }
-
-#ifndef NDEBUG
-  uintptr_t guard1_;
-#endif
-  Node* top_;
```
yfeldblum: Thinking again ... as a completely different approach to structuring this, and probably way simpler:

```
class StackTraceStack {
 public:
  void checkGuard() const {
    assert(state_[0] == 0 && state_[2] == 0);
  }

 private:
  template <typename Self>
  auto& top(Self& self) {
    return kIsDebug ? self.state_[1] : self.state_[0];
  }

  Node* state_[kIsDebug ? 3 : 1];
};
```

folly/experimental/exception_tracer/StackTrace.h
JoeLoser: A non-type template for the particular guard is a better approach, but I ended up closer to your other suggestion below which is a lot simpler.

folly/experimental/exception_tracer/StackTrace.h
JoeLoser: Nice. I like this. I implemented effectively this, but I did not see an easy way to run this against the tests in `folly/experimental/exception_tracer`. In my build, `FOLLY_USE_SYMBOLIZER` is defined, so the library gets built, but not the tests from what I can tell. What is the recommended way to run these tests, or it just a matter of adding it to the top-level `CMakeLists.txt`?

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -67,13 +69,16 @@ class StackTraceStack {
...
   bool empty() const {
-    return !top_;
+    return !top();
   }
 
   /**
    * Return the top stack trace, or nullptr if the stack is empty.
    */
-  StackTrace* top();
+  StackTrace* top() {
```
yfeldblum: There needs to be an internal function which returns a reference to the pointer, which all of the internal code uses rather than using this function.

Or all of the internal code should look like:

```
static constexpr size_t kTopIdx  = kIsDebug ? 1 : 0;
// ...
/*...*/ state_[kTopIdx] /*...*/;
```

folly/experimental/exception_tracer/StackTrace.cpp
```diff
@@ -61,48 +61,43 @@ bool StackTraceStack::pushCurrent() {
   }
   node->frameCount = n;
 
-  node->next = top_;
-  top_ = node;
+  node->next = top();
+  top() = node;
```
yfeldblum: E.g. this is no longer correct.

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -83,20 +88,14 @@ class StackTraceStack {
 
  private:
   // In debug mode, we assert that we're in zero-initialized memory by
-  // checking that the two guards around top_ are zero.
+  // checking that the two guards around the Node* from top()
   void checkGuard() const {
-#ifndef NDEBUG
-    assert(guard1_ == 0 && guard2_ == 0);
-#endif
+    if (kIsDebug) {
```
yfeldblum: This no longer needs a surrounding `kIsDebug` check.

JoeLoser:
@yfeldblum is the latest push closer to what you had in mind with `kTopIdx`?

yfeldblum:
Yep!

JoeLoser:
@yfeldblum please reimport. There was a bug where I dropped the `other` on the right-hand side in https://github.com/facebook/folly/commit/ea306d5feeffef6b446e24e5a873b9648b008d6d#diff-3d8380ef2d1172970248cd3b8944f700R87

folly/experimental/exception_tracer/StackTrace.h
```diff
@@ -82,21 +84,15 @@ class StackTraceStack {
   StackTrace* next(StackTrace* p);
 
  private:
+  static constexpr size_t kTopIdx = kIsDebug ? 1 : 0;
+
   // In debug mode, we assert that we're in zero-initialized memory by
-  // checking that the two guards around top_ are zero.
+  // checking that the two guards around the Node* from top()
```
yfeldblum: Dropped `are zero.`

JoeLoser: Thanks for spotting that. Just fixed.

PR #1133, opened 2019-05-19
Optimize ProducerConsumerQueue

Deaod:
First described in a paper available here:
http://www.cse.cuhk.edu.hk/~pclee/www/pubs/ancs09poster.pdf

Tries to avoid unnecessary sharing of read/write index. Generally profitable optimization, but the effects become less strong with larger record sizes and larger buffer sizes. Up to 3 times larger throughput at
8 bytes record size.

Time on the following two charts is (t(_read_) + t(_write_)) / 2.

Record Size 8
Record Size 64

P.S.: Throughput gets better on x64 if you use `alignas(128)` or even `alignas(256)` instead of `alignas(64)`. The charts above were captured using `alignas(128)`.

### DIFF
--- a/folly/ProducerConsumerQueue.h
+++ b/folly/ProducerConsumerQueue.h
@@ -36,7 +36,7 @@ namespace folly {
  * without locks.
  */
 template <class T>
-struct ProducerConsumerQueue {
+struct alignas(hardware_destructive_interference_size) ProducerConsumerQueue {
   typedef T value_type;
 
   ProducerConsumerQueue(const ProducerConsumerQueue&) = delete;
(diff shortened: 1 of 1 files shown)


Deaod:
May be covered by https://patents.google.com/patent/US20100332755

yfeldblum:
Is there a specific need for `ProducerConsumerQueue` v.s. `USPSCQueue`?

Deaod:
I assume you need a `ProducerConsumeQueue` if you want a small (<=L1D size), fixed upper bound on memory usage. If you need an upper bound `DSPSCQueue` is probably appropriate, `USPSCQueue` is for when you dont need an upper bound. `ProducerConsumeQueue` is also useful when you cant use `malloc` after initialization.

I can benchmark `DSPSCQueue` and `USPSCQueue` and get back to you, if you want.

PR #1142, opened 2019-05-31
Refer to 7-bit tag in comment

leventov:


leventov:
Another problem: https://github.com/facebook/folly/blob/9feea214fb9e8d4a7a141155b68643a393193c2a/folly/container/detail/F14Table.h#L2021-L2022 doesn't seem that this function returns anything now

PR #1145, opened 2019-05-31
Add some explicit casts to prevent some compiler warnings

acoates-ms:
As part of trying to get our internal usage of folly off a fork, we have various compiler warnings set ridiculously high.

This gets rid of a couple of warnings that we hit in folly code.

### DIFF
--- a/folly/Utility.h
+++ b/folly/Utility.h
@@ -295,8 +295,11 @@ constexpr auto to_signed(T const& t) -> typename std::make_signed<T>::type {
   // note: static_cast<S>(t) would be more straightforward, but it would also be
   // implementation-defined behavior and that is typically to be avoided; the
   // following code optimized into the same thing, though
+FOLLY_PUSH_WARNING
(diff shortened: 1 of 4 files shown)


folly/Utility.h
```diff
@@ -295,8 +295,11 @@ constexpr auto to_signed(T const& t) -> typename std::make_signed<T>::type {
   // note: static_cast<S>(t) would be more straightforward, but it would also be
   // implementation-defined behavior and that is typically to be avoided; the
   // following code optimized into the same thing, though
+FOLLY_PUSH_WARNING
+FOLLY_MSVC_DISABLE_WARNING(4018)
```
yfeldblum: Can this be done without suppressing warnings?

```
constexpr auto m = static_cast<T>(std::numeric_limits<S>::max());
return m < t ? ... : ...;
```

acoates-ms: Yes I like that better.  I initially was avoiding using static_cast due to the comment above, but the constexpr static_cast makes sense.

PR #1147, opened 2019-06-01
Allow building x86 and arm on msvc

acoates-ms:
A couple of changes to support building folly in MSVC/win32 for x86 and arm platforms.

### DIFF
--- a/folly/portability/Builtins.h
+++ b/folly/portability/Builtins.h
@@ -50,10 +50,20 @@ FOLLY_ALWAYS_INLINE int __builtin_clzl(unsigned long x) {
   return __builtin_clz((unsigned int)x);
 }
 
+#if defined(_M_IX86) || defined(_M_ARM)
+FOLLY_ALWAYS_INLINE int __builtin_clzll(unsigned long long x) {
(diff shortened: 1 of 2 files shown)


folly/portability/Builtins.h
```diff
@@ -50,10 +50,20 @@ FOLLY_ALWAYS_INLINE int __builtin_clzl(unsigned long x) {
   return __builtin_clz((unsigned int)x);
 }
 
+#if defined(_M_IX86) || defined(_M_ARM)
+FOLLY_ALWAYS_INLINE int __builtin_clzll(unsigned long long x) {
+  if (x == 0)
```
Orvid: FB (and folly) code style requires braces even on single line conditional statements here and in a few places below as well :)

acoates-ms: Noted.  We are using this for react-native-windows, and not currently using the sockets support.  Mostly we just need folly::dynamic to be able to use the shared c++ code of react-native.  But I haven't done a deep dive on making sure that the absolutely minimum amount of folly actually gets hit for that.

PR #1149, opened 2019-06-04
Simplify _fs UDL in FixedString.h

JoeLoser:
Summary:
- Simplify `_fs` user defined literal by removing the preprocessor branch.
- Since Folly only supports C++14 now, `#if __cplusplus >= 201402L` will always
  evaluate to true.


JoeLoser:

PR #1150, opened 2019-06-04
Cut backport of std::make_unique

JoeLoser:
- Previously, folly supported compilers which did not implement
  `std::make_unique`, so there exists a backport implementation.
- Now that folly only supports C++14-compliant compilers, remove the
  backport.
- Fix call sites which were relying on unqualified `make_unique` being
  found in the folly namespace to explicitly use `std::make_unique`.

### DIFF
--- a/folly/Memory.h
+++ b/folly/Memory.h
@@ -240,41 +240,6 @@ size_t allocationBytesForOverAligned(size_t n) {
 }
 
 /**
- * For exception safety and consistency with make_shared. Erase me when
- * we have std::make_unique().
- *
- * @author Louis Brandy (ldbrandy@fb.com)
- * @author Xu Ning (xning@fb.com)
- */
-
-#if __cplusplus >= 201402L || __cpp_lib_make_unique >= 201304L || \
-    (__ANDROID__ && __cplusplus >= 201300L) || _MSC_VER >= 1900
-
-/* using override */ using std::make_unique;
-
-#else
-
-template <typename T, typename... Args>
-typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(Args&&... args) {
-  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
-}
-
-// Allows 'make_unique<T[]>(10)'. (N3690 s20.9.1.4 p3-4)
-template <typename T>
-typename std::enable_if<std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(const size_t n) {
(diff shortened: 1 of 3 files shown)


JoeLoser:
Ping @yfeldblum

yfeldblum:
> Now that folly only supports C++14-compliant compilers, remove the backport.

Note that some parts of folly may be used with C++14 compilers but with C++11 standard libraries.

JoeLoser:
I wasn't aware of that bit, but are there any parts of folly in particular which need to be C++11-standard-library compatible? Based on the unqualified `make_unique` calls, it seems like nearly all of the call sites in folly namespace already use `std::make_unique`, so anyone depending on those headers is already using a standard library which has implemented `std::make_unique`. Do we have real code in the wild depending on this backport besides the two call sites I fixed in folly?

Does the backport live in folly as an easy way for downstream projects that are still using C++11-standard-library to get it, despite most of folly library already using `std::make_unique`?

yfeldblum:
> Are there any parts of folly in particular which need to be C++11-standard-library compatible?

Well, compatible with the gcc49 implementation of the standard library in C++14 mode. This version has a partial but not a full implementation of the C++14 standard library spec.

Note that the gcc49 implementation of the standard library in C++14 mode ships with `std::make_unique`.

Beyond `std::make_unique`, for other names, there isn't really a system to know in advance.

yfeldblum:
Re this PR, there is code internally which still uses `folly::make_unique`, so that would need to be changed first.

JoeLoser:
I grepped through Thrift and Eden and there weren't any explicit uses of `folly::make_unique` (though there is a lot of `using std::make_unique` and `using namespace std` at the top of `.cpp` files to bring in `make_unique` into lookup for ADL.

Do you have any interest in fixing the internal code in a sweeping codemod or the like?

yfeldblum:
Yes if it's easy, no if it's hard. :)

I did just land one codemod internally, although more will be needed.

Based on a quick search, it looks like [React Native](https://github.com/facebook/react-native) is using `folly::make_unique`.

JoeLoser:
[Here](https://github.com/facebook/react-native/pull/26730) is a PR which replaces the usages of `folly::make_unique` with `std::make_unique` in the `react-native` project.

JoeLoser:
ping @yfeldblum since https://github.com/facebook/react-native/pull/26730 landed

Orvid:
I believe this has been done as part of the more recent cleanup efforts. It didn't land initially as part of this PR due to the need for internal codemods in order to keep things building.

PR #1151, opened 2019-06-04
Cut GCC 5.0 preprocessor guard in Traits.h

JoeLoser:
Summary:
- Cut `#if __GNUC_PREREQ(5, 0)` preprocessor directive as folly does not
  support any GCC version lower than 5.1.
- Change from `FOLLY_GNU_DISABLE_WARNING` to `FOLLY_GCC_DISABLE_WARNING`
  since Clang does not support `-Wbool-compare` option.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -657,9 +657,7 @@ namespace detail {
 // in order to not prevent all calling code from using it.
 FOLLY_PUSH_WARNING
 FOLLY_GNU_DISABLE_WARNING("-Wsign-compare")
-#if __GNUC_PREREQ(5, 0)
-FOLLY_GNU_DISABLE_WARNING("-Wbool-compare")
-#endif
+FOLLY_GCC_DISABLE_WARNING("-Wbool-compare")
 FOLLY_MSVC_DISABLE_WARNING(4388) // sign-compare
 FOLLY_MSVC_DISABLE_WARNING(4804) // bool-compare
 


folly/Traits.h
```diff
@@ -657,9 +657,7 @@ namespace detail {
 // in order to not prevent all calling code from using it.
 FOLLY_PUSH_WARNING
 FOLLY_GNU_DISABLE_WARNING("-Wsign-compare")
-#if __GNUC_PREREQ(5, 0)
```
yfeldblum: This currently guards against older versions of clang, too.

```
folly/Traits.h:660:1: error: unknown warning group '-Wbool-compare', ignored [-Werror,-Wunknown-warning-option]
FOLLY_GNU_DISABLE_WARNING("-Wbool-compare")
^
<scratch space>:55:25: note: expanded from here
 GCC diagnostic ignored "-Wbool-compare"
```

JoeLoser: Instead, should we make this guard a clang specific guard?

yfeldblum: Yeah. I guess let's change this to guard based on `__clang__`. I don't see any version of Clang which supports this warning.

Can even ask the Clang folks to implement this warning for parity with GCC, and then update the guard here.

JoeLoser: I just changed it so we only ignore `-Wbool-compare` for GCC only since Clang doesn't support this flag.

I filed https://bugs.llvm.org/show_bug.cgi?id=42130 for asking for Clang to support this flag for symmetry with GCC.

JoeLoser:

PR #1152, opened 2019-06-04
Avoid C++-only attribute in CPortability.h 

JoeLoser:
Summary:
- In defining `FOLLY_EXPORT`, we define it differently based on the GCC
  version. In the case of using GCC or Clang 5+, it will rely on
  `[[gnu::visibility("default")]]` which is a C++-specific attribute.
- Since `CPortability.h` is intended to be made to be included for C
  projects, remove `[[gnu::visibility("default")]]` and always define
  `FOLLY_EXPORT` using `__attribute__` syntax which works in C.

### DIFF
--- a/folly/CPortability.h
+++ b/folly/CPortability.h
@@ -135,11 +135,7 @@
  * Macro for marking functions as having public visibility.
  */
 #if defined(__GNUC__)
-#if __GNUC_PREREQ(4, 9)
-#define FOLLY_EXPORT [[gnu::visibility("default")]]
-#else
 #define FOLLY_EXPORT __attribute__((__visibility__("default")))
-#endif
 #else
 #define FOLLY_EXPORT
 #endif


Orvid:
This is C portability, this shouldn't be using C++14 attributes to begin with :(

yfeldblum:
@Orvid Is that even worth it?

Orvid:
We have C code internally that includes this header, so it really should be only C compatible macros in the portability header for C code :)

JoeLoser:
@Orvid sounds like the easiest thing to do to keep C-compatibility is just to always

```
#define FOLLY_EXPORT __attribute__((__visibility__("default")))
```

regardless of `GNUC` version. Does that work for you?

Orvid:
Yep, that would work.

JoeLoser:
@Orvid just pushed up that change then so that this is now correctly C-compatible in all cases.

yfeldblum:
High-level, I prefer moving to C++ attribute syntax. I'm not sure why we need C compatibility - AFAICT from a quick check there isn't C code internally that is using this macro.

JoeLoser:
I prefer C++ attribute syntax over the `__attribute__` syntax, too. As @Orvid says though, given the name and intent of this header, it seems we should stick with the C-style `__attribute__` syntax. This would be the element of least surprise to users including this header, even if it sounds like this particular macro may not have any users (yet)?

PR #1153, opened 2019-06-04
Remove __GNUC__ >=5 checks

JoeLoser:
Summary:
- Remove preprocessor guards of `__GNUC__ >= 5` as they will always
  evaluate to true.


folly/experimental/test/StringKeyedBenchmark.cpp
```diff
@@ -59,7 +59,7 @@ static const folly::StringPiece lookupPiece{
 #define FOLLY_HAVE_COMPARE_EQUIVALENT 1
 #endif
 
-#if !defined(FOLLY_HAVE_COMPARE_EQUIVALENT) && __GNUC__ >= 5
```
JoeLoser: @yfeldblum it seems we can just get rid of `FOLLY_HAVE_COMPARE_EQUIVALENT` I think. It will be `1` for Clang 3.4 onward. GCC 5.1 is fine with it too. What's the min Clang version we support?

JoeLoser: https://github.com/facebook/folly/pull/1157 is a separate MR which removes `FOLLY_HAVE_COMPARE_EQUIVALENT` that you can test internally to make sure it doesn't break anything.

JoeLoser:
ping @yfeldblum -- just rebased. The two preprocessor guards on `__GNUC__ >=5` were previously not checking things for lower Clang versions (<5), but they should work fine with Clang 3.5 for example.

PR #1155, opened 2019-06-05
Add assert in Range::size()

JoeLoser:
Summary:
- Prior to commit 8cb615a27594078056b2e4ec2350660e594f5a89
  (Differential Revision: D3394612), `Range:size()` had an assert checking
  that the beginning iterator of the range is less than or equal to the end
  iterator. Unfortunately, due to GCC bug 71448, the `assert` was not allowed
  if we wanted `size()` to remain a `constexpr` function.
- This bug is no longer an issue with our current supported versions of GCC.
  As such, add the assert back in.

### DIFF
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -421,12 +421,7 @@ class Range {
   }
 
   constexpr size_type size() const {
-    // It would be nice to assert(b_ <= e_) here.  This can be achieved even
-    // in a C++11 compatible constexpr function:
-    // http://ericniebler.com/2014/09/27/assert-and-constexpr-in-cxx11/
-    // Unfortunately current gcc versions have a bug causing it to reject
-    // this check in a constexpr function:
-    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71448
+    assert(b_ <= e_);
     return size_type(e_ - b_);
   }
   constexpr size_type walk_size() const {


JoeLoser:

yfeldblum:
Seems fine with clang.

With gcc,

```
folly/test/RangeTest.cpp: In member function ‘virtual void StringPiece_Constexpr_Test::TestBody()’:
folly/test/RangeTest.cpp:319:3: error: non-constant condition for static assertion
   static_assert(hello1.size() == 5, "hello size should be 5 at compile time");
   ^
folly/test/RangeTest.cpp:319:28:   in constexpr expansion of ‘hello1.folly::Range<Iter>::size<const char*>()’
folly/test/RangeTest.cpp:319:3: error: ‘(((const char*)"hello") > (((const char*)"hello") + 5u))’ is not a constant expression
folly/test/RangeTest.cpp:323:3: error: non-constant condition for static assertion
   static_assert(hello2.size() == 5, "hello size should be 5 at compile time");
   ^
folly/test/RangeTest.cpp:323:28:   in constexpr expansion of ‘hello2.folly::Range<Iter>::size<const char*>()’
folly/test/RangeTest.cpp:323:3: error: ‘(((const char*)"hello") > (((const char*)"hello") + 5u))’ is not a constant expression
folly/test/RangeTest.cpp: In member function ‘virtual void Range_ConstexprAccessors_Test::TestBody()’:
folly/test/RangeTest.cpp:1417:3: error: non-constant condition for static assertion
   static_assert(piece.size() == 6u, "");
   ^
folly/test/RangeTest.cpp:1417:27:   in constexpr expansion of ‘piece.folly::Range<Iter>::size<const char*>()’
folly/test/RangeTest.cpp:1417:3: error: ‘(((const char*)"hello") > (((const char*)"hello") + 6u))’ is not a constant expression
```

With msvc,

```
folly\test\rangetest.cpp(1162): error C2131: expression did not evaluate to a constant
  folly\range.h(424): note: failure was caused by unevaluable pointer value
  folly\test\rangetest.cpp(1162): note: while evaluating 'folly::Range<const int *>::size(&numArrayRange)'
```

JoeLoser:
That's unfortunate that MSVC and GCC still reject the code. As such, I'll close this PR and we can revisit in the future when those compilers get upgraded internally.

PR #1169, opened 2019-06-16
FreeBSD compatibility patch

yurivict:


folly/Subprocess.cpp
```diff
@@ -24,6 +24,7 @@
 #include <sys/prctl.h>
 #endif
 #include <fcntl.h>
+#include <unistd.h>
```
Orvid: This is already included via the `folly/portability/Unistd.h` include further down. Remove this include.

folly/portability/Stdlib.cpp
```diff
@@ -147,6 +147,10 @@ int unsetenv(const char* name) {
 #include <string>
 #include <vector>
 
+#if defined(__FreeBSD__) || defined(__DragonFly__)
```
Orvid: Is there really nowhere in the libc used by freebsd that declares `environ`?

yurivict:
> Is there really nowhere in the libc used by freebsd that declares environ?

The man page doesn't mention  any headers:
```
ENVIRON(7)         FreeBSD Miscellaneous Information Manual         ENVIRON(7)

NAME
     environ — user environment

SYNOPSIS
     extern char **environ;
```
This may be an oversight, but it is this way currently.

-------------------------

I'll update the pull request for the other things.

yurivict:
Overcome by events.

Issue #1121, opened 2019-05-02
Unable to build on Debian 9.8 Stretch


pcampr:
cmake wrongly fails with:

```
-- Performing Test FOLLY_CPP_ATOMIC_BUILTIN
-- Performing Test FOLLY_CPP_ATOMIC_BUILTIN - Failed
-- Performing Test FOLLY_CPP_ATOMIC_WITH_LIBATOMIC
-- Performing Test FOLLY_CPP_ATOMIC_WITH_LIBATOMIC - Failed
CMake Error at CMake/folly-deps.cmake:152 (message):
  unable to link C++ std::atomic code: you may need to install GNU libatomic
```

Cmake uses `check_cxx_source_compiles` in `CMake/folly-deps.cmake:152`, but the result is falsely negative (`FOLLY_CPP_ATOMIC_BUILTIN` is false, but should be true). 

---

When I extracted the same check to standalone CMakeLists.txt, the result was correct:

```
cmake_minimum_required(VERSION 3.0.2 FATAL_ERROR)
include(CheckCXXSourceCompiles)
check_cxx_source_compiles("
  #include <atomic>
  int main(int argc, char** argv) {
	struct Test { int val; };
	std::atomic<Test> s;
	return static_cast<int>(s.is_lock_free());
  }"
  FOLLY_CPP_ATOMIC_BUILTIN
)
message(STATUS "FOLLY_CPP_ATOMIC_BUILTIN=" ${FOLLY_CPP_ATOMIC_BUILTIN})
```

here, the result was correct: `FOLLY_CPP_ATOMIC_BUILTIN=1`

Some other capabilities are detected wrongly too, e.g. `FOLLY_HAVE_MEMRCHR`, `FOLLY_HAVE_RECVMMSG`, `FOLLY_HAVE_CLOCK_GETTIME`.

Tested on fresh Debian 9.8 Stretch, with:
```
$ sudo apt -y install cmake g++ libboost-all-dev libevent-dev libdouble-conversion-dev libgoogle-glog-dev libgflags-dev libiberty-dev liblz4-dev liblzma-dev libsnappy-dev zlib1g-dev binutils-dev libjemalloc-dev libssl-dev
$ cmake --version
cmake version 3.7.2
$ g++ --version
g++ (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
```

---

I tried to force set these options:
```
# added to the top of main CMakeLists.txt:
[...]

PR #1170, opened 2019-06-19
Fix build on Debian

jmccl:
Fix for facebook/folly#1121

This issue is that folly-deps only handles gflags in cmake 'module mode', not 'config mode'. I copied the fix from the way fizz handles the [same issue](https://github.com/facebookincubator/fizz/blob/343f01a36c47168b58c0d2c2b3869f15510b65a7/fizz/CMakeLists.txt).


jmccl:
@yfeldblum 

I don't have access to phabricator. Is this pr still on its way to getting to master, or has there been some problem?

jmccl:
Fixed by subsequent commits.

PR #1173, opened 2019-06-19
handle noexcept in SignatureOf_ and ArgTypes_

### LINKED ISSUE #737: Poly does not work with noexcept methods
The `folly/Poly.md` document does not mention `noexcept` at all. Consider this example

```
#include <folly/Poly.h>
#include <iostream>

struct IDrawable {
  template <typename Base> struct Interface : Base {
    void draw(std::ostream& out) const /* noexcept */ {
      folly::poly_call<0>(*this, out);
    }
  };

  template <typename T> using Members = folly::PolyMembers<&T::draw>;
};

struct Circle {
  void draw(std::ostream& out) const /* noexcept */ { out << "Cirlce.\n"; }
};

using Drawable = folly::Poly<IDrawable>;

int main() {
  Drawable obj = Circle{};
  obj.draw(std::cout);
}
```

Uncommenting any `noexcept` breaks on my machine (using clang from 2017-11-25 and folly from 2017-12-25)

I attached the error message for uncommenting `Circle::draw`'s  `noexcept` specifier.

[error.log](https://github.com/facebook/folly/files/1639659/error.log)

cstratopoulos:
Fixes #737 

My use of `std::add_pointer_t` deviates in style from the alias decls in other `SignatureOf_` specializations because it is not possible with that syntax to declare a pointer to `noexcept` function.

### DIFF
--- a/folly/detail/PolyDetail.h
+++ b/folly/detail/PolyDetail.h
@@ -397,6 +397,16 @@ struct SignatureOf_<R (C::*)(As...) const, I> {
   using type = Ret<R, I> (*)(Data const&, Arg<As, I>...);
 };
 
+template <class R, class C, class... As, class I>
+struct SignatureOf_<R(C::*)(As...) noexcept, I> {
+	using type = std::add_pointer_t<Ret<R, I>(Data&, Arg<As, I>...) noexcept>;
+};
+
+template <class R, class C, class... As, class I>
+struct SignatureOf_<R(C::*)(As...) const noexcept, I> {
+	using type = std::add_pointer_t<Ret<R, I>(Data const&, Arg<As, I>...) noexcept>;
+};
+
 template <class R, class This, class... As, class I>
 struct SignatureOf_<R (*)(This&, As...), I> {
   using type = Ret<R, I> (*)(Data&, Arg<As, I>...);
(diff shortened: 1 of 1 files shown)


Orvid:
We ended up having to revert this, as some code was still on C++14 and we didn't realize it before I merged this. `noexcept` isn't part of the type system until C++17.

cstratopoulos:
Shoot that slipped my mind too. I'm a bit confused here because [p0012](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0012r1.html) belabors this point a fair bit:
>4.12 [conv.fctptr] Function pointer conversions

>A prvalue of type "pointer to `noexcept` function" can be converted to a prvalue of type "pointer to function"

and then the updates to the template deduction wording defer to `[conv.fctptr]` saying that the same conversion should be allowed in template deduction, which is also what the accepted answer [here](https://stackoverflow.com/questions/38581776/noexcept-to-become-part-of-the-type-system) suggests should take place.

However clang and gcc are both fine with conversion in a variable declaration but not in template deduction: https://godbolt.org/z/-WXRqS

Maybe I'm overlooking something because this seems too obvious as a compiler bug somehow.

Anyway, maybe it would make sense to conditionally provide those specializations via the `__cpp_noexcept_function_type` feature test macro?

cstratopoulos:
With some more thought I realize the behavior is just as expected: 
>the deduction process attempts to find template argument values that will make the deduced A identical to A

and the unspecialized class template declaration gives a deduction that doesn't require function pointer conversion.

I now suspect a macro-based workaround would indeed be necessary to keep things working as-written for both C++14 and C++17. 

As an aside, a different, possibly more clunky approach could be to use the return type of declared-but-not-defined functions rather than nested `type` aliases in class template specializations:
https://godbolt.org/z/-hp4kv
This wouldn't require macro-based workarounds but we'd lose `noexcept`-ness in C++17 and it would probably be a bit awkward.

@Orvid let me know which approach sounds better and if you like I can open a fresh PR for take two on this one.

Orvid:
The macro based workaround should be cleaner and wouldn't lose the `noexcept`-ness, so let's go with that one :)

PR #1174, opened 2019-06-19
do not rely on deprecated __cpp_template_auto macro

cstratopoulos:
Introduces a `FOLLY_NTTP_AUTO` macro to test if NTTP is available.

As per [cppreference](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros), the macro `__cpp_nontype_template_parameter_auto` is the correct feature test macro to use for the NTTP used to define poly member list. The `__cpp_template_auto` version is checked too, but it was [deprecated](https://fossies.org/diffs/gcc/8.2.0_vs_8.3.0/gcc/c-family/ChangeLog-diff.html) in gcc 8.3 and not supported at all in MSVC.

### DIFF
--- a/folly/detail/PolyDetail.h
+++ b/folly/detail/PolyDetail.h
@@ -32,6 +32,12 @@
 
 #include <folly/PolyException.h>
 
+#if defined(__cpp_template_auto) || defined(__cpp_nontype_template_parameter_auto)
+#define FOLLY_NTTP_AUTO 1
+#else
+#define FOLLY_NTTP_AUTO 0
+#endif
+
 namespace folly {
 /// \cond
 namespace detail {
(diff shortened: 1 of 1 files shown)


folly/detail/PolyDetail.h
```diff
@@ -32,6 +32,12 @@
 
 #include <folly/PolyException.h>
 
+#if defined(__cpp_template_auto) || defined(__cpp_nontype_template_parameter_auto)
+#define FOLLY_NTTP_AUTO 1
```
yfeldblum: Namespacing: let's use the name `FOLLY_POLY_NTTP_AUTO`.

cstratopoulos: Good call, done!

cstratopoulos: Oops misspelled poly in my first commit 😅

PR #1177, opened 2019-06-27
Clang fixes

neheb:


folly/String.h
```diff
@@ -185,9 +185,12 @@ std::string& stringAppendf(
  * As with vsnprintf() itself, the value of ap is undefined after the call.
  * These functions do not call va_end() on ap.
  */
-std::string stringVPrintf(const char* format, va_list ap);
-void stringVPrintf(std::string* out, const char* format, va_list ap);
-std::string& stringVAppendf(std::string* out, const char* format, va_list ap);
+std::string stringVPrintf(const char* format, va_list ap)
```
yfeldblum: General style - attributes should precede the signature. Here and elsewhere.

folly/ClockGettimeWrappers.cpp
```diff
@@ -23,7 +23,6 @@
 #include <time.h>
 
 #ifndef _WIN32
-#define _GNU_SOURCE 1
```
yfeldblum: Is this known not to be needed? Here and elsewhere.

neheb: _GNU_SOURCE is already defined here: https://github.com/facebook/folly/blob/master/CMake/FollyCompilerUnix.cmake#L23

I removed as it throws a redefinition error. Defining _GNU_SOURCE to 1 is wrong anyway.

folly/String.h
neheb: I only matched the way it already was.

neheb:
I think this is what was meant by the first comment.

neheb:
Rebased and added a bunch more. Please review.

PR #1178, opened 2019-06-28
Handle noexcept in SignatureOf_ and ArgTypes_, take 2

cstratopoulos:
Second attempt at https://github.com/facebook/folly/pull/1173

This time the class template specializations are guarded with the `__cpp_noexcept_function_type` feature test macro to maintain compatibility with C++14. 

Open to replacing the raw macro with something like `FOLLY_POLY_NOEXCEPT_FUNCTION_TYPE` but that was not my first approach since it's more characters and there is not an extra case to test like with `FOLLY_POLY_NTTP_AUTO`.

### DIFF
--- a/folly/detail/PolyDetail.h
+++ b/folly/detail/PolyDetail.h
@@ -404,6 +404,7 @@ struct SignatureOf_<R (C::*)(As...) const, I> {
   using type = Ret<R, I> (*)(Data const&, Arg<As, I>...);
 };
 
+#ifdef __cpp_noexcept_function_type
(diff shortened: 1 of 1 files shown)


Orvid:
I'm not sure which version of MSVC switched to actually defining feature macros like `__cpp_noexcept_function_type`, but we need to support VS 2017 Update 6, VS 2017 Update 9, and VS 2019  currently. As Poly wasn't supported prior to this at all, we can probably get away with only supporting update 9 and VS 2019, or even possibly just VS2019 if needed.

PR #1182, opened 2019-07-02
Compatibility with VS2019

cstratopoulos:
Some changes to get Poly compiling with VS2019 (and correcting an oversight from https://github.com/facebook/folly/pull/1174)

The change to `detail/TypeList.h` is straightforward I think.

The introduction of `getOpsImpl` I find a bit distasteful...I will remark that as of Visual Studio 16.2.0 Preview 3.0 it is not necessary, but it is still needed on Visual Studio 16.1.5 (latest release, non preview). 

@Orvid in your comment [here](https://github.com/facebook/folly/pull/1178#issuecomment-506864383)  you mentioned the question of supporting VS 2017 or VS 2019 given that Poly had no MSVC support at all beforehand. I re-downloaded VS 2017 on my machine to test against the fixes added here and the situation seemed discouraging with lots more errors and perhaps less obvious workarounds. 

I have not made any changes to the CMake stuff for Poly on MSVC: https://github.com/facebook/folly/blob/master/CMakeLists.txt#L235

I'm not sure if once we resolve the issue above we might want to add some logic maybe [here](https://github.com/facebook/folly/blob/master/CMakeLists.txt#L63) to handle which version of MSVC is approved for poly; will wait for further direction on that.

### DIFF
--- a/folly/detail/TypeList.h
+++ b/folly/detail/TypeList.h
@@ -168,7 +168,7 @@ using If = MetaApply<impl::If_<If_>, Then, Else>;
  */
 template <template <class...> class C, class... Ts>
 class MetaDefer {
-  template <template <class...> class D = C, class = D<Ts...>>
+  template <class = C<Ts...>>
   static char (&try_(int))[1];
   static char (&try_(long))[2];
   struct Result {
--- a/folly/detail/TypeList.h
+++ b/folly/detail/TypeList.h
@@ -168,8 +168,10 @@ using If = MetaApply<impl::If_<If_>, Then, Else>;
  */
 template <template <class...> class C, class... Ts>
 class MetaDefer {
-  template <class = C<Ts...>>
+  template <template <class...> class D, class = D<Ts...>>
   static char (&try_(int))[1];
+
+  template <template <class...> class D, class = void>
   static char (&try_(long))[2];
   struct Result {
     using type = C<Ts...>;
@@ -177,7 +179,7 @@ class MetaDefer {
 
  public:
   template <class... Us>
-  using apply = _t<If<sizeof(try_(0)) - 1 || sizeof...(Us), Empty, Result>>;
+  using apply = _t<If<sizeof(try_<C>(0)) - 1 || sizeof...(Us), Empty, Result>>;
 };
 
 /**
--- a/folly/Poly.h
+++ b/folly/Poly.h
@@ -111,7 +111,7 @@ using PolySelf = decltype(Access::template self_<Node, Tfx>());
  */
 using PolyDecay = detail::MetaQuote<std::decay_t>;
 
-#if !defined(__cpp_template_auto)
+#if !FOLLY_POLY_NTTP_AUTO
 
 /**
  * Use `FOLLY_POLY_MEMBERS(MEMS...)` on pre-C++17 compilers to specify a
--- a/folly/Poly.h
+++ b/folly/Poly.h
@@ -50,6 +50,12 @@ namespace folly {
 template <class I>
 struct Poly;
 
+// MSVC workaround
+template <class Node, class Tfx, class Access>
+struct PolySelf_ {
+  using type = decltype(Access::template self_<Node, Tfx>());
+};
+
 /**
  * Within the definition of interface `I`, `PolySelf<Base>` is an alias for
  * the instance of `Poly` that is currently being instantiated. It is
@@ -101,7 +107,7 @@ template <
     class Node,
     class Tfx = detail::MetaIdentity,
     class Access = detail::PolyAccess>
-using PolySelf = decltype(Access::template self_<Node, Tfx>());
+using PolySelf = _t<PolySelf_<Node, Tfx, Access>>;
 
 /**
  * When used in conjunction with `PolySelf`, controls how to construct `Poly`
--- a/folly/detail/PolyDetail.h
+++ b/folly/detail/PolyDetail.h
@@ -695,14 +695,19 @@ struct BasePtr {
   VTable<I> const* vptr_;
 };
 
-template <class I, class T, std::enable_if_t<inSitu<T>(), int> = 0>
-constexpr void* (*getOps() noexcept)(Op, Data*, void*) {
+template <class I, class T>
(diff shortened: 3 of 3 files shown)


folly/detail/TypeList.h
```diff
@@ -168,7 +168,7 @@ using If = MetaApply<impl::If_<If_>, Then, Else>;
  */
 template <template <class...> class C, class... Ts>
 class MetaDefer {
-  template <template <class...> class D = C, class = D<Ts...>>
+  template <class = C<Ts...>>
```
yfeldblum: Why was this necessary?

cstratopoulos: I was developing with VS2019 Preview 16.2.0 Preview 3.0 and as originally written that line gives the error `error C3202: 'C': invalid default argument, expected a class template`

However I just checked and the original is fine with VS2019 Release 16.1.5 so I will report a bug to the VS team immediately.

I think it creates the same substitution failure written either way though?

cstratopoulos: Wait no sorry it's broken on VS2019 16.1.5 as well. Will report a bug either way as I think it should be valid as written.

yfeldblum: Seeing this under recent Clang:

```
folly/detail/TypeList.h:171:21: error: too many template arguments for alias template 'AddPointer'
  template <class = C<Ts...>>
                    ^ ~~~~~~
folly/detail/TypeList.h:90:1: note: in instantiation of template class 'folly::detail::MetaDefer<AddPointer, int, short>' requested here
using MetaApply = typename Fn::template apply<Ts...>;
^
folly/test/TypeListTest.cpp:27:49: note: in instantiation of template type alias 'MetaApply' requested here
struct IsApplicable_<T, TypeList<Ts...>, void_t<MetaApply<T, Ts...>>>
                                                ^
folly/test/TypeListTest.cpp:45:18: note: during template argument deduction for class template partial specialization 'IsApplicable_<T, TypeList<Ts...>, void_t<MetaApply<T, Ts...> > >' [with T = folly::detail::MetaDefer<AddPointer, int, short>, Ts = <>]
  static_assert(!IsApplicable<MetaDefer<AddPointer, int, short>>::value, "");
                 ^
folly/test/TypeListTest.cpp:45:18: note: in instantiation of template class '(anonymous namespace)::IsApplicable_<folly::detail::MetaDefer<AddPointer, int, short>, folly::detail::TypeList<>, void>' requested here
folly/test/TypeListTest.cpp:40:1: note: template is declared here
using AddPointer = T*;
^
folly/test/TypeListTest.cpp:40:1: error: 'AddPointer' declared as a pointer to a reference of type 'int &'
using AddPointer = T*;
^
folly/detail/TypeList.h:171:21: note: in instantiation of template type alias 'AddPointer' requested here
  template <class = C<Ts...>>
                    ^
folly/detail/TypeList.h:90:1: note: in instantiation of template class 'folly::detail::MetaDefer<AddPointer, int &>' requested here
using MetaApply = typename Fn::template apply<Ts...>;
^
folly/test/TypeListTest.cpp:27:49: note: in instantiation of template type alias 'MetaApply' requested here
struct IsApplicable_<T, TypeList<Ts...>, void_t<MetaApply<T, Ts...>>>
                                                ^
folly/test/TypeListTest.cpp:46:18: note: during template argument deduction for class template partial specialization 'IsApplicable_<T, TypeList<Ts...>, void_t<MetaApply<T, Ts...> > >' [with T = folly::detail::MetaDefer<AddPointer, int &>, Ts = <>]
  static_assert(!IsApplicable<MetaDefer<AddPointer, int&>>::value, "");
                 ^
folly/test/TypeListTest.cpp:46:18: note: in instantiation of template class '(anonymous namespace)::IsApplicable_<folly::detail::MetaDefer<AddPointer, int &>, folly::detail::TypeList<>, void>' requested here
```

cstratopoulos: Ack good catch. Was not aware of `TypeListTest.cpp` which I see is not added in the root `CMakeLists.txt`. I've added it in my local copy of the repo and tried a new fix. Seems to work locally with MSVC, and my clang-8 and gcc 8.1 installs as well.

folly/detail/TypeList.h
```diff
@@ -168,16 +168,18 @@ using If = MetaApply<impl::If_<If_>, Then, Else>;
  */
 template <template <class...> class C, class... Ts>
 class MetaDefer {
-  template <template <class...> class D = C, class = D<Ts...>>
+  template <template <class...> class D, class = D<Ts...>>
```
yfeldblum: Does this need to be `void_t<D<Ts...>>` (folly's `void_t`)?

cstratopoulos: Hmm I don't think it _needs_ to be: if `D<Ts...>` is ill-formed it will cause a substitution failure and overload resolution picks `try_(long)`, else they're both valid so `try_(int)` gets called? But maybe since I'm using `void` as the default parameter on `try_(long)` it would be a clarity improvement, so I'm happy to write it that way if you'd prefer.

cstratopoulos:
Re: https://github.com/facebook/folly/pull/1182/commits/31816aec7a5ba7431adbb11647c50dc448ff2ca6

In my local dev environment I realized I had still quarantined a handful of tests--ones involving `INullable` and `IEqualityComparable` and `IAddable`. MSVC was not accepting operator overloads where both parameters were `PolyBase<T>`  with [C2803](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2803?view=vs-2019)

Some attempted workarounds lead me to conclude the issue was just with `PolySelf` being defined in terms of a `decltype`; this level of indirection solves the issue and I think it is relatively inoffensive but I will file a compiler bug nonetheless.

folly/detail/TypeList.h
yfeldblum: Can all of the changes in this file be reverted given the addition of `PolySelf_`?

cstratopoulos: Unfortunately no, without it `TypeListTest.cpp` doesn't compile on MSVC.

PR #1184, opened 2019-07-03
Check if type is instantiation of a template

JoeLoser:
Summary:
- There is not an easy way to check if a given type is an instantiation
  of a class template.
- The common solution is to write a custom trait each time and
  specialize it for the given instantiation so that the trait returns
  `std:true_type`.
- Add `is_instantiation_of` to help with DRY in custom traits for
  checking if a type is an instance of a given class template. Note that
  this does not work when the class template has a mix of type and
  non-type template parameters. It only works with types.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -213,6 +213,24 @@ struct like_<Src&&> {
   template <typename Dst>
   using apply = typename like_<Src>::template apply<Dst>&&;
 };
+
+/**
+ * A type trait to check if a given type is an instantiation of a class
+ * template.
+ *
+ * Note that this only works with template type parameters.
+ * It does not work when mixing type and non-type template parameters.
+ * It also does not work with alias templates as you cannot pass an alias
+ * template as a template template parameter.
+ */
+template <template <typename...> class, typename>
+struct is_instantiation_of  : std::false_type {};
+
+template <template <typename...> class C, typename... Ts>
+struct is_instantiation_of<C, C<Ts...>> : std::true_type {};
+
+template <template <typename...> class C, typename T>
+constexpr bool is_instantiation_of_v = is_instantiation_of<C, T>::value;
 } // namespace detail
 
 //  mimic: like_t, p0847r0
--- a/folly/Replaceable.h
+++ b/folly/Replaceable.h
@@ -372,13 +372,8 @@ struct is_convertible_from_replaceable
           std::is_convertible<const Replaceable<T>&&, T>::value> {};
 } // namespace replaceable_detail
 
-// Type trait template to statically test whether a type is a specialization of
-// Replaceable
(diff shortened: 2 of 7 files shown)


folly/Traits.h
```diff
@@ -639,6 +639,15 @@ struct IsRelocatable<std::pair<T, U>>
 template <typename T, typename... Ts>
 using IsOneOf = StrictDisjunction<std::is_same<T, Ts>...>;
 
+template <typename, template <typename...> class>
+struct is_instantiation_of  : std::false_type {};
+
+template <typename... Ts, template <typename...> class C>
+struct is_instantiation_of<C<Ts...>, C> : std::true_type {};
```
yfeldblum: Looking at `std::is_base_of`, params are backwards.

JoeLoser: We can swap the order, no biggie. What do you think of the idea itself?

yfeldblum: Well, not the biggest fan because:
* It can't be sufficiently generic and work with non-type template parameters.
* It doesn't work through aliases: https://godbolt.org/z/nzkt-S.

But there have been repeat grumblings about not having it, so maybe it's worth it anyway.

Given the limitations, I would personally prefer it to be in `namespace folly::detail`.

JoeLoser: Right, I find that the ability to not work through aliases is a bigger deal than the mixing of types v.s. non-types. With the mixing, you can always just stay in type space and wrap the values when you have control over them.

In any case, I moved it to a detail namespace. I suspect this is a useful enough utility (despite its limitations we've pointed out) to have in the folly toolbox for other libs/apps to use that already depend on folly.

JoeLoser: @yfeldblum by the way, despite not being ideal, it shouldn't be surprising that this doesn't work with alias templates. This comes from the fact that you can't pass an alias template as a template template parameter.

folly/Traits.h
```diff
@@ -212,6 +212,15 @@ struct like_<Src&&> {
   template <typename Dst>
   using apply = typename like_<Src>::template apply<Dst>&&;
 };
+
+template <template <typename...> class, typename>
+struct is_instantiation_of  : std::false_type {};
```
aary: @JoeLoser Could you add some basic comments here? LGTM otherwise as a function in the detail namespace

JoeLoser: @aary sure, just added some comments describing the trait and its limitations. In the future, if it becomes useful from other parts of the codebase, we can look into adding `is_instantiation_of_value` (or some other name, key part is it can't just be overloaded with `is_instantiation_of` unfortunately) which is just like this except works with values (NTTP) and not types. We still don't have a mechanism to work with the mixing of types and NTTP though as you know.

PR #1192, opened 2019-07-20
Replace BOOST_MPL_HAS_XXX_TRAIT_DEF with Folly analogue

compmaniak:
Folly already has such macro in `folly/Traits.h` so `boost/mpl/has_xxx.hpp` is redundant


Orvid:
So it appears that the Folly version doesn't play nice with MSVC 2017 Update 5 (the minimum version we currently support). For the moment it would be better to stick to the Boost version.

PR #1203, opened 2019-08-12
Do not use undefined behavior of adding to std namespace

acoates-ms:
Adding to the std namespace is undefined behavior, and this code is causing us some issues when updating to use libcpp stl on android.

16.5.4.2.1 [namespace.std]\1
Unless otherwise specified, the behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std.

This version of the change removes the use of adding declarations to the std namespace altogether and just includes the required files.

I submitted another PR, #1204 that uses #ifdefs to achieve the same result, if its strongly desired that traits.h shouldn't include the stl headers.

### DIFF
--- a/folly/Traits.h
+++ b/folly/Traits.h
@@ -22,6 +22,12 @@
 #include <limits>
 #include <memory>
 #include <type_traits>
+#include <string>
+#include <vector>
+#include <set>
+#include <deque>
+#include <map>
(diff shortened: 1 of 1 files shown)


lbrandy:
I'd vote to remove the UB.

PR #1218, opened 2019-09-01
Fix tests on Arch Linux

MrAnno:
Compiling folly on Arch Linux triggered a few platform-specific behaviors that were not covered correctly by tests.

- Fix F14{Map,Set} maxSize test cases
    The maximum size of `F14Table` is dependent of its allocator's max_size:

    ```
    max_size() -> std::min<std::size_t>(
        (std::numeric_limits<InternalSizeType>::max)(),
        AllocTraits::max_size(a)
    )
    ```

    This commit fixes the `F14{Map,Set}` maxSize tests, so they won't fail
    on platforms where the standard allocator's `max_size()` is based on
    `ptrdiff_t` (signed).

-  Fix TestUtilTest
    If glog is compiled without gflags support, `SetCommandLineOptionWithMode()`
    can't force glog to use stderr.

    glog has its own compatibility layer for `FLAGS_*` in case gflags is missing,
    that's why `FLAGS_logtostderr` is always available.

- Skip AsyncIOTest tests if O_DIRECT is not supported by the file system
    `fs::temp_directory_path()` returns `"/tmp"` on Linux.

    When using systemd, the /tmp directory is mounted by default as tmpfs;
    and tmpfs does not support the `O_DIRECT` flag.

    This commit skips tests in case the file can't be opened with `O_DIRECT`.

### DIFF
--- a/folly/container/test/F14MapTest.cpp
+++ b/folly/container/test/F14MapTest.cpp
@@ -1365,14 +1365,16 @@ TEST(F14ValueMap, maxSize) {
   F14ValueMap<int, int> m;
   EXPECT_EQ(
       m.max_size(),
-      std::numeric_limits<std::size_t>::max() / sizeof(std::pair<int, int>));
+      std::allocator_traits<decltype(m)::allocator_type>::max_size(
+          m.get_allocator()));
 }
 
 TEST(F14NodeMap, maxSize) {
   F14NodeMap<int, int> m;
   EXPECT_EQ(
       m.max_size(),
-      std::numeric_limits<std::size_t>::max() / sizeof(std::pair<int, int>));
+      std::allocator_traits<decltype(m)::allocator_type>::max_size(
+          m.get_allocator()));
 }
 
 TEST(F14VectorMap, vectorMaxSize) {
@@ -1381,8 +1383,8 @@ TEST(F14VectorMap, vectorMaxSize) {
       m.max_size(),
       std::min(
           std::size_t{std::numeric_limits<uint32_t>::max()},
-          std::numeric_limits<std::size_t>::max() /
-              sizeof(std::pair<int, int>)));
+          std::allocator_traits<decltype(m)::allocator_type>::max_size(
+              m.get_allocator())));
 }
 
 template <typename M>
--- a/folly/experimental/io/test/AsyncIOTest.cpp
+++ b/folly/experimental/io/test/AsyncIOTest.cpp
@@ -34,6 +34,7 @@
 #include <folly/experimental/io/FsUtil.h>
 #include <folly/portability/GTest.h>
 #include <folly/portability/Sockets.h>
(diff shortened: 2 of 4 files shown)


folly/container/test/F14MapTest.cpp
```diff
@@ -1365,14 +1365,20 @@ TEST(F14ValueMap, maxSize) {
   F14ValueMap<int, int> m;
   EXPECT_EQ(
       m.max_size(),
-      std::numeric_limits<std::size_t>::max() / sizeof(std::pair<int, int>));
+      std::min(
```
yfeldblum: From @nbronson,

> We don't need `numeric_limits` here at all.

(Or the `min`.)

folly/experimental/io/test/AsyncIOTest.cpp
```diff
@@ -90,7 +90,7 @@ class TemporaryFile {
 };
 
 TemporaryFile::TemporaryFile(size_t size)
-    : path_(fs::temp_directory_path() / fs::unique_path()) {
+    : path_("/var/tmp" / fs::unique_path()) {
```
yfeldblum: From @nbronson,

> Rather than hard-coding paths here, maybe we should probe `fs::temp_directory_path()` with an `open` in `O_DIRECT` mode, and then just skip the test if that fails.

For skipping the test, use `SKIP` in `folly/test/TestUtils.h`.

MrAnno: This is much better, thanks!

folly/container/test/F14MapTest.cpp
MrAnno: Oh, I misunderstood a few things in `F14Table` and `F14BasicMap`.

If I understand now correctly, the `max_size()` of `F14ValueMap` is the min of the following 2 components:
- the maximum value that the type `F14Table::InternalSizeType` can represent
- the `max_size()` of `F14Table`'s allocator.

Since the allocator for a simple `F14ValueMap` is `DefaultAlloc`, which is `std::allocator<T>`,
and because `BasePolicy::InternalSizeType` is `size_t`, the correct assertion is:

```C++
EXPECT_EQ(
      m.max_size(),
      std::allocator_traits<decltype(m)::allocator_type>::max_size(m.get_allocator())
)
```

MrAnno: Should I be more explicit? For example:

```C++
  F14ValueMap<int, int> m;

  std::allocator<std::pair<int, int>> alloc;
  EXPECT_EQ(
      m.max_size(),
      std::allocator_traits<decltype(alloc)>::max_size(alloc));
```

yfeldblum: The first form is fine; don't need to be more explicit.

MrAnno: OK, thanks.

Issue #1208, opened 2019-08-28
Buffer overrun in CacheLocality::readFromProcCpuinfoLines


daverigby:
When running on certain CPU configurations, `CacheLocality::readFromProcCpuinfoLines` overruns the `indexes` vector. As reported by Valgrind:

```
 Invalid write of size 8
    at 0x4CFF5F: folly::CacheLocality::readFromProcCpuinfoLines(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) (CacheLocality.cpp:248)
    by 0x4D04DA: folly::CacheLocality::readFromProcCpuinfo() (CacheLocality.cpp:271)
    by 0x4D072C: folly::CacheLocality const& folly::CacheLocality::system<std::atomic>() (CacheLocality.cpp:39)
    by 0x4D26A4: folly::AccessSpreader<std::atomic>::initialize() (CacheLocality.h:359)
    by 0x412D19: __static_initialization_and_destruction_0 (CacheLocality.h:395)
    by 0x412D19: _GLOBAL__sub_I_CacheLocality.cpp (CacheLocality.cpp:353)
    by 0x51D38C: __libc_csu_init (in /root/mad-hatter/install/bin/eventing-consumer)
    by 0x8709B94: (below main) (libc-start.c:233)

  Address 0xa653010 is 0 bytes after a block of size 192 alloc'd
    at 0x4C2A203: operator new(unsigned long) (vg_replace_malloc.c:334)
    by 0x4CFF0B: allocate (new_allocator.h:111)
    by 0x4CFF0B: allocate (alloc_traits.h:436)
    by 0x4CFF0B: _M_allocate (stl_vector.h:172)
    by 0x4CFF0B: _M_create_storage (stl_vector.h:187)
    by 0x4CFF0B: _Vector_base (stl_vector.h:138)
    by 0x4CFF0B: vector (stl_vector.h:284)
    by 0x4CFF0B: folly::CacheLocality::readFromProcCpuinfoLines(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) (CacheLocality.cpp:246)
[...]

PR #1220, opened 2019-09-04
Support offline CPUs in CacheLocality::readFromProcCpuinfo

### LINKED ISSUE #1219: CacheLocality::system() reports incorrect CPU count if offline CPUs present
### Summary

If one or more CPU cores are offline, then `CacheLocality::system()` mis-reports the number of CPU cores.

### Steps to reproduce

On a system with 24 logical cores (as reported by `lscpu -e`):

1. Offline one CPU core (e.g. the 2nd):
```
chcpu --disable 1
```
2. Call `CacheLocality::system()` - e.g. via cache_locality_test
```
./cache_locality_test --gtest_filter=CacheLocality.LogSystem
```

Note that 24 cores are still reported:
```
[----------] 1 test from CacheLocality
[ RUN      ] CacheLocality.LogSystem
I0904 16:42:49.893321 37527 CacheLocalityTest.cpp:992] numCpus= 24
...
```

The same number is reported irrespective of how many CPU cores are still online - e.g. offlining half the CPU cores will still report "24".

Note the mis-counting has changed recently. Prior to the fix for #1208 (https://github.com/facebook/folly/commit/e545b9a9802107aa531e66d4996da7c19f53402c) it _would_ report the correct CPU count, however it suffered from the `std::vector` overrun reported in #1208, depending on which CPU cores were offline (if a CPU core number `C` was online such that `C` > `num_online_cpus` then you'd buffer overflow.

daverigby:
Ensure that CacheLocality::readFromProcCpuinfo() sizes the cache
locality `indexes` vector based on the maximum core number, correclty
handling instances where one or more CPU cores are offline.

Fixes #1219.

### DIFF
--- a/folly/concurrency/CacheLocality.cpp
+++ b/folly/concurrency/CacheLocality.cpp
@@ -228,10 +228,6 @@ CacheLocality CacheLocality::readFromProcCpuinfoLines(
   if (cpus.empty()) {
     throw std::runtime_error("no CPUs parsed from /proc/cpuinfo");
   }
-  if (maxCpu != cpus.size() - 1) {
-    throw std::runtime_error(
-        "offline CPUs not supported for /proc/cpuinfo cache locality source");
-  }
 
   std::sort(cpus.begin(), cpus.end());
   size_t cpusPerCore = 1;
@@ -249,7 +245,7 @@ CacheLocality CacheLocality::readFromProcCpuinfoLines(
   numCachesByLevel.push_back(cpus.size() / cpusPerCore);
   numCachesByLevel.push_back(std::get<0>(cpus.back()) + 1);
 
-  std::vector<size_t> indexes(cpus.size());
+  std::vector<size_t> indexes(maxCpu + 1, std::numeric_limits<size_t>::max());
   for (size_t i = 0; i < cpus.size(); ++i) {
     indexes[std::get<2>(cpus[i])] = i;
   }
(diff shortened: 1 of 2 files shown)


daverigby:
> Is this still an issue after [e545b9a](https://github.com/facebook/folly/commit/e545b9a9802107aa531e66d4996da7c19f53402c)?

Indeed it is :(

The basic problem is thus - neither the procCpuInfo nor procSysFs currently (in master) correctly handle offline CPUs, however the consequences of each strategy is different:

- ProcCpuInfo currently (e545b9a) checks if any CPUs are offline, and if so falls back to the slower ProcSysFs - that "fixed" issue #1208 by avoiding the buffer overrun in ProcCpuInfo code, however it ends up calling the ProcSysFs impl...
- ProcSysFs doesn't check for CPUs being online - it assumes that failing to read from a CPU cache file means it's reached the end of the available CPUs, and stops searching. This results incorrect CPU counting - it'll report only as many CPUs existing as it found before the first offline CPU - so if CPU 1 was offline, it'll report only one CPU (#0) existing.

yfeldblum:
What is the impact of miscounting?

daverigby:
> For ourselves it means the performance of AccessSpreader is abysmal if CPUs are offline - as AccessSpreader incorrectly ends up with just 1 shard instead of NUM_CPU shards, and hence all CPU cores end up hitting a single shard.

To clarify - it isn't _always_ just one shard created, it depends on which particular CPU cores are offline, but in the worst case if CPU core #1 is offline (even if there's another 31 online) then `CacheLocality` reports just 1 CPU present.

yfeldblum:
Can you add some unit-tests which simulate offline cpu's?

daverigby:
Stale - closing.

PR #1229, opened 2019-09-18
 reverse iterator support for concurrent skip list

muralivemulapati:


muralivemulapati:
This is the reopened PR for 1161.

nbronson:
This diff needs better unit tests. In particular, there doesn't seem to be any coverage of concurrent erase.

yfeldblum:
Additionally, we're trying to deprecate `ConcurrentSkipList` from folly. We would want to avoid encouraging new dependencies by adding support for extra use cases and to avoid implying that this code is more supported than it actually is.

PR #1242, opened 2019-10-12
add cmake build shared libs support

pxps:
add cmake shared libs build option support:
  DCMAKE_BUILD_SHARED_LIBS=on


pxps:
Why are there so many tests that have not passed?

PR #1246, opened 2019-10-22
clang-tidy fixes

neheb:


folly/Conv.cpp
```diff
@@ -276,7 +276,7 @@ Expected<bool, ConversionCode> str_to_bool(StringPiece* src) noexcept {
   }
 
   bool result;
-  size_t len = size_t(e - b);
+  auto len = size_t(e - b);
```
yfeldblum: Why is this linted?

neheb:
clang-tidy goofed up on the last one. force-pushed a new version.

neheb:
On second thought, I'll split these up.

PR #1256, opened 2019-11-10
[clang-tidy] Use auto where appropriate

neheb:
Found with modernize-use-auto

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/folly/IPAddress.cpp
+++ b/folly/IPAddress.cpp
@@ -246,12 +246,12 @@ IPAddress::IPAddress(const sockaddr* addr) : addr_(), family_(AF_UNSPEC) {
   family_ = addr->sa_family;
   switch (addr->sa_family) {
     case AF_INET: {
-      const sockaddr_in* v4addr = reinterpret_cast<const sockaddr_in*>(addr);
+      auto v4addr = reinterpret_cast<const sockaddr_in*>(addr);
       addr_.ipV4Addr = IPAddressV4(v4addr->sin_addr);
       break;
     }
     case AF_INET6: {
-      const sockaddr_in6* v6addr = reinterpret_cast<const sockaddr_in6*>(addr);
+      auto v6addr = reinterpret_cast<const sockaddr_in6*>(addr);
       addr_.ipV6Addr = IPAddressV6(*v6addr);
       break;
     }
--- a/folly/compression/Compression.cpp
+++ b/folly/compression/Compression.cpp
@@ -2097,7 +2097,7 @@ constexpr Factory
 };
 
 Factory const& getFactory(CodecType type) {
-  size_t const idx = static_cast<size_t>(type);
+  auto const idx = static_cast<size_t>(type);
   if (idx >= static_cast<size_t>(CodecType::NUM_CODEC_TYPES)) {
     throw std::invalid_argument(
         to<std::string>("Compression type ", idx, " invalid"));
--- a/folly/experimental/crypto/detail/MathOperation_SSE2.cpp
+++ b/folly/experimental/crypto/detail/MathOperation_SSE2.cpp
@@ -69,8 +69,8 @@ void MathOperation<MathEngine::SSE2>::add(
   // the Endian::little() conversions when loading or storing data.
   if (bitsPerElement == 16 || bitsPerElement == 32) {
     for (size_t pos = 0; pos < b1.size(); pos += kCacheLineSize) {
-      const __m128i* v1p = reinterpret_cast<const __m128i*>(b1.data() + pos);
-      const __m128i* v2p = reinterpret_cast<const __m128i*>(b2.data() + pos);
+      auto v1p = reinterpret_cast<const __m128i*>(b1.data() + pos);
+      auto v2p = reinterpret_cast<const __m128i*>(b2.data() + pos);
       for (size_t i = 0; i < kValsPerCacheLine; ++i) {
         __m128i v1 = _mm_load_si128(v1p + i);
         __m128i v2 = _mm_load_si128(v2p + i);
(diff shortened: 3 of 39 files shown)


folly/IPAddress.cpp
```diff
@@ -246,12 +246,12 @@ IPAddress::IPAddress(const sockaddr* addr) : addr_(), family_(AF_UNSPEC) {
   family_ = addr->sa_family;
   switch (addr->sa_family) {
     case AF_INET: {
-      const sockaddr_in* v4addr = reinterpret_cast<const sockaddr_in*>(addr);
+      const auto v4addr = reinterpret_cast<const sockaddr_in*>(addr);
```
yfeldblum: This looks like it's not just a use-`auto`-where-appropriate transformation. If `v4addr` were subsequently reassigned to point to a different object, then the original code would compile but the code after clang-tidy transformation would fail to compile. The pedantically-correct transformation would be from `const sockaddr_in*` to `auto`, not to `const auto`. Is this a bug in clang-tidy?

There can certainly be a *separate* code transformation which `const`-qualifies local variables which are not subsequently modified, but the two code transformations seem like they should be independent.

neheb: You're right. const auto evaluates to const (const sockaddr_in). Will fix.

neheb: I think const auto evaluates to const (const sockaddr_in*) while const auto* evaluates to const sockaddr_in* . Seems to be a subtlety in C++.

folly/compression/Compression.cpp
```diff
@@ -2097,7 +2097,7 @@ constexpr Factory
 };
 
 Factory const& getFactory(CodecType type) {
-  size_t const idx = static_cast<size_t>(type);
+  const auto idx = static_cast<size_t>(type);
```
yfeldblum: Let's not revert anything from east-const back to west-const.

folly/experimental/crypto/detail/MathOperation_SSE2.cpp
```diff
@@ -69,8 +69,8 @@ void MathOperation<MathEngine::SSE2>::add(
   // the Endian::little() conversions when loading or storing data.
   if (bitsPerElement == 16 || bitsPerElement == 32) {
     for (size_t pos = 0; pos < b1.size(); pos += kCacheLineSize) {
-      const __m128i* v1p = reinterpret_cast<const __m128i*>(b1.data() + pos);
-      const __m128i* v2p = reinterpret_cast<const __m128i*>(b2.data() + pos);
+      const auto v1p = reinterpret_cast<const __m128i*>(b1.data() + pos);
```
yfeldblum: The same unintended type change happens here and below.

folly/io/IOBuf.cpp
```diff
@@ -136,9 +136,9 @@ void IOBuf::SharedInfo::invokeAndDeleteEachObserver(
   if (observerListHead && cb) {
     // break the chain
     observerListHead->prev->next = nullptr;
-    auto* entry = observerListHead;
+    auto entry = observerListHead;
```
yfeldblum: This change doesn't seem quite as appropriate. In the other changes, the type is already named on the RHS so it doesn't need to be copied to the LHS. Here, the type is not named on the RHS. If an author wants to be a little bit clearer that the type here is a pointer, I'm not sure that needs to be changed.

folly/io/async/AsyncServerSocket.cpp
```diff
@@ -212,7 +212,7 @@ int AsyncServerSocket::stopAccepting(int shutdownFlags) {
   // removeAcceptCallback().
   std::vector<CallbackInfo> callbacksCopy;
   callbacks_.swap(callbacksCopy);
-  for (std::vector<CallbackInfo>::iterator it = callbacksCopy.begin();
+  for (auto it = callbacksCopy.begin();
```
yfeldblum: This change is fine for this diff, since naming the type explicitly here likely comes from a time before `auto` and is likely not the intention. But this loop could, in a separate PR, be changed to a range-for.

folly/io/async/AsyncSocket.cpp
```diff
@@ -981,7 +981,7 @@ bool AsyncSocket::isZeroCopyMsg(const cmsghdr& cmsg) const {
 #ifdef FOLLY_HAVE_MSG_ERRQUEUE
   if ((cmsg.cmsg_level == SOL_IP && cmsg.cmsg_type == IP_RECVERR) ||
       (cmsg.cmsg_level == SOL_IPV6 && cmsg.cmsg_type == IPV6_RECVERR)) {
-    const struct sock_extended_err* serr =
+    const auto serr =
```
yfeldblum: Same.

folly/io/async/ssl/OpenSSLUtils.cpp
```diff
@@ -129,7 +129,7 @@ bool OpenSSLUtils::validatePeerCertNames(
     if ((addr4 != nullptr || addr6 != nullptr) && name->type == GEN_IPADD) {
       // Extra const-ness for paranoia
       unsigned char const* const rawIpStr = name->d.iPAddress->data;
-      size_t const rawIpLen = size_t(name->d.iPAddress->length);
+      const auto rawIpLen = size_t(name->d.iPAddress->length);
```
yfeldblum: Same.

folly/io/async/AsyncServerSocket.cpp
neheb: I believe that is actually done. But I'm splitting it up by parts for easier review.

folly/io/IOBuf.cpp
neheb: Right. It's more consistent with the other changes though.

neheb:
I believe all the issues have been addressed.

yfeldblum:
Yep, thanks!

PR #1281, opened 2019-12-15
FreeBSD build fix proposal.

devnexen:
- __cxa_get_globals signature closer of the libcxxabi.
- malloc.h exists now on FreeBSD but prevents stdlib.h inclusion if __STDC__.

### DIFF
--- a/folly/lang/UncaughtExceptions.h
+++ b/folly/lang/UncaughtExceptions.h
@@ -28,7 +28,7 @@ struct __cxa_eh_globals;
 extern "C" __cxa_eh_globals* __cxa_get_globals() noexcept;
 #else
 // Signature mismatch with FreeBSD case
-extern "C" __cxa_eh_globals* __cxa_get_globals();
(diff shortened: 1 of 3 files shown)


folly/CPortability.h
```diff
@@ -240,3 +240,11 @@
 #else
 #define FOLLY_ERASE_TRYCATCH FOLLY_ERASE
 #endif
+
+//  In the recent FreeBSD releases, malloc.h header does exist
+//  however when __STDC__ is defined it throws an exception
+#ifdef __FreeBSD__
+#ifdef __STDC__
+#undef __STDC__
```
yfeldblum: What defines `__STDC__`?

devnexen: I wanted to avoid forcing specific compiler flags.

yfeldblum: Overall in folly, we should stick to defining/undefining only names that are namespaced to folly.

PR #1286, opened 2019-12-21
[clang-tidy] Use uppercase literals suffixes

neheb:
Found with hicpp-uppercase-literal-suffix

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/folly/GroupVarint.cpp
+++ b/folly/GroupVarint.cpp
@@ -43,7 +43,7 @@ namespace detail {
 
 struct group_varint_table_base_make_item {
   constexpr std::size_t get_d(std::size_t index, std::size_t j) const {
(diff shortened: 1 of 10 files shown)


yfeldblum:
What's the problem with the lowercase suffixes?

neheb:
It's a style issue mostly.

The given example is differentiating 1 vs l. See: https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152241

yfeldblum:
This PR includes only changes from `u` to `U`, which seems not to be a problem. The `u` seems generally useful with `auto` because of size-conversion warnings, while `l` or `ll` by themselves seem not useful in general because widening is typically not a problem.

neheb:
Hrm fair enough.

PR #1298, opened 2020-01-02
[clang-tidy] std::move improvements

neheb:


folly/compression/Compression.cpp
```diff
@@ -70,7 +71,7 @@ namespace io {
 
 Codec::Codec(
     CodecType type,
-    Optional<int> level,
+    const Optional<int>& level,
```
yfeldblum: Passing `Optional<T>` by value should be fine when passing `T` by value is fine. Here and elsewhere.

folly/experimental/io/AsyncBase.cpp
```diff
@@ -175,7 +175,7 @@ void AsyncBaseQueue::submit(AsyncBaseOp* op) {
   submit([op]() { return op; });
 }
 
-void AsyncBaseQueue::submit(OpFactory op) {
+void AsyncBaseQueue::submit(const OpFactory& op) {
   queue_.push_back(op);
```
yfeldblum: Perhaps this should be a move instead.

folly/executors/ThreadPoolExecutor.cpp
```diff
@@ -356,7 +356,7 @@ size_t ThreadPoolExecutor::StoppedThreadQueue::size() {
   return queue_.size();
 }
 
-void ThreadPoolExecutor::addObserver(std::shared_ptr<Observer> o) {
+void ThreadPoolExecutor::addObserver(const std::shared_ptr<Observer>& o) {
```
yfeldblum: Perhaps `o` should be moved below instead.

folly/json_pointer.cpp
```diff
@@ -49,7 +49,7 @@ Expected<json_pointer, json_pointer::parse_error> json_pointer::try_parse(
 json_pointer json_pointer::parse(StringPiece const str) {
   auto res = try_parse(str);
   if (res.hasValue()) {
-    return std::move(res.value());
+    return res.value();
```
yfeldblum: Presumably we want to avoid copying `vector<string>`.

folly/logging/LogCategoryConfig.cpp
```diff
@@ -24,7 +24,7 @@ LogCategoryConfig::LogCategoryConfig(LogLevel l, bool inherit)
 LogCategoryConfig::LogCategoryConfig(
     LogLevel l,
     bool inherit,
-    std::vector<std::string> h)
+    const std::vector<std::string>& h)
```
yfeldblum: Presumably `h` should be moved instead.

folly/system/MemoryMapping.cpp
```diff
@@ -245,7 +245,7 @@ off_t memOpChunkSize(off_t length, off_t pageSize) {
  * - failure: false + amountSucceeded == nr bytes on which op succeeded.
  */
 bool memOpInChunks(
-    std::function<int(void*, size_t)> op,
+    const std::function<int(void*, size_t)>& op,
```
yfeldblum: Presumably this should be templated by the op instead or should take `FunctionRef` instead. (Both by value.)

neheb:
Replaced second commit with a different one based on feedback.

folly/json_pointer.cpp
yfeldblum: Is this from clang-tidy or is this manual?

yfeldblum:
Note that our tooling for working with PRs actually squashes all the commits in a PR into a single commit. If you want two separate commits, can you use two separate PRs?

folly/json_pointer.cpp
neheb: clang-tidy. Maybe because of the earlier std::move.

neheb:
OK

folly/json_pointer.cpp
yfeldblum: This seems wrong though. `res.value()` is an lvalue-ref and so this code without `std::move` logically invokes the copy-constructor of `json_pointer`, which contains a `vector<string>`. The goal of `std::move` is to invoke the move-constructor instead.

neheb: Reverted since this is cosmetic anyway.

PR #1312, opened 2020-01-22
Handle multiple async pauses which maybe returned by crypto engine

nidhidamodaran:
At present, after handling fist async pause handler is being reset, as such to handle cases where crypto engines may return multiple pauses, delaying resetting callback until SSL_accept completes.

### DIFF
--- a/folly/io/async/AsyncPipe.cpp
+++ b/folly/io/async/AsyncPipe.cpp
@@ -73,7 +73,7 @@ void AsyncPipeReader::handlerReady(uint16_t events) noexcept {
   VLOG(5) << "AsyncPipeReader::handlerReady() this=" << this << ", fd=" << fd_;
   assert(readCallback_ != nullptr);
 
-  while (readCallback_) {
+  if (readCallback_) {
     // - What API does callback support?
     const auto movable = readCallback_->isBufferMovable(); // noexcept
 
--- a/folly/io/async/AsyncSSLSocket.cpp
+++ b/folly/io/async/AsyncSSLSocket.cpp
@@ -1065,17 +1065,17 @@ bool AsyncSSLSocket::willBlock(
       // Our NetworkSocket::native_handle_type is type SOCKET on
       // win32, which means that we need to explicitly construct
       // a native handle type to pass to the constructor.
-      auto native_handle = NetworkSocket::native_handle_type(ofd);
-
-      auto asyncPipeReader =
-          AsyncPipeReader::newReader(eventBase_, NetworkSocket(native_handle));
-      auto asyncPipeReaderPtr = asyncPipeReader.get();
       if (!asyncOperationFinishCallback_) {
-        asyncOperationFinishCallback_.reset(
-            new DefaultOpenSSLAsyncFinishCallback(
-                std::move(asyncPipeReader), this, DestructorGuard(this)));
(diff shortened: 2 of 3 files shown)


folly/io/async/AsyncPipe.cpp
```diff
@@ -73,7 +73,7 @@ void AsyncPipeReader::handlerReady(uint16_t events) noexcept {
   VLOG(5) << "AsyncPipeReader::handlerReady() this=" << this << ", fd=" << fd_;
   assert(readCallback_ != nullptr);
 
-  while (readCallback_) {
+  if (readCallback_) {
```
afrind: I'm not sure this is the right change.  The code below has this comment

```
      // Fall through and continue around the loop if the read
      // completely filled the available buffer.
```

On the flip side, there also seems to be the possibility of starvation of other events, which could be prevented by handling this seeming TODO item: 

```
    // Max reads per loop?
```

eg: it should be something like

```
  while (readCallback_ && nReads++ < kMaxReadsPerLoop)
```

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1065,17 +1065,17 @@ bool AsyncSSLSocket::willBlock(
...
+
+          auto asyncPipeReader =
+              AsyncPipeReader::newReader(eventBase_, NetworkSocket(native_handle));
+          auto asyncPipeReaderPtr = asyncPipeReader.get();
+          asyncOperationFinishCallback_.reset(
+                  new DefaultOpenSSLAsyncFinishCallback(
+                      std::move(asyncPipeReader), this, DestructorGuard(this)));
+          asyncPipeReaderPtr->setReadCB(asyncOperationFinishCallback_.get());
       }
-      asyncPipeReaderPtr->setReadCB(asyncOperationFinishCallback_.get());
```
afrind: The old code would have cleared this callback if the member variable was unset.  Was that important behavior?

folly/io/async/AsyncPipe.cpp
nidhidamodaran: Thanks @afrind , will check that and update this PR

nidhidamodaran: @afrind , In here since call to readDataAvailable is resetting the readCallback_, is it necessary to loop in here, also since we are using the data in the pipe only as a signalling mechanism, do we need to handle the case where there maybe more data than the allocated buffer ?

If I loop in here, SSL_accept will be called multiple times without receiving any more events on the wait fd

afrind: There are other use cases for AsyncPipe, so the code here should be generic (eg: reading large amounts of data quickly).  If there is no additional data in the pipe, it's unclear to me how looping will invoke the callback again (will return in like 143), and if there is additional data in the pipe, isn't it correct to invoke the callback?

nidhidamodaran: @afrind ,  Few questions  in here
1.) Since just after restartSSLAccept, read callback is being set to NULL, is it required to do a while loop in here, as at that point readCallback is definitely NULL.

2.) Another issue I have is with destructor method of Pipe. Folly is closing the async fd which is received from external crypto library. In my case this is handle of my crypto driver and should not be closed by folly. Can you let me know if there are any implications if it is removed ?

PR #1317, opened 2020-02-12
FreeBSD and/or clang build fix.

devnexen:
clang being more strict with obsolete C++14 api, has
completely disabled them in this context.

### DIFF
--- a/CMake/folly-config.h.cmake
+++ b/CMake/folly-config.h.cmake
@@ -26,16 +26,6 @@
 #endif
 #endif // FOLLY_MOBILE
 
-#if defined(__linux__)
-#define FOLLY_ELF_NATIVE_CLASS __ELF_NATIVE_CLASS
-#elif defined(__FreeBSD__)
-#if defined(__LP64__)
-#define FOLLY_ELF_NATIVE_CLASS 64
-#else
-#define FOLLY_ELF_NATIVE_CLASS 32
-#endif
-#endif
-
 #cmakedefine FOLLY_HAVE_PTHREAD 1
 #cmakedefine FOLLY_HAVE_PTHREAD_ATFORK 1
 
--- a/CMake/folly-config.h.cmake
+++ b/CMake/folly-config.h.cmake
@@ -26,6 +26,16 @@
 #endif
 #endif // FOLLY_MOBILE
 
+#if defined(__linux__)
+#define FOLLY_ELF_NATIVE_CLASS __ELF_NATIVE_CLASS
+#elif defined(__FreeBSD__)
+#if defined(__LP64__)
+#define FOLLY_ELF_NATIVE_CLASS 64
+#else
+#define FOLLY_ELF_NATIVE_CLASS 32
+#endif
+#endif
+
 #cmakedefine FOLLY_HAVE_PTHREAD 1
 #cmakedefine FOLLY_HAVE_PTHREAD_ATFORK 1
 
--- a/folly/experimental/exception_tracer/ExceptionTracer.cpp
+++ b/folly/experimental/exception_tracer/ExceptionTracer.cpp
@@ -201,7 +201,7 @@ std::vector<ExceptionInfo> getCurrentExceptions() {
   return exceptions;
 }
 
-#if !defined(__clang__)
+#if FOLLY_USE_LIBSTDCPP
 namespace {
 
 std::terminate_handler origTerminate = abort;
--- a/folly/experimental/exception_tracer/ExceptionTracer.cpp
+++ b/folly/experimental/exception_tracer/ExceptionTracer.cpp
@@ -18,6 +18,7 @@
 
 #include <exception>
 #include <iostream>
+#include <cstdlib>
 
 #include <dlfcn.h>
 
@@ -200,6 +201,7 @@ std::vector<ExceptionInfo> getCurrentExceptions() {
   return exceptions;
 }
 
+#if !defined(__clang__)
 namespace {
 
 std::terminate_handler origTerminate = abort;
@@ -238,6 +240,7 @@ void installHandlers() {
   };
   static Once once;
 }
+#endif
 
 } // namespace exception_tracer
 } // namespace folly
--- a/folly/experimental/symbolizer/Elf.h
+++ b/folly/experimental/symbolizer/Elf.h
@@ -35,25 +35,20 @@
 namespace folly {
 namespace symbolizer {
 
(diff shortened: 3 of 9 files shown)


folly/experimental/exception_tracer/ExceptionTracer.cpp
```diff
@@ -200,6 +201,7 @@ std::vector<ExceptionInfo> getCurrentExceptions() {
   return exceptions;
 }
 
+#if !defined(__clang__)
```
yfeldblum: Why doesn't this work with clang?

devnexen: as mentioned in the comment, there is obsolete C++ 14 usage (e.g. std::unexpected_handler)

yfeldblum: Do you mean libc++ instead of clang?

devnexen: yes good point forgot FreeBSD clang uses libc++ whatsoever.

folly/experimental/symbolizer/Elf.h
```diff
@@ -35,12 +35,26 @@
 namespace folly {
 namespace symbolizer {
 
+#if !defined(__FreeBSD__)
```
yfeldblum: This way of organizing the conditions is awkward and only works with a single if/else choice. What if we need to support another OS?

```
#if defined(__FreeBSD__)
...
#else
...
#endif
```

folly/experimental/symbolizer/Elf.h
```diff
@@ -35,12 +35,26 @@
...
 using ElfSym = ElfW(Sym);
+#else
+using ElfAddr = Elf_Addr;
+using ElfEhdr = Elf_Ehdr;
+using ElfOff = Elf_Off;
+using ElfPhdr = Elf_Phdr;
+using ElfShdr = Elf_Shdr;
+using ElfSym = Elf_Sym;
+#if !defined(__LP64__)
+#define __ELF_NATIVE_CLASS 32
```
yfeldblum: Let's instead wrap this in symbol which is namespaced to folly, and then only use that namespaced symbol throughout folly. Rather than defining symbols into the global namespace.

folly/experimental/symbolizer/SignalHandler.cpp
```diff
@@ -361,8 +361,15 @@ void dumpSignalInfo(int signum, siginfo_t* siginfo) {
   printDec(getpid());
   print(" (pthread TID ");
   printHex((uint64_t)pthread_self());
+#if !defined(__FreeBSD__)
```
yfeldblum: Same comment re condition ordering.

CMake/folly-config.h.cmake
```diff
@@ -26,6 +26,16 @@
...
+#if defined(__linux__)
+#define FOLLY_ELF_NATIVE_CLASS __ELF_NATIVE_CLASS
+#elif defined(__FreeBSD__)
+#if defined(__LP64__)
+#define FOLLY_ELF_NATIVE_CLASS 64
+#else
+#define FOLLY_ELF_NATIVE_CLASS 32
+#endif
+#endif
+
```
yfeldblum: Let's move all of this into `folly/Portability.h`.

folly/experimental/symbolizer/Symbolizer.cpp
```diff
@@ -189,7 +189,7 @@ void Symbolizer::symbolize(
 
       // Get the unrelocated, ELF-relative address by normalizing via the
       // address at which the object is loaded.
-      auto const adjusted = addr - lmap->l_addr;
+      auto const adjusted = addr - (uintptr_t)lmap->l_addr;
```
yfeldblum: Let's use C++ casts in C++.
```suggestion
      auto const adjusted = addr - static_cast<uintptr_t>(lmap->l_addr);
```

folly/experimental/exception_tracer/ExceptionTracer.cpp
yfeldblum: Check for libstdc++ instead of !clang?

folly/experimental/symbolizer/Symbolizer.cpp
```diff
@@ -374,7 +374,7 @@ void SymbolizePrinter::println(
 namespace {
 
 int getFD(const std::ios& stream) {
-#if defined(__GNUC__) && FOLLY_HAS_RTTI
+#if (defined(__GNUC__) && !defined(__clang__)) && FOLLY_HAS_RTTI
```
yfeldblum: Check for libstdc++ instead of !clang?

folly/experimental/symbolizer/Symbolizer.cpp
devnexen: ah nope this is the old fashion caddr_t type we re dealing with here :-) reinterpret_cast works in this case.

folly/experimental/symbolizer/SignalHandler.cpp
```diff
@@ -361,8 +361,15 @@ void dumpSignalInfo(int signum, siginfo_t* siginfo) {
...
   print(") (linux TID ");
   printDec(syscall(__NR_gettid));
+#else
+  long tid;
+  thr_self(&tid);
```
luciang: The functions called here need to be async-signal-safe.

I can't find info saying this is. It seems to be multi-thread-safe, but not async-signal-safe.

According to https://illumos.org/man/3C/thr_self
```

ATTRIBUTES

       See attributes(5) for descriptions of the following attributes:


       +---------------+-----------------+
       |ATTRIBUTE TYPE | ATTRIBUTE VALUE |
       +---------------+-----------------+
       |MT-Level       | MT-Safe         |
       +---------------+-----------------+
```

Is there any async-signal-safe way to implement this (eg. call a system call directly).

folly/experimental/symbolizer/SignalHandler.cpp
```diff
@@ -366,7 +366,7 @@ void dumpSignalInfo(int signum, siginfo_t* siginfo) {
   printDec(syscall(__NR_gettid));
 #elif defined(__FreeBSD__)
   long tid;
-  thr_self(&tid);
+  syscall(432, &tid);
```
yfeldblum: Is there a symbolic constant for this?

devnexen: nope

PR #1321, opened 2020-02-21
Make calling convention explicit.

JunielKatarn:
This allows for projects that depend on folly built with different calling conventions to share Folly headers and  be binary-compatible.

By default, compilers use `__cdecl` as the calling convention on `x86` platforms.
This change ensures changing that compiler flag won't make binaries incompatible with other consuming software code bases or build configurations.


WillerZ:
I don't think this is a use-case we can really support 

Folly doesn't support 32-bit Windows (except by coincidence). Internally it's not a platform we use, test on, or even build for.

More generally folly is not intended to be at all ABI stable - the expectation is that clients will statically link to a version of folly that they have validated works for them.

PR #1325, opened 2020-02-27
Make GFlags actually optional

daverigby:
The absence of GFlags is already partially optional -
portability/GFlags.h defines a number of stub macros / clases if
FOLLY_HAVE_LIBGFLAGS is false.

However the top-level CMake rules don't correctly handle when GFlags
isn't present - they unconditionally add GFlags-related variables to
CMAKE_REQUIRED_xxx variables.

Additionally there are a couple of source files which should not be
compiled if GFlags isn't available.


CMakeLists.txt
```diff
@@ -272,6 +272,18 @@ else()
     ${FOLLY_DIR}/experimental/crypto/LtHash.h
   )
 endif()
+if (NOT ${LIBGFLAGS_FOUND})
+  list(REMOVE_ITEM files
+    ${FOLLY_DIR}/experimental/NestedCommandLineApp.cpp
+    ${FOLLY_DIR}/experimental/ProgramOptions.cpp
+    ${FOLLY_DIR}/init/Init.cpp
```
yfeldblum: The intention is that `folly/init/Init.cpp` works whether or not gflags is available.

Perhaps there is something which `folly/init/Init.cpp` uses which is not yet provided by `folly/portability/GFlags.h`.

daverigby: Thanks for the feedback. I'll fix to just skip the parts of Init.cpp which rely on GFlags.

folly/portability/GFlags.h
```diff
@@ -62,9 +64,11 @@
 #define DEFINE_string(_name, _default, _description) \
   FOLLY_DEFINE_FLAG(std::string, S, _name, _default)
 
-namespace google {
+namespace gflags {
 class FlagSaver {};
-} // namespace google
+
+using int32 = int32_t;
```
yfeldblum: We should now be able to undo adding this alias after 581e3369a3e44ba7ed4b33114ced5ff6bf5ff7d9.

daverigby: Roger. Pull request updated.

PR #1334, opened 2020-03-10
include <bit> if we're going to use std::bit_cast

### LINKED ISSUE #1333: Folly failed to build due to ’error C2039: 'bit_cast': is not a member of 'std'‘ on MSVC
Hi All,

Folly failed to build due to error C2039 on MSVC. Could you please help take a look at this? Thanks in advance!

**Environment**
VCPKG(latest revision from github) + VS 2019 + Windows Server 2016

**Repro steps:**

1. open X64 native tools command prompt for VS2019
2. git clone https://github.com/Microsoft/vcpkg D:\VCPKG\src
3. cd d:\vcpkg\src
4. devenv /upgrade .\toolsrc\vcpkg.sln
5. set CL=/D_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
6. msbuild /p:Configuration=Release /p:Platform=x86 /p:WindowsTargetPlatformVersion 10.0.18362.0 toolsrc\vcpkg.sln /t:Rebuild
7. set CL=
8. set path=%cd%\toolsrc\Release;%path%
9. vcpkg install folly:x64-windows –head

**Actual result:**
F:\gitP\microsoft\vcpkg\buildtrees\folly\src\head\master-71ebdf935b\folly/lang/Bits.h(71): error C2039: 'bit_cast': is not a member of 'std'
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\cinttypes(22): note: see declaration of 'std'
F:\gitP\microsoft\vcpkg\buildtrees\folly\src\head\master-71ebdf935b\folly/lang/Bits.h(71): error C2873: 'bit_cast': symbol cannot be used in a using-declaration

barcharcraz:
This fixes #1333 and makes folly work with the most recent version of microsoft/STL (we just added bit_cast)

Note: I've not tested this locally so I am relying on CI to make sure things work.

### DIFF
--- a/folly/lang/Bits.h
+++ b/folly/lang/Bits.h
@@ -64,11 +64,16 @@
 #include <folly/lang/Assume.h>
 #include <folly/portability/Builtins.h>
 
+#ifdef __has_include
+#if __has_include(<bit>)
+#include <bit>
+#endif
+#endif
+
 namespace folly {
 
 #if __cpp_lib_bit_cast
 
-#include <bit>
 using std::bit_cast;
 
 #else
--- a/folly/lang/Bits.h
+++ b/folly/lang/Bits.h
@@ -68,6 +68,7 @@ namespace folly {
 
 #if __cpp_lib_bit_cast
 
+#include <bit>
 using std::bit_cast;
 
 #else


folly/lang/Bits.h
```diff
@@ -68,6 +68,7 @@ namespace folly {
 
 #if __cpp_lib_bit_cast
 
+#include <bit>
```
CaseyCarter: This will work for MSVC - we define all feature-test macros when any header is included - but it's not portable. For a portable fix the code needs to include `<bit>` or `<version>` before checking `__cpp_lib_bit_cast`. I suggest:
```c++
#ifdef __has_include
#if __has_include(<bit>)
#include <bit>
#endif
#endif

#if __cpp_lib_bit_cast
using std::bit_cast
#else
// ...
```

barcharcraz: I know, I was trying to be minimal

CaseyCarter:
It looks like CI is failing (https://github.com/facebook/folly/pull/1334/checks?check_run_id=499308553#step:25:233) because https://github.com/facebook/folly/blob/b5ef0bf95fa474bacf05831596d70e997e59ecf5/CMakeLists.txt#L823 wants to run a test that was removed in https://github.com/facebook/folly/commit/9489a9c10ca08f747ccafcfc69ad313f796d70fb.

Orvid:
@barcharcraz You'll need to make sure you're included in the microsoft CLA before we can merge this.

Sorry about the cacheline_padded_test stuff, not sure why that wasn't causing our internal tests to fall over, will put up a fix :(

PR #1340, opened 2020-03-24
Fix cache line sizing on s390x

adamretter:
We are using Folly inside RocksDB, and we noticed a build failure due to cache line sizing on s390x. I fixed it in our copy of Folly, and am sending you the changes upstream.

CC @pdillinger

### DIFF
--- a/folly/Portability.h
+++ b/folly/Portability.h
@@ -107,11 +107,18 @@ constexpr bool kHasUnalignedAccess = false;
 #define FOLLY_PPC64 0
 #endif
 
+#if defined(__s390__)
+#define FOLLY_S390 1
+#else
+#define FOLLY_S390 0
+#endif
+
 namespace folly {
 constexpr bool kIsArchArm = FOLLY_ARM == 1;
 constexpr bool kIsArchAmd64 = FOLLY_X64 == 1;
 constexpr bool kIsArchAArch64 = FOLLY_AARCH64 == 1;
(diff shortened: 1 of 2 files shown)


folly/Portability.h
```diff
@@ -107,11 +107,18 @@ constexpr bool kHasUnalignedAccess = false;
 #define FOLLY_PPC64 0
 #endif
 
+#if defined(__s390__)
+#define FOLLY_S390 1
```
Orvid: It's probably worth having this macro and the constexpr below refer to the system as `S390X` to avoid confusion in the future, as searching for `S390` on it's own brings you to the earlier iteration that, at first glance support was discontinued for in 1998.

folly/lang/Align.h
```diff
@@ -120,7 +120,7 @@ struct alignas(max_align_v) max_align_t {};
 //
 //  mimic: std::hardware_destructive_interference_size, C++17
 constexpr std::size_t hardware_destructive_interference_size =
-    kIsArchArm ? 64 : 128;
+    kIsArchArm | kIsArchS390 ? 64 : 128;
```
WillerZ: This should be boolean or `||` and not bitwise or `|` and I would put the condition in parens, so `(kIsArchArm || kIsArchS390)`

adamretter:
@Orvid @WillerZ Thanks for the feedback, I have updated the PR as suggested

PR #1342, opened 2020-03-28
Properly set working directory for shell builder generator

dingxiangfei2009:
cf. https://github.com/facebook/folly/blob/9d715f1f0f683baa25706ed0c442aababf59275d/build/fbcode_builder/fbcode_builder.py#L93

`FBCodeBuilder` constructor fixes `_github_dir`, which implies that the shell builder should initialize this value at construction time.


dingxiangfei2009:
CI fails because `bintray` is busted?

PR #1348, opened 2020-04-06
Fix data race/UB in RCU documentation example

GaneshRapolu:
The read of globalConfigData outside of sm.lock() races with any other writer that is modifying
globalConfigData inside of sm.lock(). This is a documentation only change.


paulmckrcu:
This change looks good, thank you!

And it just might prevent someone from adding a hard-to-find bug into their software.

PR #1349, opened 2020-04-08
Fix clang10 warnings

JoeLoser:
Summary:
- Clang 10 has a new warning: "-Wimplicit-int-float-conversion" which
  flags some code. Fix these up by explicitly casting where appropriate.
- Clang 10 has a new warning: "-Wsizeof-array-div" which flags some
  code. Fix these up by inserting parenthesis.
- Clang 10 uses "-Wdefaulted-function-deleted" which warns about
  `ExecutorTimeoutManager` special member functions being implicitly
  deleted due to the base class of `TimeoutManager` having a deleted move
  constructor. Instead of defaulting them when they will be deleted,
  just remove them.


JoeLoser:
ping @yfeldblum. Just rebased and resolved conflicts.

PR #1351, opened 2020-04-08
Check the cpp attribute is introduced in current c++ standard.

Shylock-Hg:


Shylock-Hg:
As https://en.cppreference.com/w/cpp/feature_test say, the `__has_cpp_attribute` return the time which introduced into cpp draft, so I think we need to check it.

yfeldblum:
Looks like that is only true for standard attributes, such as `[[fallthrough]]`. But not for compiler-specific attributes, such as `[[clang::fallthrough]]` or `[[gnu::fallthrough]]`.

Shylock-Hg:
> Looks like that is only true for standard attributes, such as `[[fallthrough]]`. But not for compiler-specific attributes, such as `[[clang::fallthrough]]` or `[[gnu::fallthrough]]`.

Workable for clang, and gcc too which return true directly.

PR #1352, opened 2020-04-13
FreeBSD build fix proposal.

devnexen:
no tcmalloc support on FreeBSD.


CMake/folly-config.h.cmake
```diff
@@ -26,6 +26,10 @@
 #endif
 #endif // FOLLY_MOBILE
 
+#if defined(__FreeBSD__)
+#define FOLLY_ASSUME_NO_TCMALLOC 1
```
yfeldblum, line 30: This is supposed to be an override.

folly/memory/Malloc.h
```diff
@@ -145,6 +141,10 @@ inline bool getTCMallocNumericProperty(const char* name, size_t* out) noexcept {
     return true;
   }
 #else
+inline bool getTCMallocNumericProperty(const char* name, size_t* out) noexcept {
+  return MallocExtension_Internal_GetNumericProperty(name, strlen(name), out);
```
yfeldblum, line 145: This weak symbol should always be present - folly declares it -, even if it does not link to anything.

CMake/folly-config.h.cmake
devnexen: it was to make it handier for the FreeBSD package

folly/memory/Malloc.h
devnexen: not the case unfortunately but anyway only called within this context.

yfeldblum: Is it a problem to declare these weak symbols also on freebsd?

PR #1369, opened 2020-05-25
Add 0-RTT read early data OpenSSL solution

anirudhaps:
Summary:
In TLS 1.3 handshake, early data comes as an extension of clientHello.
It is read using SSL_read_early_data() OpenSSL api and corresponding
cases based on its state machine are handled. A callback
(parseEarlyDataCb_) is added to verify the read early data. This
callback can be used to reject the early data if not found suitable.

### DIFF
--- a/folly/io/async/AsyncSSLSocket.cpp
+++ b/folly/io/async/AsyncSSLSocket.cpp
@@ -94,6 +94,34 @@ void setup_SSL_CTX(SSL_CTX* ctx) {
 #endif
 }
 
+#if FOLLY_OPENSSL_HAS_TLS13
+// This function is used for logging OpenSSL's early data status
+static std::string earlyDataStatusString(SSL *ssl) {
+  switch(SSL_get_early_data_status(ssl)) {
+    case SSL_EARLY_DATA_REJECTED:
+      return "SSL_EARLY_DATA_REJECTED";
+    case SSL_EARLY_DATA_ACCEPTED:
+      return "SSL_EARLY_DATA_ACCEPTED";
+    default:
+      return "SSL_EARLY_DATA_NOT_SENT";
+  }
+}
+
+// This function is used for logging OpenSSL's early data err code
+static std::string earlyDataErrCodeString(int errCode) {
+  switch(errCode) {
+    case SSL_READ_EARLY_DATA_ERROR:
+      return "SSL_READ_EARLY_DATA_ERROR";
+    case SSL_READ_EARLY_DATA_SUCCESS:
+      return "SSL_READ_EARLY_DATA_SUCCESS";
+    case SSL_READ_EARLY_DATA_FINISH:
+      return "SSL_READ_EARLY_DATA_FINISH";
+    default:
+      return std::to_string(errCode);
+  }
+}
+#endif
+
 // Note: This is a Leaky Meyer's Singleton. The reason we can't use a non-leaky
 // thing is because we will be setting this BIO_METHOD* inside BIOs owned by
 // various SSL objects which may get callbacks even during teardown. We may
@@ -1218,6 +1246,13 @@ void AsyncSSLSocket::handleAccept() noexcept {
     SSL_set_msg_callback_arg(ssl_.get(), this);
   }
 
+#if FOLLY_OPENSSL_HAS_TLS13
+  if (!earlyDataPhaseComplete_ &&
+      handleReadEarlyData() != SSL_READ_EARLY_DATA_FINISH) {
+    return;
+  }
+#endif
+
   DCHECK(ctx_->sslAcceptRunner());
   updateEventRegistration(
       EventHandler::NONE, EventHandler::READ | EventHandler::WRITE);
@@ -1252,6 +1287,9 @@ void AsyncSSLSocket::handleReturnFromSSLAccept(int ret) {
     }
   }
 
+#if FOLLY_OPENSSL_HAS_TLS13
+  earlyDataPhaseComplete_ = false;
+#endif
   handshakeComplete_ = true;
   updateEventRegistration(0, EventHandler::READ | EventHandler::WRITE);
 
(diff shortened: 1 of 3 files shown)


anirudhaps:
Hi @afrind.
I am trying to add a feature to AsyncSSLSocket to handle the early data sent as part of the TLS handshakes clientHello extension when AsyncSSLSocket is used in server mode. This is an OpenSSL solution and uses SSL_read_early_data OpenSSL function to handle incoming early data.

Please review the changes and share any guidelines if this PR is not up to the mark.
As this change is not associated with any open issue, please let me know if an issue is required to be created, if this PR looks ok.

folly/io/async/SSLContext.h
```diff
@@ -574,6 +574,15 @@ class SSLContext {
 
   [[deprecated("Use folly::ssl::init")]] static void initializeOpenSSL();
 
+#if FOLLY_OPENSSL_HAS_TLS13
+  /**
+   * Callback to parse 0-RTT early data.
+   */
+  std::function<bool(
+      const unsigned char *buf,
+      size_t bufSize)> parseEarlyDataCb_;
```
mingtaoy, line 583: This can probably be merged into `AsyncSSLSocket::HandshakeCB` (maybe something like `handshakeEarlyData(folly::Range)`)?

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1218,6 +1246,13 @@ void AsyncSSLSocket::handleAccept() noexcept {
     SSL_set_msg_callback_arg(ssl_.get(), this);
   }
 
+#if FOLLY_OPENSSL_HAS_TLS13
+  if (!earlyDataPhaseComplete_ &&
+      handleReadEarlyData() != SSL_READ_EARLY_DATA_FINISH) {
```
mingtaoy, line 1251: `handleReadEarlyData()` needs to return early to avoid potential buffer allocations if we weren't planning on handling any early data (`SSL_get_max_recv_early_data()` returns 0).

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1359,6 +1397,63 @@ void AsyncSSLSocket::handleConnect() noexcept {
   AsyncSocket::handleInitialReadWrite();
 }
 
+#if FOLLY_OPENSSL_HAS_TLS13
+int AsyncSSLSocket::handleReadEarlyData() noexcept {
+  VLOG(3) << "AsyncSSLSocket::handleReadEarlyData() this=" << this
+          << ", fd=" << fd_ << ", state=" << int(state_) << ", "
+          << "sslState=" << sslState_ << ", events=" << eventFlags_;
+  size_t bufSize = SSL_CTX_get_max_early_data(ctx_->getSSLCtx());
+  unsigned char buf[bufSize];
```
mingtaoy, line 1406: Variable length arrays are discouraged (especially since `SSL_set_max_early_data` blindly accepts a value).

One concrete reason is that AsyncSSLSocket::handleAccept() can be called under a fiber context, where the fiber's stack size can be fairly small. 

Unfortunately, I don't think it's guaranteed that the application's read callback is set at this point (normally, we would defer buffer allocation via the `ReadCallback`). The best thing that I can think of here is to just reserve a small fixed size buffer (~e.g. 512 bytes) and continuously call SSL_read_early_data() to handle it.

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1359,6 +1397,63 @@ void AsyncSSLSocket::handleConnect() noexcept {
...
+    case SSL_READ_EARLY_DATA_SUCCESS:
+      if (ctx_->parseEarlyDataCb_ &&
+          !ctx_->parseEarlyDataCb_(buf, readBytes)) {
+        VLOG(3) << "Early data content rejected ";
+        sslState_ = STATE_ERROR;
+        AsyncSocketException ex(AsyncSocketException::EARLY_DATA_REJECTED,
+                                "Invalid Early data content");
+        failHandshake(__func__, ex);
+        break;
+      }
```
mingtaoy, line 1442: This does not mean that you fully received all early data application records. Quoting OpenSSL's documentation:

> SSL_READ_EARLY_DATA_SUCCESS
This indicates that early data was successfully read. This should be treated in the same way as a 1 return value from SSL_read_ex(3). **You should continue to call SSL_read_early_data() to read more data.**

(emphasis mine).

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1359,6 +1397,63 @@ void AsyncSSLSocket::handleConnect() noexcept {
...
+  VLOG(3) << "AsyncSSLSocket::handleReadEarlyData() this=" << this
+          << ", fd=" << fd_ << ", state=" << int(state_) << ", "
+          << "sslState=" << sslState_ << ", events=" << eventFlags_;
+  size_t bufSize = SSL_CTX_get_max_early_data(ctx_->getSSLCtx());
+  unsigned char buf[bufSize];
+  size_t readBytes;
+  int edRet = SSL_READ_EARLY_DATA_ERROR;
+  int sslErr = SSL_ERROR_NONE;
+
+  edRet = SSL_read_early_data(ssl_.get(), buf, bufSize, &readBytes);
```
mingtaoy, line 1411: `SSL_read_early_data` may need to be called multiple times (if the buffer was not large enough). We should drain the early data buffer in the same event loop iteration (until FINISH, or EAGAIN) to avoid an unnecessary re-loop.

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1359,6 +1397,63 @@ void AsyncSSLSocket::handleConnect() noexcept {
...
+      if (SSL_ERROR_SSL == sslErr || SSL_ERROR_SYSCALL == sslErr) {
+        // Unrecoverable error in OpenSSL lib occured due to abrupt EOF sent
+        // by peer. Aborting the early data phase.
+        earlyDataPhaseComplete_ = true;
+        break;
+      }
+      // Either SSL_accept or SSL_read_ex inside SSL_read_early_data() returned
+      // SSL_ERROR_WANT_READ. Retry the operation by calling
+      // SSL_read_early_data again in next read event
+      updateEventRegistration(EventHandler::READ, EventHandler::WRITE);
```
mingtaoy, line 1430: It's not sufficient to just assume this. Again, for future compatibility, if additional error codes are added that are not related to I/O, then this would be incorrect to assume.

folly/io/async/AsyncSSLSocket.cpp
```diff
@@ -1359,6 +1397,63 @@ void AsyncSSLSocket::handleConnect() noexcept {
...
+      sslState_ = STATE_ACCEPTING;
```
mingtaoy, line 1431: All errors other than WANT_READ and WANT_WRITE should be fatal to the connection, and should end in a call to `failHandshake()`, rather than just setting `earlyDataPhaseComplete`.

anirudhaps:
Thanks @mingtaoy for review. I will address all your comments.

folly/io/async/AsyncSSLSocket.cpp
anirudhaps: According to OpenSSL [documentation](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_get_max_early_data.html),

> The recv_max_early_data value also has an impact on early data that is accepted. The amount of data that is accepted will always be the lower of the max_early_data for the session and the recv_max_early_data setting for the server. If a client sends more data than this then the connection will abort.


Going with the above description, suppose the server sets the max_early_data = 1024 bytes and the same is notified to the client. Now, if recv_max_early_data = 0 and clients sends some early data whose size > 0, then the connection must abort. But the fact that client has sent some illegal amount of data will be recognised only if we call `SSL_read_early_data()` which will also require potential buffer allocation. If `handleReadEarlyData()` is returned early and `SSL_read_early_data()` is not invoked, then we cannot check if the client has sent some illegal amount of early data (in this case, size > 0 when recv_max_early_data = 0).
Thus, I think returning early might not be good idea in case of recv_max_early_data = 0.

anirudhaps:
Hi @mingtaoy
I have addressed most of the review comments for this PR. Please review.

PR #1387, opened 2020-06-14
Handle multiple SSL_ERROR_WANT_ASYNC which is returned by ASYNC engine

nidhidamodaran:
@afrind , Please review the updated changes as per your comments. 

The changes are required to handle cases where multiple SSL_ERROR_WANT_ASYNC is returned by the async engine.

Also I have removed the code which closes the async fd as this is my crypto driver handle and should not be closed by upper layers as I require it further in my SSL connection

### DIFF
--- a/folly/io/async/AsyncPipe.cpp
+++ b/folly/io/async/AsyncPipe.cpp
@@ -48,8 +48,6 @@ void AsyncPipeReader::close() {
 
     if (closeCb_) {
       closeCb_(fd_);
-    } else {
-      netops::close(fd_);
     }
     fd_ = NetworkSocket();
   }
--- a/folly/io/async/AsyncSSLSocket.h
+++ b/folly/io/async/AsyncSSLSocket.h
@@ -175,9 +175,9 @@ class AsyncSSLSocket : public AsyncSocket {
 
     void readDataAvailable(size_t len) noexcept override {
       CHECK_EQ(len, 1);
-      sslSocket_->restartSSLAccept();
       pipeReader_->setReadCB(nullptr);
       sslSocket_->setAsyncOperationFinishCallback(nullptr);
+      sslSocket_->restartSSLAccept();
     }
 
     void getReadBuffer(void** bufReturn, size_t* lenReturn) noexcept override {


folly/io/async/AsyncPipe.cpp
```diff
@@ -48,8 +48,6 @@ void AsyncPipeReader::close() {
 
     if (closeCb_) {
       closeCb_(fd_);
-    } else {
-      netops::close(fd_);
```
afrind, line 52: Removing this seems wrong - close() no longer closes the underlying fd when closeCb_ is not defined?

nidhidamodaran: This fd was added in async list by underlying openssl. Isn't it better to let the ssl library close the fd once operation is complete. 

Also, If I close here, my driver handle will be closed and cannot be used for any further in the SSL connection.

nidhidamodaran:
[test_patch.txt](https://github.com/facebook/folly/files/4812172/test_patch.txt)

@afrind , please use the above patch to reproduce issue I currently face.  I created a test case to return multiple SSL_ERROR_WANT_ASYNC multiple times, but when run it always returns a SSL_ERROR_WANT_READ between two continuous SSL_ERROR_WANT_ASYNC errors.  Though the attached patch is not adding really creating an async job, but you can use it to quickly reproduce the error.

The original folly code will hang with this change, ie when 2 continuous  SSL_ERROR_WANT_ASYNC is received, and this PR fixes the issue. Please verify.
Let me know if you need any further details

folly/io/async/AsyncPipe.cpp
nidhidamodaran: @afrind  I will elaborate my use case too here:

I have a crypto device to which asymmetric and symmetric operations from openssl are offloaded. Instead of pipe fd, I am using the driver handle of this device as an async fd, which is signalled whenever the crypto operation is completed.

So this handle has to be open throughout the length of an SSL connection. Also in this design,  for each of crypto  operation done in ASYNC mode, the SSL_ERROR_WANT_ASYNC is returned. As such after such a request is processed at folly layer and SSL_accept restarted, another crypto operation may return same error and thereby returning multiple WANT_ASYNC errors, which the existing folly is not handling correctly. 

The issue is that if sslSocket_->restartSSLAccept again returns WANT_ASYNC error,  asyncOperationFinishCallback_ is not be set null and further events not processed

nidhidamodaran:
Also please note :Existing folly code works correctly if between two SSL_ERRROR_WANT_ASYNC error codes,  a non-fatal error code is returned as this will clean all stale data structures.

nidhidamodaran:
@afrind , I have added a testcase to return SSL_ERROR_WANT_ASYNC multiple times. I have also committed cert/key/CAcert as the one given in repo doesn't seem to be working. 

Please review. Please note the error is only reproducible if between 2 SSL_ERROR_WANT_ASYNC errors, no other error codes are returned, say SSL_ERROR_WANT_READ (2)

folly/io/async/AsyncPipe.cpp
afrind: AsyncPipeWriter is generic and can have uses other than this use case.  I think what you want to do instead is set the closeCB_ to a no-op in your case?

nidhidamodaran: @afrind , Thanks for the review. I have removed the change so as to not break other use cases. Please review rest of the code changes.

yfeldblum:
Legit.

Can we avoid checking in new keys/certs? It should be possible to generate small keys quickly and in-memory at unit-test time.

nidhidamodaran:
@afrind @yfeldblum , I have removed the certs committed earlier and used the existing certs. Please review.

folly/io/async/test/AsyncSSLSocketTest.cpp
```diff
@@ -1525,6 +1525,79 @@ static int kRSAEvbExIndex = -1;
 static int kRSASocketExIndex = -1;
 static constexpr StringPiece kEngineId = "AsyncSSLSocketTest";
 
+int customRsaPubDec(int flen, const unsigned char *from, unsigned char *to, RSA *rsa,
```
afrind, line 1528: Should be 'static'

folly/io/async/test/AsyncSSLSocketTest.cpp
```diff
@@ -1525,6 +1525,79 @@ static int kRSAEvbExIndex = -1;
...
+    writer->write(nullptr, &byte, 1);
+  });
+
+  LOG(INFO) << "About to pause job";
+  ASYNC_pause_job();
+  LOG(INFO) << "Resumed job with ret: " << ret;
+  return ret;
+}
+
+int verifyCb(X509_STORE_CTX *ctx, void *arg) {
```
afrind, line 1581: Should be 'static'

folly/io/async/test/AsyncSSLSocketTest.cpp
```diff
@@ -1525,6 +1525,79 @@ static int kRSAEvbExIndex = -1;
...
+  int ok = 1;
+  X509 *cert = X509_STORE_CTX_get0_cert(ctx);
+  CHECK(cert);
+
+  ssl::EvpPkeyUniquePtr publicEvpPkey(X509_get_pubkey(cert));
+  EVP_PKEY *pkey = publicEvpPkey.get();
+  RSA* rsa = EVP_PKEY_get1_RSA(publicEvpPkey.get());
```
afrind, line 1588: I think this is a leak (openssl _get1 increments (adds by 1) the reference count of the undelrying object), and the reason why tests aren't catching this is because we disable this test when running under ASAN mode.

afrind: See line 1762 for where we disable

folly/io/async/AsyncSSLSocket.h
```diff
@@ -175,9 +175,9 @@ class AsyncSSLSocket : public AsyncSocket {
 
     void readDataAvailable(size_t len) noexcept override {
       CHECK_EQ(len, 1);
-      sslSocket_->restartSSLAccept();
       pipeReader_->setReadCB(nullptr);
       sslSocket_->setAsyncOperationFinishCallback(nullptr);
+      sslSocket_->restartSSLAccept();
```
afrind, line 180: Please add a comment here explaining why order is significant

nidhidamodaran: @afrind  I have addressed all your comments, please review

afrind:
The test failed with a SEGV internally in opt mode?

```
*** Aborted at 1595628991 (Unix time, try 'date -d @1595628991') ***
*** Signal 11 (SIGSEGV) (0x0) received by PID 2984783 (pthread TID 0x7fbba672a3c0) (linux TID 2984783) (code: 128), stack trace: ***
    @ 000000000061b6f5 folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo_t*, void*)
                       ./folly/experimental/symbolizer/SignalHandler.cpp:442
    @ 0000000000000000 (unknown)
    @ 00000000001d6e1e RSA_free
    @ 0000000000193f22 EVP_PKEY_free
    @ 000000000022cbfb pubkey_cb
    @ 00000000000a365e asn1_item_embed_free
    @ 00000000000a397f asn1_template_free
    @ 00000000000a3642 asn1_item_embed_free
    @ 00000000000a397f asn1_template_free
    @ 00000000000a3642 asn1_item_embed_free
    @ 00000000000a38b6 ASN1_item_free
    @ 000000000020b53a OPENSSL_sk_pop_free
    @ 000000000003dffa SSL_free
    @ 000000000056f100 folly::AsyncSSLSocket::~AsyncSSLSocket()
                       ./folly/Memory.h:302
                       -> ./folly/io/async/AsyncSSLSocket.cpp
    @ 000000000056f6bd folly::AsyncSSLSocket::~AsyncSSLSocket()
                       ./folly/io/async/AsyncSSLSocket.cpp:315
    @ 000000000058b634 folly::AsyncSocket::destroy()
                       ./folly/io/async/DelayedDestruction.h:55
                       -> ./folly/io/async/AsyncSocket.cpp
    @ 0000000000448019 folly::AsyncSSLSocketTest_OpenSSL110MultipleAsyncTest_Test::TestBody()
                       ./folly/io/async/DelayedDestruction.h:70
                       -> ./folly/io/async/test/AsyncSSLSocketTest.cpp
    @ 000000000069433e void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:2473
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 0000000000685281 testing::Test::Run() [clone .part.595]
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:2490
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 0000000000685701 testing::TestInfo::Run() [clone .part.596]
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:2481
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 0000000000685894 testing::TestCase::Run() [clone .part.597]
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:2801
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 0000000000686e34 testing::internal::UnitTestImpl::RunAllTests()
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:4735
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 0000000000687124 testing::UnitTest::Run()
                       /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest.cc:2473
                       -> /home/engshare/third-party2/googletest/master/src/googletest/googletest/src/gtest-all.cc
    @ 000000000048abe5 main
                       ./third-party-buck/platform007/build/googletest/include/gtest/gtest.h:2255
                       -> ./folly/io/async/test/AsyncSSLSocketTest2.cpp
    @ 00000000000211a5 __libc_start_main
    @ 0000000000420679 _start
                       /home/engshare/third-party2/glibc/2.26/src/glibc-2.26/csu/../sysdeps/x86_64/start.S:120

```

nidhidamodaran:
@afrind please help me reproduce this error. How do I build folly in opt mode ?

afrind:
I'm actually not sure since we use a different build system internally.  I think it's a combination of -NDEBUG and -O3 mostly?  If you are linking against jemalloc, try MALLOC_CONF=junk:true,abort:true ?

Issue #1383, opened 2020-06-09
Incomplete build documentation for Fedora


martind1111:
The README.md file provides instructions on building folly for various operating systems. There is a section that talks about other Linux distributions with mentions to Fedora. I have used these instructions to build folly on my Fedora VM, which runs the latest Fedora build (Fedora release 32). On this system, the folly build fails with the following error:
```
[ 41%] Building CXX object CMakeFiles/folly_base.dir/folly/experimental/symbolizer/StackTrace.cpp.o
/home/martin/git/folly/folly/experimental/symbolizer/StackTrace.cpp:21:10: fatal error: libunwind.h: No such file or directory
   21 | #include <libunwind.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/folly_base.dir/build.make:1487: CMakeFiles/folly_base.dir/folly/experimental/symbolizer/StackTrace.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:221: CMakeFiles/folly_base.dir/all] Error 2
make: *** [Makefile:150: all] Error 2
```
Which shows that the dependency libunwind-devel is required to build folly.

The README.md file also lists the optional dependencies to build folly. I noticed that a number of packages were missing from this list:
liburing
liburing-devel
libsodium
libsodium-devel
xz
xz-devel
bzip2
bzip2-devel
zstd
libzstd-devel

The following optional dependencies also need to be revisted: libdwarf-dev should be replaced with libdwarf-devel and libundwind8-dev should be removed (libunwind is actually a mandatory requirement).

The purpose of this issue is to update documentation for the building of folly on Fedora systems.

yfeldblum:
Current instructions in the readme describe using `getdeps.py` to build on Fedora.

PR #1405, opened 2020-07-02
Update the documentation to build folly on Fedora systems

martind1111:
Following instructions in README.md causes build to fail after running cmake on Fedora systems as explained in #1383 . This pull request addresses this problem and more thoroughly documents the packages that are optional for the folly library.


yfeldblum:
Current instructions in the readme describe using `getdeps.py` to build on Fedora.

PR #1409, opened 2020-07-09
Fix #1391: Patch cmake files to fix cmake build issue on CentOS

### LINKED ISSUE #1391: cmake fails on CentOS 8.1 and issues with folly compile
I have tried to build folly on CentOS 8.1 with cmake. There are some issues with the cmake files. Most of the compile checks fail to work and this prevents a developer from building the library on CentOS. After some research, I discovered that this issue had already been encountered (and fixed) in another github project (facebookincubator mvfst). I have tested the fix and it also resolves the issue with building on CentOS for folly. The issue is the handling of gflags. From my understanding, folly-deps only handles gflags module mode, but not config mode. The fix is to change the handling of gflags in folly-deps.cmake so that it looks like this:
```
find_package(gflags CONFIG QUIET)
if (gflags_FOUND)
  message(STATUS "Found gflags from package config")
  set(FOLLY_HAVE_LIBGFLAGS ${gflags_FOUND})
  if (TARGET gflags-shared)
    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags-shared)
  elseif (TARGET gflags)
    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags)
  else()
    message(FATAL_ERROR "Unable to determine the target name for the GFlags package.")
  endif()
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${GFLAGS_LIBRARIES})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${GFLAGS_INCLUDE_DIR})
else()
  find_package(Gflags MODULE)
  set(FOLLY_HAVE_LIBGFLAGS ${LIBGFLAGS_FOUND})
  list(APPEND FOLLY_LINK_LIBRARIES ${LIBGFLAGS_LIBRARY})
  list(APPEND FOLLY_INCLUDE_DIRECTORIES ${LIBGFLAGS_INCLUDE_DIR})
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${LIBGFLAGS_LIBRARY})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${LIBGFLAGS_INCLUDE_DIR})
endif()
```
And then, create a file named FindGflags.cmake under CMake with the following content:
```
# Copyright (c) Facebook, Inc. and its affiliates.
# Find libgflags.
# There's a lot of compatibility cruft going on in here, both
# to deal with changes across the FB consumers of this and also
# to deal with variances in behavior of cmake itself.
#
# Since this file is named FindGflags.cmake the cmake convention
# is for the module to export both GFLAGS_FOUND and Gflags_FOUND.
# The convention expected by consumers is that we export the
# following variables, even though these do not match the cmake
# conventions:
#
#  LIBGFLAGS_INCLUDE_DIR - where to find gflags/gflags.h, etc.
#  LIBGFLAGS_LIBRARY     - List of libraries when using libgflags.
#  LIBGFLAGS_FOUND       - True if libgflags found.
#
# We need to be able to locate gflags both from an installed
# cmake config file and just from the raw headers and libs, so
# test for the former and then the latter, and then stick
# the results together and export them into the variables
# listed above.
#
# For forwards compatibility, we export the following variables:
#
#  gflags_INCLUDE_DIR - where to find gflags/gflags.h, etc.
#  gflags_TARGET / GFLAGS_TARGET / gflags_LIBRARIES
#                     - List of libraries when using libgflags.
#  gflags_FOUND       - True if libgflags found.
#

IF (LIBGFLAGS_INCLUDE_DIR)
  # Already in cache, be silent
  SET(Gflags_FIND_QUIETLY TRUE)
ENDIF ()

find_package(gflags CONFIG QUIET)
if (gflags_FOUND)
  if (NOT Gflags_FIND_QUIETLY)
    message(STATUS "Found gflags from package config ${gflags_CONFIG}")
  endif()
  # Re-export the config-specified libs with our local names
  set(LIBGFLAGS_LIBRARY ${gflags_LIBRARIES})
  set(LIBGFLAGS_INCLUDE_DIR ${gflags_INCLUDE_DIR})
  set(LIBGFLAGS_FOUND ${gflags_FOUND})
  # cmake module compat
  set(GFLAGS_FOUND ${gflags_FOUND})
  set(Gflags_FOUND ${gflags_FOUND})
else()
  FIND_PATH(LIBGFLAGS_INCLUDE_DIR gflags/gflags.h)

  FIND_LIBRARY(LIBGFLAGS_LIBRARY_DEBUG NAMES gflagsd gflags_staticd)
  FIND_LIBRARY(LIBGFLAGS_LIBRARY_RELEASE NAMES gflags gflags_static)

  INCLUDE(SelectLibraryConfigurations)
  SELECT_LIBRARY_CONFIGURATIONS(LIBGFLAGS)

  # handle the QUIETLY and REQUIRED arguments and set LIBGFLAGS_FOUND to TRUE if
  # all listed variables are TRUE
  INCLUDE(FindPackageHandleStandardArgs)
  FIND_PACKAGE_HANDLE_STANDARD_ARGS(gflags DEFAULT_MSG LIBGFLAGS_LIBRARY LIBGFLAGS_INCLUDE_DIR)
  # cmake module compat
  set(Gflags_FOUND ${GFLAGS_FOUND})
  # compat with some existing FindGflags consumers
  set(LIBGFLAGS_FOUND ${GFLAGS_FOUND})

  # Compat with the gflags CONFIG based detection
  set(gflags_FOUND ${GFLAGS_FOUND})
  set(gflags_INCLUDE_DIR ${LIBGFLAGS_INCLUDE_DIR})
  set(gflags_LIBRARIES ${LIBGFLAGS_LIBRARY})
  set(GFLAGS_TARGET ${LIBGFLAGS_LIBRARY})
  set(gflags_TARGET ${LIBGFLAGS_LIBRARY})

  MARK_AS_ADVANCED(LIBGFLAGS_LIBRARY LIBGFLAGS_INCLUDE_DIR)
endif()

# Compat with the gflags CONFIG based detection
if (LIBGFLAGS_FOUND AND NOT TARGET gflags)
  add_library(gflags UNKNOWN IMPORTED)
  set_target_properties(gflags PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${LIBGFLAGS_INCLUDE_DIR}")
  set_target_properties(gflags PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "C" IMPORTED_LOCATION "${LIBGFLAGS_LIBRARY}")
endif()
```
See https://github.com/facebookincubator/mvfst/pull/28/files and https://github.com/facebook/folly/pull/1170.

If you need me to, I could create a pull request to solve this issue in folly.

martind1111:


Zenix27 (approved):

martind1111:
OK. Will do.

Martin

martind1111:
I have removed FindGflags.cmake from the CMake directory. The changes to folly-deps.cmake alone seems to address the gflags problem.

martind1111:
Here is the output of building folly on CentOS 8.2 prior to the changes to folly-deps.cmake included in this pull request:

[user@localhost cmake-build-release]$ cmake ..
-- Boost version: 1.66.0
-- Found the following Boost libraries:
--   context
--   filesystem
--   program_options
--   regex
--   system
--   thread
--   chrono
--   date_time
--   atomic
-- Found double-conversion: /usr/lib64/libdouble-conversion.so
-- Found gflags from package config /usr/lib64/cmake/gflags/gflags-config.cmake
-- Found libevent: /usr/lib64/libevent.so
-- Looking for lzma_easy_encoder in /usr/lib64/liblzma.so
CMake Error in /home/user/git/folly/cmake-build-release/CMakeFiles/CMakeTmp/CMakeLists.txt:
  Imported target "gflags" includes non-existent path

    "//include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



CMake Error in /home/user/git/folly/cmake-build-release/CMakeFiles/CMakeTmp/CMakeLists.txt:
  Imported target "gflags" includes non-existent path

    "//include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.



CMake Error at /usr/share/cmake/Modules/CheckLibraryExists.cmake:56 (try_compile):
  Failed to generate test project build system.
  /usr/share/cmake/Modules/FindLibLZMA.cmake:47 (CHECK_LIBRARY_EXISTS)
  CMake/folly-deps.cmake:92 (find_package)
  CMakeLists.txt:113 (include)


martind1111:
> 
> 
> Thanks for contributing!
> 
> Would you mind revising this PR so that it updates the existing https://github.com/facebook/folly/blob/master/build/fbcode_builder/CMake/FindGflags.cmake file instead of adding a second, different copy?
> We use a common build library for this stuff and it took a lot of effort to get it where it is; as written this PR is moving us away from that.
> 
> Please also share the exact error messages that you were seeing before this change as that helps to understand the nature of the original problem and thus to evaluate the solution.

Addressed both concerns.

simpkins:
This logic should be changed in the existing build/fbcode_builder/CMake/FindGflags.cmake file, rather than putting it directly in folly-deps.cmake.

The FindGflags.cmake file is shared across multiple different Facebook open source repositories.  (e.g., https://github.com/facebook/watchman/; https://github.com/facebook/proxygen; https://github.com/facebook/proxygen; etc). All logic for how to find gflags in a platform independent way should live in this file so that all repositories benefit from the changes.

martind1111:
> 
> 
> This logic should be changed in the existing build/fbcode_builder/CMake/FindGflags.cmake file, rather than putting it directly in folly-deps.cmake.
> 
> The FindGflags.cmake file is shared across multiple different Facebook open source repositories. (e.g., https://github.com/facebook/watchman/; https://github.com/facebook/proxygen; https://github.com/facebook/fboss; etc). All logic for how to find gflags in a platform independent way should live in this file so that all repositories benefit from the changes.

I have moved all code related to finding gflags from CMake/folly-deps.cmake to build/fbcode_builder/CMake/FindGflags.cmake.

simpkins:
This still doesn't look quite right.  All folly-specific logic needs to remain in folly-deps.cmake.  The code in FindGflags.cmake is used across multiple projects, and should not have folly specific code.  e.g., it should not update FOLLY_SHINY_DEPENDENCIES or FOLLY_INCLUDE_DIRECTORIES, and it should not update CMAKE_REQUIRED_LIBRARIES or CMAKE_REQUIRED_INCLUDES.

The purpose of FindGflags.cmake is to provide consistent behavior for finding gflags that works across multiple platforms and gflags versions, and then export targets and variables so that individual projects can use this data.  It primarily exists because gflags has changed the config variables they export across different versions, so it attempts to export a consistent set of variables + targets regardless of what version is available.

Basically, FindGflags.cmake should be updated to make sure it consistently exports LIBGFLAGS_INCLUDE_DIR, LIBGFLAGS_LIBRARY, gflags_INCLUDE_DIR, and gflags_TARGET correctly on CentOS 8.2, so that folly-deps.cmake can use these variables.

build/fbcode_builder/CMake/FindGflags.cmake
```diff
@@ -45,6 +45,16 @@ if (gflags_FOUND)
...
+  set(FOLLY_HAVE_LIBGFLAGS ${gflags_FOUND})
+  if (TARGET gflags-shared)
+    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags-shared)
+  elseif (TARGET gflags)
+    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags)
+  else()
+    message(FATAL_ERROR "Unable to determine the target name for the GFlags package.")
+  endif()
+  list(APPEND CMAKE_REQUIRED_LIBRARIES ${GFLAGS_LIBRARIES})
+  list(APPEND CMAKE_REQUIRED_INCLUDES ${GFLAGS_INCLUDE_DIR})
```
simpkins, line 57: Are these really right?  Current gflags cmake config packages look like they export gflags_LIBRARIES and gflags_INCLUDE_DIR, not GFLAGS_LIBRARIES and GFLAGS_INCLUDE_DIRECTORIES.

martind1111:
As stated in issue #1391, I have copied the code that was introduced in another Facebook project (facebookincubator/mvfst) to handle Gflags in folly. I thought simpkins that you wanted me to push the logic of the FindGflags into build/fbcode_builder/CMake/FindGflags.cmake and I can see now that it does not make much sense to move anything that is folly specific there. But now, I am wondering what you wanted me to change with regards to the version you commented on. Isn't anything in the following specific to folly?
`
find_package(gflags CONFIG QUIET)
if (gflags_FOUND)
  message(STATUS "Found gflags from package config")
  set(FOLLY_HAVE_LIBGFLAGS ${gflags_FOUND})
  if (TARGET gflags-shared)
    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags-shared)
  elseif (TARGET gflags)
    list(APPEND FOLLY_SHINY_DEPENDENCIES gflags)
  else()
    message(FATAL_ERROR "Unable to determine the target name for the GFlags package.")
  endif()
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${GFLAGS_LIBRARIES})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${GFLAGS_INCLUDE_DIR})
else()
  find_package(Gflags MODULE)
  set(FOLLY_HAVE_LIBGFLAGS ${LIBGFLAGS_FOUND})
  list(APPEND FOLLY_LINK_LIBRARIES ${LIBGFLAGS_LIBRARY})
  list(APPEND FOLLY_INCLUDE_DIRECTORIES ${LIBGFLAGS_INCLUDE_DIR})
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${LIBGFLAGS_LIBRARY})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${LIBGFLAGS_INCLUDE_DIR})
endif()
`
What is it that you would want me to move in build/fbcode_builder/CMake/FindGflags.cmake?
I'll revert the pull request to what it was prior to my last commit for the time being and then I can apply changes that you would like to see.

simpkins:
The desired goal is to have the code in build/fbcode_builder/CMake/FindGflags.cmake find gflags and correctly export the ${LIBGFLAGS_INCLUDE_DIR} and ${LIBGFLAGS_LIBRARY} variables for the current platform.  Then the code in folly-deps.cmake should use these variables to set variables specific to the folly build.  (e.g., ${FOLLY_LINK_LIBRARIES}, ${FOLLY_INCLUDE_DIRECTORIES}, ${CMAKE_REQUIRED_LIBRARIES}, and ${CMAKE_REQUIRED_INCLUDES}).

If the current FindGflags.cmake behavior is not correctly exporting ${LIBGFLAGS_INCLUDE_DIR} then FindGflags.cmake should be updated to export a correct value.  The code in folly-deps.cmake likely won't do change if ${LIBGFLAGS_INCLUDE_DIR} is set correctly.

I just had a chance to take a look at this on a CentOS 8.1 system and this looks potentially like a bug in the CentOS gflags-devel RPM, or how CMake is configured by default on CentOS.  The problem appears to stem from the fact that `/lib64` is just a symlink to `/usr/lib64`, but `/lib64` appears first in CMake's default search path.  CMake finds the gflags config file at `/lib64/cmake/gflags/gflags-config.cmake`, and the gflags code exports ${gflags_INCLUDE_DIR} relative to the location of this config file.  This ends up causing it to incorrectly export `${gflags_INCLUDE_DIR}` as `//include`, when it should be `/usr/include`.

It seems like this bug ideally should be fixed upstream in the CentOS RPMs, although it would seem reasonable to add a workaround for this in `FindGflags.cmake`.

That said, another alternative would be to simply build folly with a local gflags installation.  We haven't run across this failure in our own continuous integration builds since we always build folly against a more recent local build of gflags, and we always link statically against gflags.  You can run `./build/fbcode_builder/getdeps.py build folly --scratch-path _build` to build folly the same way we do normally.  This script will download and build most of folly's dependencies first.

simpkins:
The following changes to FindGflags.cmake appear to work for me to build with the system's gflags-devel RPM on CentOS 8.1: https://github.com/simpkins/folly/commit/835fe8262d52cf4a13c09cb69f4cf168ed5f2d94

I've submitted these changes as an internal pull request.

I'm also working on some documentation updates to the build instructions in the README.  Our continuous integrations scripts all build with the build/fbcode_builder/getdeps.py script.  This script builds gflags statically from source during the build, and never attempts to use the CentOS gflags-devel RPM.  That's one of the reasons we never noticed the issue when attempting to build with the system gflags-devel RPM.

martind1111:
> 
> 
> The following changes to FindGflags.cmake appear to work for me to build with the system's gflags-devel RPM on CentOS 8.1: [simpkins@835fe82](https://github.com/simpkins/folly/commit/835fe8262d52cf4a13c09cb69f4cf168ed5f2d94)
> 
> I've submitted these changes as an internal pull request.
> 
> I'm also working on some documentation updates to the build instructions in the README. Our continuous integrations scripts all build with the build/fbcode_builder/getdeps.py script. This script builds gflags statically from source during the build, and never attempts to use the CentOS gflags-devel RPM. That's one of the reasons we never noticed the issue when attempting to build with the system gflags-devel RPM.

How do you want to handle this? Would you like that I update my pull request to reflect your changes, dropping whatever changes I had put forward in my pull request? Or will you create a new pull request yourself? I can update the description of the pull request to include the comments that you have listed under your commit and state that these changes comes from you.

martind1111:
> 
> 
> The following changes to FindGflags.cmake appear to work for me to build with the system's gflags-devel RPM on CentOS 8.1: [simpkins@835fe82](https://github.com/simpkins/folly/commit/835fe8262d52cf4a13c09cb69f4cf168ed5f2d94)
> 
> I've submitted these changes as an internal pull request.
> 
> I'm also working on some documentation updates to the build instructions in the README. Our continuous integrations scripts all build with the build/fbcode_builder/getdeps.py script. This script builds gflags statically from source during the build, and never attempts to use the CentOS gflags-devel RPM. That's one of the reasons we never noticed the issue when attempting to build with the system gflags-devel RPM.

Thanks for addressing the issue. Looking forward being able to successfully build folly in CentOS 8.1.

PR #1433, opened 2020-08-31
CMake: Handle generator expressions requiring a target for pkg-config

### LINKED ISSUE #1414: $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets
I'm building folly `v2020.07.13.00` on Ubuntu 18.04 x86_64 with the standard system gcc but a newer cmake 3.18.0. On configure, I get the cmake error:
```
[...]
 -- Found gmock via config, defines=, include=/data/Debug/include, libs=GTest::gmock_main;GTest::gmock;GTest::gtest
 -- Found GTest: /data/Debug/lib/libgtest.a  
 -- Configuring done
 CMake Error at CMakeLists.txt:446 (file):
   Error evaluating generator expression:
     $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>
   $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets to
   specify include directories, compile definitions, and compile options.  It
   may not be used with the add_custom_command, add_custom_target, or
   file(GENERATE) commands.
 CMake Error at CMakeLists.txt:446 (file):
   Error evaluating generator expression:
     $<COMPILE_LANG_AND_ID:CUDA,NVIDIA>
   $<COMPILE_LANG_AND_ID:lang,id> may only be used with binary targets to
   specify include directories, compile definitions, and compile options.  It
   may not be used with the add_custom_command, add_custom_target, or
   file(GENERATE) commands.
 -- Generating done
```
cmake ends in error after that. I googled for this problem but could not find anything really useful. Also I can not find the lines `$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>` in folly so I'm slightly confused where the problem originates from. Can someone point me in the right direction please?

tambry:
CMake's find modules may have target-based generator expressions, which get propagated to us when generating the pkg-config file.
CMake 3.19 introduces the TARGET argument to file(GENERATE), which allows resolving such generator expressions and will avoid future issues.
A workaround proposed by @adriaandegroot is added for CMake 3.18.

Additionally changed list->string conversion to use list(JOIN) instead of string replacement.

Fixes #1414.

### DIFF
--- a/CMake/GenPkgConfig.cmake
+++ b/CMake/GenPkgConfig.cmake
@@ -92,7 +92,14 @@ function(gen_pkgconfig_vars)
 
   # Set the output variables
   string(REPLACE ";" " " cflags "${cflags}")
-  set("${var_prefix}_CFLAGS" "${cflags}" PARENT_SCOPE)
   string(REPLACE ";" " " private_libs "${private_libs}")
+
+  # Since CMake 3.18 FindThreads may include a generator expression requiring a target, which gets propagated to us through INTERFACE_COMPILE_OPTIONS.
+	# Before CMake 3.19 there's no way to solve this in a general way, so we work around the specific case. See #1414 and CMake bug #21074.
+  if(CMAKE_VERSION VERSION_LESS 3.19)
+    string(REPLACE "<COMPILE_LANG_AND_ID:CUDA,NVIDIA>" "<COMPILE_LANGUAGE:CUDA>" cflags "${cflags}")
+  endif()
+
+  set("${var_prefix}_CFLAGS" "${cflags}" PARENT_SCOPE)
   set("${var_prefix}_PRIVATE_LIBS" "${private_libs}" PARENT_SCOPE)
 endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -436,10 +436,17 @@ configure_file(
   ${CMAKE_CURRENT_BINARY_DIR}/libfolly.pc.gen
   @ONLY
 )
+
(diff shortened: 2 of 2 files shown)


danobi:
Thanks for the fix. I hit this issue as well.

CMake/GenPkgConfig.cmake
```diff
@@ -91,8 +91,15 @@ function(gen_pkgconfig_vars)
   endif()
 
   # Set the output variables
-  string(REPLACE ";" " " cflags "${cflags}")
+  list(JOIN cflags " " cflags)
```
Orvid, line 94: `list(JOIN` is not supported on all versions of CMake that we support building with, so this will need to stay as replace for now.

tambry: Didn't notice that Folly supports such ancient CMake versions. Fixed.

CMakeLists.txt
```diff
@@ -436,10 +436,17 @@ configure_file(
   ${CMAKE_CURRENT_BINARY_DIR}/libfolly.pc.gen
   @ONLY
 )
+
+# Specify target to allow resolving generator expressions requiring a target for CMake >=3.19. See #1414.
+if(CMAKE_VERSION VERSION_GREATER 3.18)
```
tambry, line 441: I also replaced `VERSION_GREATER_EQUAL` here as it's only available in CMake >=3.7.

MrAnno: `VERSION_GREATER 3.18` matches on non-zero 3.18 patch versions as well, so it breaks with v3.18.2, for example.

tambry: See #1439.

PR #1450, opened 2020-09-15
Fix Folly to build with current MSVC

### LINKED ISSUE #1449: Folly can't be built by current MSVC
MSVC added support for __builtin_clz etc back in 16.3 (MSC_VER=1923) but it is removing support in 16.8 (MSC_VER=1928).
Support for adding it was factored into Folly in this commit: https://github.com/facebook/folly/commit/6e58e89bf1a23f37c29ca8e4d1ef42b4e54408b0

Then this other commit noted that support will be taken out so that the intrinsics should be defined via functions: https://github.com/facebook/folly/commit/7f1bda25627503ca8af2536cc81298a7d4688375#diff-eb1a50189208c610a908adf1e1e44964

This means Folly will be able to be built with compilers >= 16.8 or 16.3, but 16.8 is not out yet, and we need to have folly be buildable on different VS versions.

The fix is to revert the change in that commit and change the condition to be <1923 || >=1928

asklar:
Fixes #1449 

The current public VS releases (16.3 up to 16.7) have builtins support. In commit https://github.com/facebook/folly/commit/7f1bda25627503ca8af2536cc81298a7d4688375 the #ifdef that was gating the definition of builtins was removed, as 16.8 will remove it. However by removign this gate, it makes current VS versions unable to build folly which breaks projects like react-native-windows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,10 @@ jobs:
   build:
     # The type of runner that the job will run on
     runs-on: windows-latest
+    strategy:
+      matrix:
+        arch: [ 'x86', 'x64', 'arm', 'arm64' ]
+        config: 'Debug'
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
@@ -25,7 +29,6 @@ jobs:
       - name: setup-msbuild
         uses: microsoft/setup-msbuild@v1.0.1
       
-      # Runs a single command using the runners shell
       - name: Build 
-        run: msbuild /p:Platform=x86,Configuration=Debug
+        run: msbuild /p:Platform=${{ arch }} /p:Configuration=${{ config }}
         working-directory: folly/build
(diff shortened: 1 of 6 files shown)


yfeldblum:
Thanks for the PR - looks legit, just a few requests.

Let's stick this new symbol into the `FOLLY_DETAIL_...` preprocessor-symbol namespace, always define it to either `0` or `1`, and `#undef` it at the bottom.

asklar:
@yfeldblum @Orvid done, I've also added a VS2019 build action to build a minimal portability test (i.e. we can call the builtins.h functions and things build).

Orvid:
The portability stuff should be split into another PR, and we really don't want to check in actual project files because they will go out of date very quickly. We support using Cmake to generate msbuild project files, so shouldn't need to check them in.

The base changes for this PR that get Folly building with the latest MSVC look good though.

PR #1455, opened 2020-09-15
<bit> produces warnings when building with C++17/13/11/...

### LINKED ISSUE #1454: Bits.h produces warning "The contents of <bit> are available only with C++20 or later."
https://github.com/facebook/folly/blob/3fdcf359fd1bd5f8a5acecd8d87205f0c7a17c82/folly/lang/Bits.h#L70-L72

Bits.h checks for __has_includes(<bit>) but it needs to also check `__cplusplus >= 202002L`

asklar:
Constant from: http://eel.is/c++draft/cpp.predefined#1.1

Fixes #1454

### DIFF
--- a/folly/lang/Bits.h
+++ b/folly/lang/Bits.h
@@ -67,7 +67,7 @@
 #include <folly/lang/Assume.h>
 #include <folly/portability/Builtins.h>
 
-#if __has_include(<bit>)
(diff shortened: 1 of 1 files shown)


asklar:
CC @Orvid

Orvid (approved):
Thanks for this :)

PR #1497, opened 2020-12-09
cmake: set `FOLLY_HAVE_LIBUNWIND` for cmake builds

mhx:
While there was a #cmakedefine for this in `folly-config.h.cmake`,
this was never defined as the variable was never set.


yash620:
I believe this was fixed in https://github.com/facebook/folly/commit/acef4a6449336c8757c87ef1cd463bfaba2690d9

mhx:
> I believe this was fixed in [acef4a6](https://github.com/facebook/folly/commit/acef4a6449336c8757c87ef1cd463bfaba2690d9)

It was indeed, albeit with unnecessary complexity. ;)

PR #1510, opened 2020-12-30
Fixing usage of FOLLY_SANITIZE_ADDRESS in folly/fibers

yash620:
`FOLLY_SANITIZE_ADDRESS` is used invalidly in these cases. This is causing ASAN build failures when using the library if the library is build without ASAN but the program using it is built with ASAN. Instead these should be calls to `FOLLY_LIBRARY_SANITIZE_ADDRESS`.


CMakeLists.txt
```diff
@@ -149,7 +149,6 @@ auto_sources(hfiles "*.h" "RECURSE" "${FOLLY_DIR}")
 REMOVE_MATCHES_FROM_LISTS(files hfiles
   MATCHES
     "^${FOLLY_DIR}/build/"
-    "^${FOLLY_DIR}/experimental/exception_tracer/"
```
yfeldblum, line 152: Is this delta intended to be part of this pull request?

yash620: Oh sorry that was unintentional. I got that from https://github.com/facebook/folly/pull/1484 which was needed to fix exception stack traces.

dutor:
@yfeldblum Is this PR ready to be merged?
I encountered the same problem when depending on folly as a prebuilt library.

yfeldblum:
This is a bit tricky since builds fail with this PR based on things being inconsistently guarded.

PR #1514, opened 2021-01-13
Make folly compile on Debian systems

chaiken:
The attached patches make Folly compile on Debian systems without affecting the build on Ubuntu 18.04.   They fix the error message

> folly/folly/portability/Time.cpp:278:2: error: No clock_gettime(3) compatibility wrapper available for this platform.
> #error No clock_gettime(3) compatibility wrapper available for this platform.

which suggested, apparently erroneously, that a compatibility wrapper is needed for Linux systems other than Ubuntu 18.04.   clock_gettime() is a Linux system call.

### DIFF
--- a/build/fbcode_builder/getdeps.py
+++ b/build/fbcode_builder/getdeps.py
@@ -745,7 +745,7 @@ class GenerateGitHubActionsCmd(ProjectCmdBase):
 
     def run_project_cmd(self, args, loader, manifest):
         platforms = [
-            HostType("linux", "ubuntu", "18"),
+            HostType("linux", None, None),
             HostType("darwin", None, None),
             HostType("windows", None, None),
         ]
--- a/folly/portability/Time.cpp
+++ b/folly/portability/Time.cpp
@@ -275,7 +275,9 @@ extern "C" int clock_gettime(clockid_t clock_id, struct timespec* tp) {
   }
 }
 #else
+#if !__linux__
 #error No clock_gettime(3) compatibility wrapper available for this platform.
+#endif
 #endif
 #endif
 


chaiken:
The test failures like "file_test.File.Simple" clearly have nothing to do with my patch, which only changes the Linux build and does not affect Mac or Windows at all.   Unfortunately I see no way to prove that assertion, so I'll close the pull request and carry the patch locally.

Orvid:
The test failures are fine, we just need the CLA signed before we can do anything with it.

chaiken:
On Linux localhost I find that EventBaseTest/EventBaseTest/0.RunInThread reproducibly fails but that everything else passes.     Also, there are 928 tests that the test frameworks reports are not found.   Presumably if they actually failed to compile or link, that would be reported as a build failure.

chaiken:
Another note: clock_gettime_wrappers_test works fine with this patch.   Assuredly it would fail if the patch were wrong.   Here is the result:

`$ /home/alison/gitsrc/fbcode-scratch-9Jan2021/build/folly/clock_gettime_wrappers_test 
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from ClockGettimeWrappers
[ RUN      ] ClockGettimeWrappers.clockGettimeWrapperIsWithin120SecsOfSystemClock
[       OK ] ClockGettimeWrappers.clockGettimeWrapperIsWithin120SecsOfSystemClock (0 ms)
[ RUN      ] ClockGettimeWrappers.clockGettimeNsWrapperIsWithin120SecsOfSystemClock
[       OK ] ClockGettimeWrappers.clockGettimeNsWrapperIsWithin120SecsOfSystemClock (0 ms)
[----------] 2 tests from ClockGettimeWrappers (0 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 2 tests.
`

build/fbcode_builder/getdeps.py
```diff
@@ -745,7 +745,7 @@ class GenerateGitHubActionsCmd(ProjectCmdBase):
 
     def run_project_cmd(self, args, loader, manifest):
         platforms = [
-            HostType("linux", "ubuntu", "18"),
+            HostType("linux", None, None),
```
Orvid, line 748: This needs to stay as ubuntu to run the github action on the right distro.

folly/portability/Time.cpp
```diff
@@ -275,9 +275,11 @@ extern "C" int clock_gettime(clockid_t clock_id, struct timespec* tp) {
   }
 }
 #else
+#if !__linux__
```
Orvid, line 278: What's causing `FOLLY_HAVE_CLOCK_GETTIME` to not be defined? This shouldn't be getting reached if that is defined properly.

Orvid (changes requested):

chaiken:
Replaced by https://github.com/facebook/folly/pull/1517.

PR #1517, opened 2021-01-25
Linux: fix clock_gettime and recvmmsg build problems

chaiken:
Replaces https://github.com/facebook/folly/pull/1514.  This more minimal change makes folly compile successfully by showing where clock_gettime() is found on Linux and solving the problem created by ee1e6c7b7219a545093e07b785efb5f396fac1ef when recvmmsg is not found.   Further changes to code in //folly/portability would be needed in order to allow Ninja (for example) to find these Linux syscalls at compile time.   Nonetheless, clock_gettime_wrappers_test passes.

### DIFF
--- a/CMake/FollyConfigChecks.cmake
+++ b/CMake/FollyConfigChecks.cmake
@@ -85,12 +85,12 @@ string(REGEX REPLACE
 
 check_symbol_exists(pthread_atfork pthread.h FOLLY_HAVE_PTHREAD_ATFORK)
 
-check_symbol_exists(accept4 sys/socket.h FOLLY_HAVE_ACCEPT4)
+check_symbol_exists(accept4 "sys/socket.h;linux/socket.h" FOLLY_HAVE_ACCEPT4)
 check_symbol_exists(getrandom sys/random.h FOLLY_HAVE_GETRANDOM)
 check_symbol_exists(preadv sys/uio.h FOLLY_HAVE_PREADV)
-check_symbol_exists(pwritev sys/uio.h FOLLY_HAVE_PWRITEV)
-check_symbol_exists(clock_gettime time.h FOLLY_HAVE_CLOCK_GETTIME)
-check_symbol_exists(pipe2 unistd.h FOLLY_HAVE_PIPE2)
+check_symbol_exists(pwritev sys/uio. FOLLY_HAVE_PWRITEV)
+check_symbol_exists(clock_gettime "time.h;linux/time.h" FOLLY_HAVE_CLOCK_GETTIME)
+check_symbol_exists(pipe2 "unistd.h;linux/unistd.h" FOLLY_HAVE_PIPE2)
 check_symbol_exists(sendmmsg sys/socket.h FOLLY_HAVE_SENDMMSG)
 check_symbol_exists(recvmmsg sys/socket.h FOLLY_HAVE_RECVMMSG)
 
--- a/folly/net/NetOps.cpp
+++ b/folly/net/NetOps.cpp
@@ -33,7 +33,9 @@
 #endif
 
 #if !FOLLY_HAVE_RECVMMSG
-#if FOLLY_HAVE_WEAK_SYMBOLS
+#if defined(__linux__)
+#include <linux/socket.h>
+#elif FOLLY_HAVE_WEAK_SYMBOLS
 extern "C" FOLLY_ATTR_WEAK int recvmmsg(
     int sockfd,
     struct mmsghdr* msgvec,
--- a/folly/portability/Time.cpp
+++ b/folly/portability/Time.cpp
@@ -274,6 +274,8 @@ extern "C" int clock_gettime(clockid_t clock_id, struct timespec* tp) {
       return -1;
   }
 }
+#elif defined(__linux__)
+#include <time.h>
 #else
 #error No clock_gettime(3) compatibility wrapper available for this platform.
 #endif


chaiken:
On my Linux localhost, the SmallLocks.MicroLockSlotsAsData passes.   Of course, the test result could depend on the kernel version.

> $ /home/alison/gitsrc/fbcode-scratch-24Jan2021/build/folly/small_locks_test 
> [==========] Running 17 tests from 2 test suites.
> [----------] Global test environment set-up.
> [----------] 16 tests from SmallLocks
> [ RUN      ] SmallLocks.SpinLockCorrectness
> [       OK ] SmallLocks.SpinLockCorrectness (7391 ms)
> [ RUN      ] SmallLocks.PicoSpinCorrectness
> [       OK ] SmallLocks.PicoSpinCorrectness (25 ms)
> [ RUN      ] SmallLocks.PicoSpinSigned
> [       OK ] SmallLocks.PicoSpinSigned (0 ms)
> [ RUN      ] SmallLocks.PicoSpinLockThreadSanitizer
> [  SKIPPED ] SmallLocks.PicoSpinLockThreadSanitizer (0 ms)
> [ RUN      ] SmallLocks.RegClobber
> [       OK ] SmallLocks.RegClobber (0 ms)
> [ RUN      ] SmallLocks.MicroLock
> [       OK ] SmallLocks.MicroLock (93 ms)
> [ RUN      ] SmallLocks.MicroLockTryLock
> [       OK ] SmallLocks.MicroLockTryLock (0 ms)
> [ RUN      ] SmallLocks.MicroLockSlotsAsData
> [       OK ] SmallLocks.MicroLockSlotsAsData (0 ms)
> [ RUN      ] SmallLocks.MicroSpinLockStressTestLockTwoThreads
> [       OK ] SmallLocks.MicroSpinLockStressTestLockTwoThreads (2000 ms)
> [ RUN      ] SmallLocks.MicroSpinLockStressTestLockHardwareConcurrency
> [       OK ] SmallLocks.MicroSpinLockStressTestLockHardwareConcurrency (2000 ms)
> [ RUN      ] SmallLocks.PicoSpinLockStressTestLockTwoThreads
> [       OK ] SmallLocks.PicoSpinLockStressTestLockTwoThreads (2001 ms)
> [ RUN      ] SmallLocks.PicoSpinLockStressTestLockHardwareConcurrency
> [       OK ] SmallLocks.PicoSpinLockStressTestLockHardwareConcurrency (2000 ms)
> [ RUN      ] SmallLocks.MicroSpinLockStressTestTryLockTwoThreads
> [       OK ] SmallLocks.MicroSpinLockStressTestTryLockTwoThreads (2000 ms)
> [ RUN      ] SmallLocks.MicroSpinLockStressTestTryLockHardwareConcurrency
> [       OK ] SmallLocks.MicroSpinLockStressTestTryLockHardwareConcurrency (2001 ms)
> [ RUN      ] SmallLocks.PicoSpinLockStressTestTryLockTwoThreads
> [       OK ] SmallLocks.PicoSpinLockStressTestTryLockTwoThreads (2000 ms)
> [ RUN      ] SmallLocks.PicoSpinLockStressTestTryLockHardwareConcurrency
> [       OK ] SmallLocks.PicoSpinLockStressTestTryLockHardwareConcurrency (2000 ms)
> [----------] 16 tests from SmallLocks (23511 ms total)
> 
> [----------] 1 test from SmallLocksk
> [ RUN      ] SmallLocksk.MicroSpinLockThreadSanitizer
> [  SKIPPED ] SmallLocksk.MicroSpinLockThreadSanitizer (0 ms)
> [----------] 1 test from SmallLocksk (0 ms total)
> 
> [----------] Global test environment tear-down
> [==========] 17 tests from 2 test suites ran. (23511 ms total)
> [  PASSED  ] 15 tests.
> [  SKIPPED ] 2 tests, listed below:
> [  SKIPPED ] SmallLocks.PicoSpinLockThreadSanitizer
> [  SKIPPED ] SmallLocksk.MicroSpinLockThreadSanitizer

chaiken:
I ran an the unit tests here on this system; only Random.SecureFork fails.   36 tests are skipped, mostly because more code is needed in folly/Portability.h so that cmake finds Linux's pthread_atfork() implementation.

Orvid:
The .cpp changes here should have no affect, and the config changes don't make sense. If `clock_gettime` is not accessible via `time.h`, `pipe2` is not in `unistd.h`, and `accept4` is not in `sys/sockets.h` then I believe something has gone wrong with your linux install, or else something has failed to add the root include path when building Folly.

chaiken:
> The .cpp changes here should have no affect, and the config changes don't make sense. If `clock_gettime` is not accessible via `time.h`, `pipe2` is not in `unistd.h`, and `accept4` is not in `sys/sockets.h` then I believe something has gone wrong with your linux install, or else something has failed to add the root include path when building Folly.

The Debian Testing release has much newer packages than Ubuntu 18.04.   In particular, Ubuntu 18.04 includes linux-libc-dev version 4.15.0-135.139, while Debian ships linux-libc-dev version 5.10.12-1.   The implementation of clock_gettime in 5.10.12-1 fixes the "Y2038 problem":
https://sourceware.org/pipermail/libc-alpha/2020-May/113643.html
The consequence is that the implementation of clock_gettime() has changed.   Perhaps either portability/Time.cpp or CMake/FollyConfigChecks.cmake needs logic that detects the libc version and then chooses the right headers.   Undoubtedly the same problem will occur with newer versions of Ubuntu that include the newer libc.

The problems with the other system calls are likely similar, but   De I have not had a chance to analyze them yet.   Debian also has newer versions of Cmake and boost than Ubuntu which might require changes to detection scripts.

CMake/FollyConfigChecks.cmake
```diff
@@ -85,12 +85,12 @@ string(REGEX REPLACE
 
 check_symbol_exists(pthread_atfork pthread.h FOLLY_HAVE_PTHREAD_ATFORK)
 
-check_symbol_exists(accept4 sys/socket.h FOLLY_HAVE_ACCEPT4)
+check_symbol_exists(accept4 "sys/socket.h;linux/socket.h" FOLLY_HAVE_ACCEPT4)
 check_symbol_exists(getrandom sys/random.h FOLLY_HAVE_GETRANDOM)
 check_symbol_exists(preadv sys/uio.h FOLLY_HAVE_PREADV)
-check_symbol_exists(pwritev sys/uio.h FOLLY_HAVE_PWRITEV)
```
yfeldblum, line 91: Was this a typo?

folly/portability/Time.cpp
```diff
@@ -274,6 +274,8 @@ extern "C" int clock_gettime(clockid_t clock_id, struct timespec* tp) {
       return -1;
   }
 }
+#elif defined(__linux__)
+#include <time.h>
```
yfeldblum, line 278: What is this for? The include is not used below.

folly/net/NetOps.cpp
```diff
@@ -33,7 +33,9 @@
 #endif
 
 #if !FOLLY_HAVE_RECVMMSG
-#if FOLLY_HAVE_WEAK_SYMBOLS
+#if defined(__linux__)
+#include <linux/socket.h>
```
yfeldblum, line 37: Why is the include necessary? Presumably weak symbols are available.

CMake/FollyConfigChecks.cmake
chaiken: Thanks for looking at that existing patch version.   As suggested above, the right solution is probably to have a new macro that detects whether linux-libc-dev is version 4.x or 5.x, and to use it to fix the clock_gettime, pthread_atfork, pipe2 and accept4 all with one test.    I have not had time yet to analyze whether all 4 system calls changed between libc-header versions.  What puzzles me is that the boost part of the build appears to run its own separate detection step, so that any fix would have to be made in two cmake scripts.   Is that correct?

folly/portability/Time.cpp
rcari: Moreover, it's already included in the `<Time.h>` header in `folly/portability`.

PR #1526, opened 2021-02-16
Remove include to log

efiks:
Summary: Remove external dependency to be able  https://godbolt.org/z/f7nWvs

Differential Revision: D26460959


yfeldblum (changes requested):
The file uses `DCHECK`, which comes from glog.

Orvid:
Internal diff abandoned.

PR #1535, opened 2021-03-06
fix linking of StaticTracepointTest

chaiken:
The placement of a macro definition inside a namespace led to errors
like

/usr/bin/ld: /usr/bin/ld: DWARF error: could not find variable specification at offset 51

The test passes once built.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -951,6 +951,9 @@ if (BUILD_TESTS)
       TEST unit_test SOURCES UnitTest.cpp
       TEST uri_test SOURCES UriTest.cpp
       TEST varint_test SOURCES VarintTest.cpp
+
+    DIRECTORY tracing/test/
+      TEST static_tracepoint_test SOURCES StaticTracepointTest.cpp StaticTracepointTestModule.cpp
   )
 
   if (${LIBSODIUM_FOUND})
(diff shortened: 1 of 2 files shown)


yfeldblum:
Curious. We do use the macro within C++ namespaces. Does the choice of linker make a difference here?

chaiken:
> Curious. We do use the macro within C++ namespaces. Does the choice of linker make a difference here?

Yes.  I find that the build runs to completion without the patch with g++ 10.2.1, but fails to link with clang 11.0.1.  I'll submit another version that uses preprocessor __GNUC__ and __clang__ macros to incorporate the DEFINE_SEMAPHORE macro in one position or the other if doing so would be helpful.

chaiken:
Sigh, the macros should be `__GNUC__` and `__clang__` above.

yfeldblum:
Is this a bug in clang++? Can it be fixed upstream?

Or is this just fundamentally something that won't work? Will we have to move the tracepoint semaphores we have that are within C++ namespaces?

chaiken:
> Is this a bug in clang++? Can it be fixed upstream?

The problem was a bug in binutils that is now fixed upstream: 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=965247

clang version 11.0.1-2 apparently incorporates the fix and will compile the code.   I'll submit a new version that only turns on the test.   Thanks for your patience while I figured this out!

PR #1542, opened 2021-03-19
Fix Range.h to compile in c++20

asklar:
`std::is_pod` is deprecated in c++20, using `is_trivially_copyable` instead

### DIFF
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -474,7 +474,7 @@ class Range {
   template <typename ValueType>
   struct StringViewType
       : std::conditional<
-            std::is_pod<std::remove_const_t<ValueType>>::value,
+            std::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
             std::basic_string_view<std::remove_const_t<ValueType>>,
             NotStringView> {};
 
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -474,7 +474,7 @@ class Range {
   template <typename ValueType>
   struct StringViewType
       : std::conditional<
-            std::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
+            folly::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
             std::basic_string_view<std::remove_const_t<ValueType>>,
             NotStringView> {};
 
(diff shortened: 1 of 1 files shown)


folly/Range.h
```diff
@@ -474,7 +474,7 @@ class Range {
   template <typename ValueType>
   struct StringViewType
       : std::conditional<
-            std::is_pod<std::remove_const_t<ValueType>>::value,
+            std::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
```
yfeldblum, line 477: Need to use `folly::is_trivially_copyable` here.

yfeldblum (changes requested):
Seems like one tweak is necessary.

folly/Range.h
asklar:
```suggestion
            folly::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
```

folly/Range.h
```diff
@@ -474,7 +474,7 @@ class Range {
   template <typename ValueType>
   struct StringViewType
       : std::conditional<
-            std::is_pod<std::remove_const_t<ValueType>>::value,
+            folly::is_trivially_copyable<std::remove_const_t<ValueType>>::value,
```
yfeldblum, line 477: This change has had a chance to run on internal CI. There, it is observed to cause a large number of build failures.

In practice, libstdc++ `std::basic_string_view` has a static-assert that the type parameter is trivial. We use this test to choose whether conversion to `std::basic_string_view` is available, so if the condition here is too broad, the test can allow conversion to an instance of `std::basic_string_view` which immediately fails a static-assert.

Can we switch this to `std::is_trivial`?

yfeldblum (changes requested):

folly/Range.h
asklar: thanks for the heads up; yeah let's try is_trivial, possibly combined with is_standard_layout; 
I think the original trait was probably not right either as it allows the elements in the string view to be arrays which seems wrong

```suggestion
            std::is_trivial<std::remove_const_t<ValueType>>::value,
```

PR #1546, opened 2021-03-27
compile StaticTracepointTest

chaiken:
The test was accidentally omitted from CMakeLists.


chaiken:
Successor to closed PR #1535.

chaiken:
Even if the unchecked pointer dereference is not the cause of the segfault, it's still a bad idea.   Obviously checking (nullptr == get_r_debug()) would also work, the value should be non-null if the macro is TRUE.

yfeldblum:
Seems like a macOS build failure?

chaiken:
> Seems like a macOS build failure?

Apparently on macOS the build cannot find the StaticTracepoint.h header.   I have no access to a macOS system to try to reproduce the failure.    The failure is a bit odd since the macro comes almost directly from folly/experimental/symbolizer/detail/Debug.h.   Maybe I need 

`#if defined(__APPLE__) && FOLLY_HAVE_ELF && FOLLY_HAVE_DWARF  || defined(__linux__)  && FOLLY_HAVE_ELF && FOLLY_HAVE_DWARF `

? But that is guessing, as I have no way to test the idea besides pushing a new version.

yfeldblum:
> The test was accidentally omitted from CMakeLists.

We do not have all of the tests in the cmake build. Some are accidentally omitted. Others do not build as-is on all platforms. E.g., the test source uses `FOLLY_SDT_NOTE_NAME` which exists only on some platforms. And the test source uses `struct r_debug`, but it is only complete on some platforms.

We can add the test to the cmake build if we mark it excluded from the platforms where it would give us trouble.

Also, prefer checking `!= nullptr` v.s. using the macros.

chaiken:
The macro could still just be omitted, I guess, or the existence of /usr/include/link.h could be checked.   Of course even if it exists, it could have irrelevant content.

CMakeLists.txt
```diff
@@ -952,6 +952,10 @@ if (BUILD_TESTS)
       TEST unit_test SOURCES UnitTest.cpp
       TEST uri_test SOURCES UriTest.cpp
       TEST varint_test SOURCES VarintTest.cpp
+
+    DIRECTORY tracing/test/
+      #Only meaningful on Linux with its bpf() system call
+      #TEST static_tracepoint_test SOURCES StaticTracepointTest.cpp StaticTracepointTestModule.cpp
```
yfeldblum, line 958: Possible now to enable this but to mark this `WINDOWS_DISABLED APPLE_DISABLED`?

chaiken: Just rebased on master and found APPLED_DISABLED.   Oddly, 

WINDOWS_DISABLED APPLE_DISABLED

results in an error saying that WINDOWS_DISABLED is not defined, but 

APPLE_DISABLED WINDOWS_DISABLED 

works.    Also, the comment appeared to cause a problem.

chaiken:
@yfeldblum is this patch now okay?

Issue #1550, opened 2021-03-31
new fmt dependency reduces usability of folly


asklar:
The last time React Native for Windows took an ingestion of Folly was Oct 5 2020.
On Oct 26, a new dependency was added in this commit:  Add a StringPiece formatter · facebook/folly@6d95d30 
Instructions were added sometime in November: Add instructions for installiing fmt from source by SashaNullptr · #1263 

However there are a few problems:
- There's no specification of what version of fmt corresponds to each Folly version. Are any combinations of folly/fmt supposed to be compatible? is this tested?
- The addition of fmt as a dependency makes usage of folly more difficult (need to keep up to date with yet another library, look out for vulnerabilities, there's more code that can break with compiler updates, etc.)

Is there a way to make the fmt dependency optional?

CC @yfeldblum @Orvid

Orvid:
Which compilers & runtimes do you need to support? 

As far as I know we're on the way towards fully replacing all uses of folly::format with fmt as it is strictly better in terms of compile time, runtime perf, and binary size, which is fairly important for iOS and Android users of Folly.

@vitaut Do you think it would be viable to support an interface in folly that forwards to either std::format or fmt

asklar:
We use MSVC. 

We currently use only parts of folly (only as much as is needed to support [React Native for Windows](https://github.com/microsoft/react-native-windows/)). Note that there are parts of folly that do not work UWP on Windows altogether (#1452) so we can't take all of it even if we wanted to :)

We want to limit our exposure to breakages coming from both msvc compiler changes (which we've seen a good amount), as well as from folly not being exercised/built for Windows and 'bit-rotting' over time (#1456).

asklar:
For what it's worth, it doesn't seem like anything in RN (or at least the way we are consuming it), actually relies on fmt, which is why I was able to just gate the include in the PR on whether the include exists (#1551)

EtoDemerzel0427:
What is the `fmt`'s version Folly uses right now? I am using Folly/2022.01.31.00 (because it is the newest version available on Conan center), but it relies on `fmt` 7.x and has a conflict with my `spdlog`.

vitaut:
fmt 8.0.1: https://github.com/facebook/folly/blob/93c44c56cda1c57e48fb42c30aff55acf2436814/build/fbcode_builder/manifests/fmt#L5

yfeldblum:
Sorry for the trouble. We do have very limited dependencies and we try to add dependencies only very rarely.

PR #1551, opened 2021-03-31
Conditionally include fmt/format.h

asklar:
Addresses - at least partly - issue #1550 

We'd like to limit our exposure to additional libraries. Fmt is a new dependency. It doesn't look like we need it in actuality; gating the include based on whether the file exists or not makes it possible for clients (specifically React Native for Windows) to not need to fork Folly.

### DIFF
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -41,7 +41,9 @@
 #include <string_view> // @manual
 #endif
 
+#if __has_include(<fmt/format.h>)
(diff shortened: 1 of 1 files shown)


asklar:
Writing up some of the challenges with consuming folly in RN and some questions for RN and folly folks here:
https://gist.github.com/asklar/12069510eb609f5575b9471b70f495d2

Issue #1558, opened 2021-04-12
Deadlock with SharedMutex across multiple DSOs and visibility=hidden


daverigby:
### Summary

When accessing a {{folly::SharedMutex}} instance from multiple DSOs and compiled with {{-fvisibility=hidden}}, a deadlock is seen when at least one thread is attempting to acquire a write lock. Backtraces suggest that the writer is waiting for some outstanding reader to unlock; but no such reader appears to have the lock acquired.

* Version: v2010.08.24.00 (but nothing relevant seems to have changed recently so I suspect master is also affected - will confirm shortly).
* Reproducer at: https://github.com/daverigby/folly_SharedMutexIssue

Note: this is a similar issue to #1431.

### Details

From my limited understanding of {{SharedMutex}}, what seems to be occurring is that there's multiple copies of the {{SharedMutex<>::deferredReaders}} static array; and as such the tracking of which readers are outstanding is broken. I believe this is due to {{deferredReaders}} having local (hidden) visibility - for example examining the two DSOs in my test program, note the LOCAL visibility:

```
$ readelf -sW main | c++filt | grep 'folly::SharedMutexImpl<false, void.*deferredReaders' 
   483: 0000000000625400 16384 OBJECT  LOCAL  DEFAULT   28 folly::SharedMutexImpl<false, void, std::atomic, false, false>::deferredReaders
$ readelf -sW libshared.so | c++filt | grep 'folly::SharedMutexImpl<false, void.*deferredReaders' 
   434: 000000000022c500 16384 OBJECT  LOCAL  DEFAULT   26 folly::SharedMutexImpl<false, void, std::atomic, false, false>::deferredReaders
```

If I instantiate SharedMutex with it's own tag:
```
struct FOLLY_EXPORT MyTag {};
using Mutex = folly::SharedMutexImpl<false, MyTag>;
```

 _and_ modify the declaration of {{SharedMutexImpl}} to mark it with {{FOLLY_EXPORT}} the test doesn't hang (run for > 1 hour with no issues):

```
diff --git a/folly/SharedMutex.h b/folly/SharedMutex.h
index 95faa76f5..44102168b 100644
--- a/folly/SharedMutex.h
+++ b/folly/SharedMutex.h
@@ -330,3 +330,3 @@ template <
     bool TrackThreadId = false>
-class SharedMutexImpl : std::conditional_t<
+class FOLLY_EXPORT SharedMutexImpl : std::conditional_t<
                             TrackThreadId,
```

[...]

PR #1559, opened 2021-04-13
Mark SharedMutexImpl with FOLLY_EXPORT (#1558)

daverigby:
When accessing a folly::SharedMutex instance from multiple DSOs and
program and folly compiled with -fvisibility=hidden, a deadlock is
seen when at least one thread is attempting to acquire a write
lock.

Backtraces suggest that the writer is waiting for some outstanding
reader to unlock; but no such reader appears to have the lock
acquired. In actual fact there's multiple copies of the (supposedly
singleton) deferredReaders array, one per DSO, it has hidden (local)
visibility.  This essentially corrupts the details of which readers
are outstanding.

Fix by flagging SharedMutexImpl as FOLLY_EXPORT - meaning it has weak
linkage and hence only a single instance will be present in memory at
runtime.

Fixes: https://github.com/facebook/folly/issues/1558

### DIFF
--- a/folly/SharedMutex.h
+++ b/folly/SharedMutex.h
@@ -328,7 +328,7 @@ template <
     bool BlockImmediately = false,
     bool AnnotateForThreadSanitizer = kIsSanitizeThread && !ReaderPriority,
     bool TrackThreadId = false>
-class SharedMutexImpl : std::conditional_t<
+class FOLLY_EXPORT SharedMutexImpl : std::conditional_t<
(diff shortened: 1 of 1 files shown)


yfeldblum:
Is this still a problem? After 5d1b9736e1d3b0bfb339ccea665d3be0e8efab65 in particular.

daverigby:
Stale - closing (I'm no longer working on a project depending on Folly).

PR #1564, opened 2021-04-21
Remove sysfs cachelocality when compiling for WIN32

acoates-ms:
sysfs doesn't exist on windows, so these functions do not make sense.  

This was hit while bringing in a bunch of the fabric code from react-native into react-native-windows.

### DIFF
--- a/folly/concurrency/CacheLocality.cpp
+++ b/folly/concurrency/CacheLocality.cpp
@@ -105,6 +105,7 @@ static size_t parseLeadingNumber(const std::string& line) {
   return val;
 }
 
+#ifndef _WIN32
 CacheLocality CacheLocality::readFromSysfsTree(
     const std::function<std::string(std::string)>& mapping) {
   // number of equivalence classes per level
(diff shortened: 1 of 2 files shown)


yfeldblum:
Can we do something similar, but more C++ish?

```
if (kIsWindows) {
  return {};
}
```

acoates-ms:
Thus far react-native-windows has not taken a dependency on fmt.  Using the #ifdef we avoid the call to fmt.  With the if (kIsWindows) we'd have to add the fmt dependency for some code that will never be used on windows.

Other options would be to move the method somewhere other than a static within CacheLocality so that we dont include the functionality at all.  -- But that would presumably involve changing call sites within FB so I was avoiding that.

terrelln:
> Thus far react-native-windows has not taken a dependency on fmt. Using the #ifdef we avoid the call to fmt. With the if (kIsWindows) we'd have to add the fmt dependency for some code that will never be used on windows.

Folly depends on `fmt` and I would only expect usage of `fmt` within `folly` to increase over time. And sooner or later some component you do use will use `fmt`. E.g. in formatting an exception message. I would recommend adding the `fmt` dependency.

acoates-ms:
That fair enough.  We try to avoid bringing in additional dependencies, but it does seem like we're going to have to take `fmt` sooner or later.  I just brought it into `react-native-windows`, so we no longer need this change.

PR #1565, opened 2021-04-22
Fix several C++ warnings

AgneLukoseviciute:
Fixes warnings about implicit type truncation. By making the conversions explicit, react-native-windows will be able to remove suppressions for several warnings required per SDL.

### DIFF
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -56,17 +56,23 @@ using to_ascii_alphabet_upper = to_ascii_alphabet<true>;
 
 namespace detail {
 
+#if defined(_M_IX86)
+FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
+  return __builtin_clzll(v);
+}
+#else
 FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
 #if _MSC_VER
 #if FOLLY_X64
   return __lzcnt64(v);
 #else
-  return __assume(0), 0;
+  __assume(0), 0;
 #endif
 #else
   return __builtin_clzll(v);
 #endif
 }
+#endif
 
 template <uint64_t Base, typename Alphabet>
 struct to_ascii_array {
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -59,7 +59,7 @@ namespace detail {
 FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
 #if _MSC_VER
 #if FOLLY_X64
-  return __lzcnt64(v);
+  return static_cast<int>(__lzcnt64(v));
 #else
   return __assume(0), 0;
 #endif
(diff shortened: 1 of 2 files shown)


folly/lang/ToAscii.h
```diff
@@ -59,7 +59,7 @@ namespace detail {
 FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
 #if _MSC_VER
 #if FOLLY_X64
-  return __lzcnt64(v);
+  return static_cast<int>(__lzcnt64(v));
```
yfeldblum, line 62: Let's move this cast into the caller, since the return type of this function is intended to be the return type of the intrinsic or the builtin which it calls.

AgneLukoseviciute: ahh yes, I didn't notice the return type had changed to auto. Thanks!

AgneLukoseviciute: I tested it locally and looks like the 'auto' keyword takes care of the warning for us, so I will just remove this change altogether.

AgneLukoseviciute: @yfeldblum we're getting some undefined behavior with x86 using auto return type here, this method returns void so casting in the caller is not enough. Any insight for the purpose of line 63-64? Is this expected to never be called from x86?

yfeldblum: The type of the expression `__assume(0), 0` should be `int`.

folly/lang/ToAscii.h
```diff
@@ -56,17 +56,23 @@ using to_ascii_alphabet_upper = to_ascii_alphabet<true>;
 
 namespace detail {
 
+#if defined(_M_IX86)
```
yfeldblum, line 59: This is way more complicated than I would like.

IIUC, only MSVC has the property where the 64-bit intrinsic `__lzcnt64` is unavailable under 32-bit builds. For GNU-compatible compilers, the 64-bit builtin `__builtin_clzll` is implemented natively under 64-bit builds and emulated under 32-bit builds.

yfeldblum (changes requested):

AgneLukoseviciute:
@yfeldblum This looks good, it will take care of our warning suppressions. I'm still working on getting some clarity on to_ascii_port_clzll() support for x86, if a change is needed we can address that separately : ) Thanks for taking a look!

folly/lang/ToAscii.h
```diff
@@ -224,7 +224,7 @@ FOLLY_ALWAYS_INLINE size_t to_ascii_size_clzll(uint64_t v) {
   }
 
   //  log2 is approx log<2>(v)
-  size_t const vlog2 = 64 - to_ascii_port_clzll(v);
+  size_t const vlog2 = 64 - static_cast<int>(to_ascii_port_clzll(v));
```
yfeldblum, line 227: Seems like this explicit cast could be to `size_t`. That would avoid any issue of signed-to-unsigned conversion.

PR #1567, opened 2021-04-30
Adding x86 support for to_ascii_port_clzll() 

### LINKED ISSUE #1566: to_ascii_port_clzll() does not support x86
https://github.com/facebook/folly/blob/9552fbcd6540b9908d657ee2edde666680e21c04/folly/lang/ToAscii.h#L64

to_ascii_port_clzll() in ToAscii.h always returns 0 in x86. We should be calling __builtin_clzll() for x86 instead which will allow us at React Native for Windows to remove a fork for this file with the [fix](https://github.com/microsoft/react-native-windows/commit/29481ff5a5e93785e47735b8bf67ba6a3f911a92) for x86.

AgneLukoseviciute:
Currently to_ascii_port_clzll() in ToAscii.h always returns 0 in x86. This adds a check for x86 and a call to __builtin_clzll() instead. 

Closes #1566.

@yfeldblum @Orvid

### DIFF
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -60,6 +60,8 @@ FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
 #if _MSC_VER
 #if FOLLY_X64
   return __lzcnt64(v);
+#elif defined(_M_IX86)
+  return __builtin_clzll(v);
 #else
   return __assume(0), 0;
 #endif
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -24,6 +24,7 @@
 #include <folly/Utility.h>
 #include <folly/lang/Align.h>
 #include <folly/lang/CArray.h>
+#include <folly/portability/builtins.h>
 
 #if _MSC_VER
 #include <intrin.h>
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -56,20 +56,6 @@ using to_ascii_alphabet_upper = to_ascii_alphabet<true>;
 
 namespace detail {
 
-FOLLY_ERASE auto to_ascii_port_clzll(uint64_t v) {
-#if _MSC_VER
-#if FOLLY_X64
-  return __lzcnt64(v);
-#elif defined(_M_IX86)
-  return __builtin_clzll(v);
-#else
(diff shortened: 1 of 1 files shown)


yfeldblum:
GNU-compatible compilers (gcc, clang) have `__builtin_clzll`.

I didn't find any documentation that MSVC has it, though.

AgneLukoseviciute:
> GNU-compatible compilers (gcc, clang) have `__builtin_clzll`.
> 
> I didn't find any documentation that MSVC has it, though.

`builtin_clz` was included only in MSVC 16.3 through 16.8, #1450 added a check for this https://github.com/facebook/folly/blob/4e7efffdcadf373d1ae7d17722634d3a849c1f7a/folly/portability/Builtins.h#L29

in the case when MSVC does not have it, I believe `builtin_clzzll` will call the polyfill `builtin_clz` implementation that uses `_BitScanReverse` https://github.com/facebook/folly/blob/4e7efffdcadf373d1ae7d17722634d3a849c1f7a/folly/portability/Builtins.h#L55

yfeldblum:
Ah, in that case, the whole `_port` function can be removed.

AgneLukoseviciute:
> Ah, in that case, the whole `_port` function can be removed.

delete to_ascii_port_clzll()? -- it gets called on line 227: 

https://github.com/facebook/folly/blob/93d52d898521a16a1fe59a575f052a534c6715ec/folly/lang/ToAscii.h#L227

and then calls __lzcnt64 or __builtin_clzll or returns 0 depending on the macros.

yfeldblum:
Seems like the caller can just call `__builtin_clzll` directly with the builtins portability header included.

folly/lang/ToAscii.h
```diff
@@ -226,7 +212,7 @@ FOLLY_ALWAYS_INLINE size_t to_ascii_size_clzll(uint64_t v) {
   }
 
   //  log2 is approx log<2>(v)
-  size_t const vlog2 = 64 - static_cast<size_t>(to_ascii_port_clzll(v));
+  size_t const vlog2 = 64 - static_cast<size_t>(__builtin_clzll(v));
```
yfeldblum, line 215: Presumably need to add the `#include` above.

AgneLukoseviciute: oops. Should be good now :)

PR #1580, opened 2021-05-16
Correcting and adding a coarse_ * clock

Roman-Koshelev:


Roman-Koshelev:
"Thank you!
Thank you for taking the time to register as an open source contributor. You may print the submission below for your records. You can resubmit this form to update your name or email. If you would like to make any other changes, please email cla@fb.com with your details.

We will now be able to accept any pull requests that you make to projects on our GitHub account. If you already submitted a pull request before signing the CLA, then we will add a confirmation comment on that pull request (this might take several minutes). If you haven't submitted a pull request yet, then we look forward to your future contributions!"

folly/Chrono.h
```diff
@@ -175,28 +175,57 @@ namespace folly {
...
 
 struct coarse_steady_clock {
-  using rep = std::chrono::milliseconds::rep;
-  using period = std::chrono::milliseconds::period;
-  using duration = std::chrono::duration<rep, period>;
-  using time_point = std::chrono::time_point<coarse_steady_clock, duration>;
+  using duration = std::chrono::milliseconds;
+  using rep = duration::rep;
+  using period = duration::period;
+  using time_point = std::chrono::time_point<std::chrono::steady_clock, duration>;
```
yfeldblum, line 181: It seems odd for `some_clock::time_point::clock` to be different from `some_clock`. Do we have any examples of this being general practice?

Roman-Koshelev: I've been thinking about it for a long time. This corresponds to the idea of the chrono library. If there is a guarantee that *_COARCE watches are the same watches as their fine-grained counterparts, then they simply must have time points pamametrized by one type of watch. It would be strange to use time_point_cast for example to compare the time points of the same clock.

yfeldblum: The question is whether these are the same clock even when they're different types. It makes more sense to me from a low-level perspective to treat them as different clocks but then to add a concept of compatible sets of clocks, where, within a compatible set of clocks, time-points of one clock are convertible to equivalent time-points of another clock. That way it remains possible to distinguish where it is necessary to distinguish and to treat as equivalent where equivalence makes sense.

Roman-Koshelev: This concept should be introduced in the c++standard. In the meantime, what do we do if we just want to compare the time of these two hours? (except to compare their time_since_epoch)

yfeldblum: Perhaps it should exist in the standard. It should help to have an implementation when suggesting for something to exist in the standard though.

In the meantime, the concept of clock compatibility sets, with corresponding conversion functions, can exist in folly since the clocks being proposed are in folly.

Roman-Koshelev: I agree. I removed the commit with this change from review

folly/Chrono.h
```diff
@@ -175,28 +175,57 @@ namespace folly {
 namespace chrono {
 
 struct coarse_steady_clock {
-  using rep = std::chrono::milliseconds::rep;
-  using period = std::chrono::milliseconds::period;
-  using duration = std::chrono::duration<rep, period>;
-  using time_point = std::chrono::time_point<coarse_steady_clock, duration>;
+  using duration = std::chrono::milliseconds;
```
yfeldblum, line 178: Stepping back, I'm curious why we want this to be `milliseconds` rather than alias `steady_clock::duration`.

To use `milliseconds`, we must round (`duration_cast`). Is this a necessity?

Roman-Koshelev: As far as I remember, once upon a time, `struct timespec res;
    clock_getres(CLOCK_MONOTONIC_COARSE, &res);` returned an accuracy of 1 ms (on my system), and I always thought that it was. Right now, on my system, the accuracy is 4 ms (apparently due to CONFIG_HZ = 250)

Roman-Koshelev: In any case, this is a less accurate watch. It would be strange if they had the same accuracy. I think most systems will have> = 1ms precision for this clock. It's okay if they are somewhere more accurate (for example, libc ++ has 1us precision on the system clock for the sake of maintaining compatibility with abi)

Roman-Koshelev: I found out that the possible values of HZ are 100, 250, 300, 1000. Those the maximum accuracy at the current moment is 1ms

Roman-Koshelev:
Is there any further discussion on Phabricator? Can i see him?

Issue #1279, opened 2019-12-13
ndk build folly  use of undeclared identifier 'getdtablesize'


stevezhou6:
folly/Subprocess.cpp:542:19: error: use of undeclared identifier 'getdtablesize'; did you mean 'getpagesize'?
    for (int fd = getdtablesize() - 1; fd >= 3; --fd) {

Issue #918, opened 2018-08-27
getdtablesize was removed from Android NDK and causing Android build to break


OleksiyA:
See bug for details https://b.corp.google.com/issues/37001193

One possible fix is to replace with sysconf(_SC_OPEN_MAX).

PR #1584, opened 2021-05-18
Fix table size for android

karol-bisztyga:
## Description

Hi!
I wanted to fix a build of folly on Android.

Trying to use folly's `MPMCQueue` on Android triggers the following error:
```
{xxx}/rnfbjni/android/app/build/third-party-ndk/folly/folly/Subprocess.cpp:542:19: error: use of undeclared identifier 'getdtablesize'; did you mean 'getpagesize'?
```
where `{xxx}` is a local path on my machine.

[Here](https://github.com/karol-bisztyga/rnfbjni/tree/patch-folly-repro) is a branch of my project that reproduces the problem(I've already fixed it by patching `folly` so you can see what patch has to be removed in order to make it fail [here](https://github.com/karol-bisztyga/rnfbjni/commit/e6d8059e55efda5f88d5edbc1e932b94223b052c)).

From [the man page for `getdtablesize`](https://linux.die.net/man/2/getdtablesize):

> SVr4, 4.4BSD (the getdtablesize() function first appeared in 4.2BSD). It is not specified in POSIX.1-2001; portable applications should employ sysconf(_SC_OPEN_MAX) instead of this call.

### Related issues:
https://github.com/facebook/folly/issues/1279
https://github.com/facebook/folly/issues/918 - probably, I cannot see bug details.

### DIFF
--- a/folly/Subprocess.cpp
+++ b/folly/Subprocess.cpp
@@ -560,8 +560,13 @@ int Subprocess::prepareChild(
   // If requested, close all other file descriptors.  Don't close
   // any fds in options.fdActions_, and don't touch stdin, stdout, stderr.
   // Ignore errors.
+  #ifdef __ANDROID__
+  int tableSize = sysconf(_SC_OPEN_MAX);
+  #else
+  int tableSize = getdtablesize();
+  #endif
   if (options.closeOtherFds_) {
-    for (int fd = getdtablesize() - 1; fd >= 3; --fd) {
+    for (int fd = tableSize - 1; fd >= 3; --fd) {
       if (options.fdActions_.count(fd) == 0) {
         ::close(fd);
       }


folly/Subprocess.cpp
```diff
@@ -560,8 +560,13 @@ int Subprocess::prepareChild(
   // If requested, close all other file descriptors.  Don't close
   // any fds in options.fdActions_, and don't touch stdin, stdout, stderr.
   // Ignore errors.
+  #ifdef __ANDROID__
+  int tableSize = sysconf(_SC_OPEN_MAX);
+  #else
+  int tableSize = getdtablesize();
```
yfeldblum, line 566: Any particular reason to keep the call to `getdtablesize` on non-Android platforms, seeing as it's less portable?

karol-bisztyga: I've never really tested it on **all** other platforms - I just checked android and it works on it. To change it for every platform I should check every one of them, don't you think?

PR #1588, opened 2021-05-25
Windows: Fixing compile-time warning 

AgneLukoseviciute:
We are [hitting a compile-time warning](https://github.com/microsoft/react-native-windows/pull/7830) that is treated as an error in RNW CI build for x64 -- warning C4127 due to a compile-time constant being used in a conditional expression. This replaces an `if`  with an `if constexpr` in lang/ToAscii.h to avoid the warning.

### DIFF
--- a/folly/lang/ToAscii.h
+++ b/folly/lang/ToAscii.h
@@ -212,7 +212,7 @@ FOLLY_ALWAYS_INLINE size_t to_ascii_size_clzll(uint64_t v) {
   size_t const vlog2 = 64 - static_cast<size_t>(__builtin_clzll(v));
 
   //  handle directly when Base is power-of-two
-  if (!(Base & (Base - 1))) {
+  if constexpr(!(Base & (Base - 1))) {
     constexpr auto const blog2 = constexpr_log2(Base);
     return vlog2 / blog2 + size_t(vlog2 % blog2 != 0);
   }


yfeldblum:
We have encountered this warning as well.

The warning strikes me as problematic for two reasons:
1. It adds a warning on perfectly-correct C++14 code because C++17 has a new feature. This makes it impossible for normal C++ library code to support both C++14 and C++17 without cumbersome workarounds. We still need `folly::to_ascii` to support C++14.
2. `if constexpr` has different semantics from, and asks different things of the compiler than does, `if`. Sometimes it is better; sometimes worse.

We have disabled this warning internally. I'm not sure whether we would enable it even once everything is at or past C++17.

It is possible to address this instance of the warning in practice by adding in a `|| expr` where `expr` constant-folds to zero but which is technically not `constexpr`.

yfeldblum:
Can you let me know whether #1590 solves this problem?

AgneLukoseviciute:
> Can you let me know whether #1590 solves this problem?

Unfortunately, I'm still hitting the warning with the change in #1590

yfeldblum:
Any luck with that PR but with `const` removed? Just pushed that change to the PR.

AgneLukoseviciute:
@yfeldblum Yeah, that works. Thanks!

PR #1590, opened 2021-05-25
work around msvc warning C4127 in to_ascii

yfeldblum:
Summary:
MSVC warning C4127 (conditional expression is constant) goes a bridge to far but we work around it.

* It warns against perfectly-correct C++14 code which makes it harder to write normal C++ which works with both C++14 and C++17.
* `if constexpr` asks the compiler to do different things from `if`. It is not a drop-in replacement when the conditional expression is constant.

Fixes: https://github.com/facebook/folly/pull/1588.

Differential Revision: D28692386


AgneLukoseviciute:
@yfeldblum friendly ping -- looks like the Internal Builds & Tests check has been stuck "in progress"

yfeldblum:
It's merged internally. It's just that the syncing of internal commits to folly's github isn't working right now, but we have some engineers looking into this. Once we get the sync working again, this PR will be merged.

PR #1593, opened 2021-05-30
Fix openNoInt building error after android ndk r21

Kudo:
Android NDK bionic with FORTIFY will override original `open()` definition and making folly `wrapNoInt` template failed to deduct.
The issue may happen only after NDK r21 because [this commit landed after r21](https://android.googlesource.com/platform/bionic/+/9349b9e51b41d12fd054b925802b626ca2db0afb%5E%21/#F0)

References:
https://github.com/android/ndk/issues/1328
https://github.com/llvm/llvm-project/commit/0a0e411204a2baa520fd73a8d69b664f98b428ba

### DIFF
--- a/folly/FileUtil.cpp
+++ b/folly/FileUtil.cpp
@@ -35,7 +35,26 @@ namespace folly {
 using namespace fileutil_detail;
 
 int openNoInt(const char* name, int flags, mode_t mode) {
+#if defined(__ANDROID__)
+  // NDK bionic with FORTIFY has this definition:
+  // https://android.googlesource.com/platform/bionic/+/9349b9e51b/libc/include/bits/fortify/fcntl.h
+  // ```
+  // __BIONIC_ERROR_FUNCTION_VISIBILITY
+  // int open(const char* pathname, int flags, mode_t modes, ...) __overloadable
+  //         __errorattr(__open_too_many_args_error);
+  // ```
+  // This is originally to prevent open() with incorrect parameters.
+  //
+  // However, combined with folly wrapNotInt, template deduction will fail.
+  // In this case, we create a custom Open lambda to bypass the error.
+  // The solution is referenced from
+  // https://github.com/llvm/llvm-project/commit/0a0e411204a2baa520fd73a8d69b664f98b428ba
+  //
+  auto Open = [&]() { return open(name, flags, mode); };
+  return int(wrapNoInt(Open));
+#else
   return int(wrapNoInt(open, name, flags, mode));
+#endif
 }
 
 static int filterCloseReturn(int r) {


Ashoat:
Thanks for this PR @Kudo! Our team has a React Native app with some C++ code mixed in, and we use the Folly library. We've had to [patch around](https://github.com/CommE2E/comm/blob/master/native/android/app/build.gradle#L152-L159) this issue  for a while, so will be great to have a long-term fix. It's something that [a lot of other people](https://github.com/facebook/react-native/issues/28298) seem to be experiencing too!

Excited to see a Folly release with this and #1584 merged, so we can get rid of our custom patches.

gengjiawen:
@yfeldblum Can you review this ? Thanks.

gengjiawen (approved):

folly/FileUtil.cpp
```diff
@@ -35,7 +35,26 @@ namespace folly {
...
+  //         __errorattr(__open_too_many_args_error);
+  // ```
+  // This is originally to prevent open() with incorrect parameters.
+  //
+  // However, combined with folly wrapNotInt, template deduction will fail.
+  // In this case, we create a custom Open lambda to bypass the error.
+  // The solution is referenced from
+  // https://github.com/llvm/llvm-project/commit/0a0e411204a2baa520fd73a8d69b664f98b428ba
+  //
+  auto Open = [&]() { return open(name, flags, mode); };
```
yfeldblum, line 53: Let's write this as:

```
auto openWrap = +[](const char* pathname, int flags, mode_t modes) {
  return open(name, flags, mode);
}
auto openFunc = kIsAndroid ? openWrap : openFunc;
return int(wrapNoInt(openFunc, name, flags, mode));
```

simpkins: I think it would be simpler to just use the lambda in all cases, and don't bother checking `kIsAndroid`.  The use of the lambda doesn't appear to affect the code generated by either gcc or clang in optimized builds.

I think we can just say
```
auto openWrapper = [&] { return open(name, flags, mode); };
return int(wrapNoInt(openWrapper));
```

Kudo: thanks for the review.
should i revise changes in this pr, as imported to facebook internal phabricator already?

yfeldblum: No need to revise.

Kudo: Thanks!

gengjiawen: Is the import blocked due to some reason ? I see it has been blocked for two days.

yfeldblum: This is merged internally. The syncing from the internal repo to github is not working at the moment, but once it is working again this change will be merged on github as well.

gengjiawen: When this synced to github. Can you draft a new release for this ? react-native has been stuck for a while.

gengjiawen: @simpkins Looks the import is stuck.

gengjiawen: @simpkins ping

PR #1598, opened 2021-06-04
Update RateLimiter.cpp

Liam0205:
Side effects after L23 is not necessary to be observed in other threads, hence the `std::memory_order_relaxed` for `load` operation.

When comparison failed, `std::memory_order_relaxed` is enough for loading `timestamp_` into `ts`. Otherwise, `std::memory_order_release` is required.

### DIFF
--- a/folly/logging/RateLimiter.cpp
+++ b/folly/logging/RateLimiter.cpp
@@ -20,13 +20,13 @@ namespace folly {
 namespace logging {
 
 bool IntervalRateLimiter::checkSlow() {
-  auto ts = timestamp_.load();
+  auto ts = timestamp_.load(std::memory_order_relaxed);
   auto now = clock::now().time_since_epoch().count();
   if (now < (ts + interval_.count())) {
     return false;
   }
 
-  if (!timestamp_.compare_exchange_strong(ts, now)) {
+  if (!timestamp_.compare_exchange_strong(ts, now, std::memory_order_release, std::memory_order_relaxed)) {
     // We raced with another thread that reset the timestamp.
     // We treat this as if we fell into the previous interval, and so we
     // rate-limit ourself.


folly/logging/RateLimiter.cpp
```diff
@@ -20,13 +20,13 @@ namespace folly {
...
 bool IntervalRateLimiter::checkSlow() {
-  auto ts = timestamp_.load();
+  auto ts = timestamp_.load(std::memory_order_relaxed);
   auto now = clock::now().time_since_epoch().count();
   if (now < (ts + interval_.count())) {
     return false;
   }
 
-  if (!timestamp_.compare_exchange_strong(ts, now)) {
+  if (!timestamp_.compare_exchange_strong(ts, now, std::memory_order_release, std::memory_order_relaxed)) {
```
yfeldblum, line 29: There is no load-acquire anymore, so what does the store-release do that a store-relaxed would not?

simpkins: Yeah, it feels like the success order here should probably be std::memory_order_acq_rel.

In practice we modify `count_` with acquire/release ordering just before and after these accesses to timestamp_, so the barriers around accesses to count_ will probably prevent things from being reordered meaningfully with the timestamp access even if we did the timestamp exchange with relaxed ordering.  Using acq_rel here too seems like a more conservative choice though, unless we have real-world examples of this being a performance problem.

Liam0205: > There is no load-acquire anymore, so what does the store-release do that a store-relaxed would not?

Sorry for the late response.

Yes, this was a typo here, a relaxed-store is enough here.

Liam0205: > Yeah, it feels like the success order here should probably be std::memory_order_acq_rel.
> 
> In practice we modify `count_` with acquire/release ordering just before and after these accesses to timestamp_, so the barriers around accesses to count_ will probably prevent things from being reordered meaningfully with the timestamp access even if we did the timestamp exchange with relaxed ordering. Using acq_rel here too seems like a more conservative choice though, unless we have real-world examples of this being a performance problem.

Well, I believe that relax order is enough for `timestamp_`.

Threads that request a rate limiter check interact with each other by `count_` and `timestamp_`. While each time we update `timestamp_`, a release-store to `count_` follows, which will then synchronizes with the acquire-load of `count_` (in the quick check), just before load and store to `timestamp_`. 

That is, store operation to `timestamp_` happens-before the release-load to `count_`, and the release-load inter-thread happens before the acquire-load of `count_` in another thread, and the acquire-load happens before the load operation to `timestamp_` in the later thread. Considering the transition of happens-before relationship, the store operation on `timestamp_` happens before the load operation on it in another thread.

PR #1605, opened 2021-06-14
Remove semicolons at the end of macros after `do {} while (0)`

JTJL:
The semicolons at the end of macros after `do {} while (0)` is useless and may cause potential compile errors in the future.


yfeldblum (changes requested):
Seems `;` is required somewhere - check out the build failures.

PR #1609, opened 2021-06-24
add deps fmt; .gitignore and macos openssl path

wendajiang:


CMakeLists.txt
```diff
@@ -15,6 +15,23 @@
...
+  MESSAGE(STATUS "current platform: Linux ")
+ELSEIF (CMAKE_SYSTEM_NAME MATCHES "Windows")
+  MESSAGE(STATUS "current platform: Windows")
+ELSEIF (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+  MESSAGE(STATUS "current platform: FreeBSD")
+ELSEIF (CMAKE_SYSTEM_NAME MATCHES "Darwin")
+  MESSAGE(STATUS "current platform: MacOS")
+  set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl" )
+  set(OPENSSL_LIBRARIES "/usr/local/opt/openssl/lib" )
+  set(OPENSSL_INCLUDE_DIR "/usr/local/opt/openssl/include" )
```
yfeldblum, line 29: I'm not sold that we should hardcode system openssl in the build.

It may be possible to set these in the make invocation, and to choose system libs using the higher-layer fbcode-builder scripts.

autoantwort: We should not. On arm64 macos these path does not exists.

autoantwort: I have created https://github.com/facebook/folly/pull/2016 for that.

PR #1611, opened 2021-06-26
Fix typo in Hazptr

okian:


Orvid:
We've fixed the sync so this will need to be rebased

okian:
> We've fixed the sync so this will need to be rebased

Done

PR #1617, opened 2021-07-15
[folly/logging] Fix mixed MSVC/clang-cl build

ahupp:
An executable built by clang-cl that is linked against folly/logging
from vcpkg will fail to link due to a few missing symbols.  folly/logging
has a number of functions that are only defined when __INCLUDE_LEVEL__ is defined.
vcpkg always uses MSVC, MSVC does not support this macro while clang-cl does,
resulting in mismatch between what is defined, used, and declared.

These ifdefs mostly came from fc507b26548882a8492de57abf0a8923ecf7709a.

The (minor IMO) negative consequence of this diff is that
XlogFileScopeInfo will be unnecessarily larger (a pointer and an uint32)
 on compilers that don't support __INCLUDE_LEVEL__ (e.g MSVC).

The actual condition of whether to make use of __INCLUDE__LEVEL__ remains, via the XLOG_IS_IN_HEADER_FILE macro.

### DIFF
--- a/folly/logging/LogStreamProcessor.cpp
+++ b/folly/logging/LogStreamProcessor.cpp
@@ -111,7 +111,6 @@ LogStreamProcessor::LogStreamProcessor(
       message_{std::move(msg)},
       stream_{this} {}
 
-#ifdef __INCLUDE_LEVEL__
 namespace {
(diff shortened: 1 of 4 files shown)


JeffBotshire:
That's just awesome. Open source now? That was my idea!!! Good work team.
#superbøts

PR #1639, opened 2021-08-31
Adding ECN support to folly

mjoras:
Summary:
Added ECN support to folly/io/async/AsyncUDPSocket.

The support consists:
1. Configuring TOC (for IPv4) and TCLASS (for IPv6) fields for sending UDP packets
2. Parsing and passing TOC/TCLASS header as part of AsyncUDPSocket::ReadCallback::OnDataAvailableParams

Notes:
- The implementation always use `msghdr` parsing, This required to properly extract `TOS/TCLASS` IP header
- The `AsyncUDPSocket` already has `setTraficClass` function that somewhat similar to `setTos`. The difference: `setTraficClass` works after socket creation and only for IPv6. `setTos` executes during socket initialization (e.g. binding) and works both for IPv4 and IPv6.

Differential Revision: D27573238


mjoras:
Exported this for reference of a potential external contributor.

Issue #1658, opened 2021-09-22
Typedef redefinition


arthurgeron-work:
When building a react native project with Folly `0.110.0` I get the following error:
```ios/Pods/RCT-Folly/folly/portability/Time.h:52:17: Typedef redefinition with different types ('uint8_t' (aka 'unsigned char') vs 'enum clockid_t')```

arthurgeron-work:
created on my work account, closing

Simar05519:
Do we have to explain typedef?

DominicGBauer:
I experience the same issue.

I've gotten it down to changing from platform :ios, '11.0' to 12.0'.

yfeldblum:
Closing on the assumption that #1724 fixes the problem.

PR #1660, opened 2021-09-22
fix: typedef redefinition

arthurgeron:
Solves #1658 .
Typedef redefinition error prevents apps using latest folly version from building


rcari:
Please see https://github.com/facebook/folly/pull/1724 which provides a better alternative by preventing the redefinition altogether.

yfeldblum:
Closing on the assumption that #1724 fixes the problem.

PR #1675, opened 2021-11-22
Addition of std::string_view support in FBString

sjanel:
Make `folly::fbstring` support constructors and other methods work with `std::string_view` (similar to `std::string`).

Updated the unit test accordingly.

Also, I changed `ssize_t` to `intmax_t` as it is a POSIX type not always defined for Windows machines.


sjanel:
Hello, can someone relaunch the builds please ? The errors do not seem to come from the PR.

PR #1694, opened 2021-12-29
Fix compilation when using Clang + libstdc++ (fixes #1685)

### LINKED ISSUE #1685: C++ coroutines on Clang 13 and libstdc++
Hi all,

I use folly in combination with clang 13 and libstdc++.
As far as I know, this combination should not support c++ coroutines as coroutines on clang require libc++.

However, [`Portability.h`](https://github.com/facebook/folly/blob/8bdd2efdd5f57e2e7480b9172e6619b3774f828f/folly/Portability.h#L554) sets `FOLLY_HAS_COROUTINES` to 1. 

As far as I see the problem is that folly only checks if `(__has_include(<coroutine>)` or `__has_include(<experimental/coroutine>)`  is valid.
This is true for my combination. 

Thus, I think that we should extend this check to something like:

`(__has_include(<coroutine>) && defined(__GNUC__)` or `__has_include(<experimental/coroutine>) && defined(__clang__)`  

Best,
Philipp

mhx:
The logic in Portability.h would enable coroutines as soon as the
<coroutine> header from libstdc++ is found. However, this header
won't build properly with Clang and -std=c++20. The fix explicitly
only checks for <experimental/coroutine> when building with Clang,
and <coroutine> with GCC.


Orvid:
The current version is already in use with clang 12 + libstd++, and this actually appears to break that, causing it to fail to detect that coroutines are available at all.

Orvid (changes requested):
See comment

mhx:
I'll try and take another look!

PhilippGrulich:
This pr would close #1685

mhx:
@Orvid, I finally found some time to look into this again.

Could you please check if my updated PR still causes problems in your setup?

Orvid:
This still breaks things. I think the issue here may be a bit more fundamental: Our internal setup builds using clang, and uses libstdc++, and has coroutines enabled. I'm not sure if we've done anything special to make that happen, but I'd be surprised if we did. From the sound of things though, this combination isn't supposed to be working? Is that correct?

mhx:
> This still breaks things. I think the issue here may be a bit more fundamental: Our internal setup builds using clang, and uses libstdc++, and has coroutines enabled. I'm not sure if we've done anything special to make that happen, but I'd be surprised if we did. From the sound of things though, this combination isn't supposed to be working? Is that correct?

Oh, I somehow misread your earlier comment as you're using *libc++* with clang.

So yeah, I *think* the combination clang + libstdc++ doesn't work out-of-the-box, at least not with clang-13 or lower. The reason it's not working is that libstdc++ requires `__cpp_impl_coroutine` to be defined by the compiler, which clang-13 or lower don't define at all. With gcc, you need to pass `-fcoroutines` or `-std=c++20` to get the compiler to define `__cpp_impl_coroutine`, but with clang, it just doesn't define it at all (only `__cpp_coroutines`). There's [a long discussion](https://reviews.llvm.org/D109433) on clang/libc++ coroutine support that mentions some of this and I've just double checked with some test code.

Ultimately I think the bit that's wrong is this part of the check:

```
(__cpp_coroutines >= 201703L || __cpp_impl_coroutine >= 201902L)
```

I'm assuming the intent here was to support both gcc (doesn't define `__cpp_coroutines`) and clang (doesn't define `__cpp_impl_coroutine` until clang-14). If your setup using clang-12 + libstdc++ is indeed working, I would hazard a guess that you're setting `__cpp_impl_coroutine` manually somewhere in order to make libstdc++ work.

From all I've read now, I *think* a better check would be:

```
#if (__cpp_coroutines >= 201703L && __has_include(<experimental/coroutine>)) || \
    (__cpp_impl_coroutine >= 201902L && __has_include(<coroutine>))
```

The first line would cover clang + libc++ (up until version 13). The second line would cover gcc + libstdc++ and clang (from version 14). In your setup (if I'm correct about manually defining `__cpp_impl_coroutine`), the first check would fail (no `<experimental/coroutine>`), but the second check should succeed. There's probably a whole bunch of combinations of compilers and libraries that are not covered, but I would hope that it works for your setup and fixes the out-of-the-box configuration.

I've updated the PR with a new diff that implements the above check. If that doesn't work, I'm happy to just live with setting `FOLLY_CFG_NO_COROUTINES` in my project for now.

Issue #1701, opened 2022-01-07
Support of Arch Linux in `getdeps.py`


beryll1um:
## Info
Project support only `dnf` and `apt` package managers. I want to add `pacman` package manager support.

## Steps
- Add Arch distro and `pacman` package manager option for it
- Add `pps` (`pacman-packages`, thx for the default manifest options names guys) option in `manifest.py`

PR #1702, opened 2022-01-08
Support of Arch Linux in `getdeps.py`

beryll1um:
I don't sure that I make all in accordance with the your contribution pipeline, so please correct me If there is needed.
So, I've made some changes to support the `pacman` package manager in your `getdeps.py` script. In `manifests` I'm also duplicated some packages from `debs` and `rpms` sections and create a new `pps` sections with according packages for  `pacman`.

Issue:  #1701

### DIFF
--- a/build/fbcode_builder/getdeps.py
+++ b/build/fbcode_builder/getdeps.py
@@ -347,7 +347,7 @@ class InstallSysDepsCmd(ProjectCmdBase):
         )
         parser.add_argument(
             "--package-type",
-            choices=["rpm", "deb"],
+            choices=["rpm", "deb", "pacman-package"],
             default=None,
             help="Allow overriding the package type so can see deb from centos",
         )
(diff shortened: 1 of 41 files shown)


quark-zju:
Mostly look good but names could be a bit more explicit:
- Since "pacman" is also used by MinGW, could we use "arch-package" instead of "pacman-package"?
- Could we use "arch-packages" instead of "pps"? It seems unclear what "pps" means without reading the code.

Otherwise the change looks good to me. Thanks for fixing Arch issues!

beryll1um:
> Mostly look good but names could be a bit more explicit:
> 
>     * Since "pacman" is also used by MinGW, could we use "arch-package" instead of "pacman-package"?
> 
>     * Could we use "arch-packages" instead of "pps"? It seems unclear what "pps" means without reading the code.
> 
> 
> Otherwise the change looks good to me. Thanks for fixing Arch issues!

Yes. I think your comments make sense. Will be changed.

beryll1um:
Ping! Is it still relevant?

PR #1704, opened 2022-01-09
Update README.md

ravithemore:
Small Grammartical Mistake.


Orvid:
The grammar was already correct as written.

PR #1717, opened 2022-02-10
Move callables another time in exception_wrapper

benlandrum:
The `exception_wrapper` moves callables around in `with_exception,` except for one place. This forces callables to be copyable. Use `std::move` in the remaining place.

### DIFF
--- a/folly/test/ExceptionWrapperTest.cpp
+++ b/folly/test/ExceptionWrapperTest.cpp
@@ -16,6 +16,7 @@
 
 #include <folly/ExceptionWrapper.h>
 
+#include <memory>
 #include <stdexcept>
 
 #include <folly/Conv.h>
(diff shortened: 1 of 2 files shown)


folly/test/ExceptionWrapperTest.cpp
```diff
@@ -151,6 +151,18 @@ TEST(ExceptionWrapper, with_exception_test) {
...
+  // Just needs to compile.
+  struct MoveOnly {
+    MoveOnly() = default;
+    MoveOnly(const MoveOnly&) = delete;
+    MoveOnly& operator=(const MoveOnly&) = delete;
+    MoveOnly(MoveOnly&&) noexcept = default;
+    void Foo() const {}
+  };
+  MoveOnly move_only;
+  cew.with_exception([move_only = std::move(move_only)](const std::exception&) { move_only.Foo(); };
```
yfeldblum, line 165: I think we can get away with choosing an existing move-only type, such as `std::unique_ptr<int>`.

```
  // a move-only callback type:
  cew.with_exception([v = std::make_unique<int>(7)](const std::exception&) {});
```

yfeldblum (commented):
Thanks! One request, listed in the code.

benlandrum:
Hi, @yfeldblum. Any other changes requested? Could you please estimate when you'd be able to merge?

benlandrum:
Thank you, @yfeldblum.

PR #1723, opened 2022-02-23
Fix typdef error

KrisLau:
Was getting a `typedef redefinition with different types ('uint8_t' (aka 'unsigned char') vs 'enum clockid_t')` error until I did this in my file. same error as mentioned in https://github.com/facebook/folly/issues/1670


rcari:
Please see https://github.com/facebook/folly/pull/1724 which provides a better alternative by preventing the redefinition altogether.

Issue #1470, opened 2020-10-07
Time.h `typedef redefinition` with iOS 10 set as a deployment target


grabbou:
Hey,

I am working on [a Hermes support for React Native](https://github.com/facebook/react-native/pull/29914) and while working on it, I have run into the following issue, when building the React Native project:

<img width="500" alt="Screenshot_2020-10-01_at_11 07 41" src="https://user-images.githubusercontent.com/2464966/95354686-fe6b7700-08c4-11eb-9bbc-739e27fd44f7.png">

React Native has deployment target set to `10.0` for all its parts, including third party pod specs such as [`RCT-Folly`](https://github.com/facebook/react-native/blob/master/third-party-podspecs/RCT-Folly.podspec), that it defines to download and compile Folly with pieces that it needs.  After changing the deployment target for `RCT-Folly` specifically from `10.0` to `9.0`, the issue goes away. 

This is [a temporary workaround](https://github.com/facebook/react-native/pull/29914/commits/16452d0c7509a839a3887a8e26a6147a0060e132) that we have implemented to unblock the PR.

I believe it is caused by [the following code](https://github.com/facebook/folly/blob/master/folly/portability/Time.h#L30-L33):
```cpp
#if __MACH__ &&                                                       \
        ((!defined(TARGET_OS_OSX) || TARGET_OS_OSX) &&                \
         (MAC_OS_X_VERSION_MIN_REQUIRED < MAC_OS_X_VERSION_10_12)) || \
    (TARGET_OS_IPHONE && (__IPHONE_OS_VERSION_MIN_REQUIRED < __IPHONE_10_0))

#ifdef FOLLY_HAVE_CLOCK_GETTIME
#undef FOLLY_HAVE_CLOCK_GETTIME
#endif

#define FOLLY_HAVE_CLOCK_GETTIME 1
#define FOLLY_FORCE_CLOCK_GETTIME_DEFINITION 1

#endif
```

The way I read it is - when building for iPhone and targeting iOS versions lower than 10.0 (e.g. 9.0), don't define macros. When building for newer iOS release, such as 10.0, define it. Shouldn't it be the opposite, if the purpose was to provide backwards compatibility?

The issue is being brought up from time to time in various projects - that's where I have learnt about the "deployment target" workaround:
- https://github.com/facebook/react-native/issues/28810
- https://github.com/facebook/react-native/issues/27845
- https://github.com/facebook/flipper/issues/834

Here are my questions:
[...]

PR #1724, opened 2022-02-24
Fix detection of clock_gettime availability

rcari:
On Apple platforms, following the SDK development [1] guidelines, we can check
for the availability (as in definition) of symbols using macros. This is more
reliable notably when targeting iOS platforms or simulators where the previous
approach used to create redefinition problems [2] & [3].

[1] https://developer.apple.com/forums/thread/67102
[2] https://github.com/facebook/folly/issues/1470
[3] https://github.com/facebook/flipper/issues/834

### DIFF
--- a/folly/portability/Time.h
+++ b/folly/portability/Time.h
@@ -27,11 +27,7 @@
 // solve that by pretending we have it here in the header and
 // then enable our implementation on the source side so that
 // gets linked in instead.
-#if defined(__MACH__) &&                                               \
-        ((!defined(TARGET_OS_OSX) || TARGET_OS_OSX) &&                 \
-         (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_12)) || \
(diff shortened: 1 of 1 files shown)


rcari:
@Orvid this is still relevant today and still causes symbol redefinition issues when used in React Native. Can we get this reviewed? Thanks!

Orvid:
Do you know which version of the xcode SDK introduced this define?

rcari:
This macro expands to `__OSX_AVAILABLE(10.12) __IOS_AVAILABLE(10.0) __TVOS_AVAILABLE(10.0) __WATCHOS_AVAILABLE(3.0)` so I assume it shipped with the macOS 10.12 SDK, and iOS 10.0.
The beauty of this define is that it does exactly what we want and is only set if the `clock_` methods and associated types are defined.
However there is still an issue with Folly not checking whether the `clock_gettime` symbol is weak at runtime which I tried to address in https://github.com/facebook/folly/pull/1725.

rcari:
@Gownta could you help or point me to someone who can?

Issue #1602, opened 2021-06-09
Time.cpp's clock_gettime fallback breaks libc++ steady_clock::now


tux3:
Hello,

We maintain a Native Module for React-Native that supports Apple platforms, and so we (and our users) have a dependency on Folly through the RN ecosystem.

Our module statically links with libc++, and some of our own C++ dependencies use `std::steady_clock`, which in recent versions of libc++ [calls `clock_gettime(CLOCK_MONOTONIC_RAW, &tp)`](https://github.com/llvm/llvm-project/blob/llvmorg-11.1.0/libcxx/src/chrono.cpp#L123-L129) on Apple platforms.
Our users have been reporting that `std::steady_clock::now()` throws after calling `clock_gettime`, resulting in an uncaught exception and a crash:

```cpp
* thread #23, stop reason = signal SIGABRT
  * frame #0: 0x00007fff6113133a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff61166e60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff200fab94 libsystem_c.dylib`abort + 120
    frame #3: 0x000000010da81ff3 ExampleApp`abort_message + 195
    frame #4: 0x000000010d878fae ExampleApp`demangling_terminate_handler() + 238
    frame #5: 0x000000010d92f0f8 ExampleApp`std::__terminate(void (*)()) + 8
    frame #6: 0x000000010d92f079 ExampleApp`std::terminate() + 41
    frame #7: 0x000000010d85dfbb ExampleApp`__clang_call_terminate + 11
    frame #8: 0x000000010e04bdf9 ExampleApp`std::__1::chrono::steady_clock::now() + 73
```

## Why do things break

It turns out that Folly's fallback implementation exports its `clock_gettime` implementation globally, as a weak symbol! So when our native module is statically linked into the user's application, Folly's clock_gettime symbol will bind to libc++'s `clock_gettime` import, even on devices >= iOS 10 where the symbol would otherwise be loaded from `libsystem_c.dylib` at runtime.

That would be fine, except Folly's clock_gettime [does not support CLOCK_MONOTONIC_RAW](https://github.com/facebook/folly/blob/master/folly/portability/Time.cpp#L117-L119). It returns EINVAL, which causes steady_clock to throw, and our users get to keep both pieces. This is suboptimal :(

## The macro in Time.h

[...]

PR #1725, opened 2022-02-25
Multiple fixes notably clock_gettime on Apple platforms

rcari:
The clock_gettime situation on Apple platforms is complex enough as it is.
Linking an incomplete implementation in replacement of the global symbol breaks
other libraries including libc++ which uses clock_gettime internally [1].
Depending on the approach taken by the libc++ the current implementation could
trigger infinite recursion calls as well.
    
Since the goal of the /portability/ is to make Folly portable, we should adapt
and use an approach which is compatible with Apple SDK development while
maintaining compatibility on other platforms.
    
This patch does that by adding a folly_ prefix to clock functions to avoid
clashing with the native symbol. It's the role of other libraries to adhere to
Apple SDK development guidelines when using the clock_ family of functions.
    
As a side effect, this also addresses the redefinition issues that this [2]
other proposed fix solved.
    
[1] https://github.com/facebook/folly/issues/1602
[2] https://github.com/facebook/folly/pull/1724


Orvid:
The portability layer here is intended primarily to port other projects, rather than folly itself, so we try to maintain at least the appearance of the standard version of the API. For the issue you're running into here, would something similar to how fnctl functions are handled on windows work? (https://github.com/facebook/folly/blob/main/folly/portability/Fcntl.h) Basically declare them in a namespace and then add a global using namespace to make them appear in the right place.

The other changes look good though. Would it be possible to split those off into a different PR?

rcari:
Hi @Orvid and thanks for your feedback!
I'm not sure I follow how Folly helps port other projects here. Such projects would have to `#include <folly/portability/Time.h>` to find these definitions but cannot use `folly_clock_` instead of `clock_`? It looks like these other projects explicitly depend on Folly rather than standard library calls if they include Folly headers specifically.
My point in the PR description was that such library trying to achieve macOS portability should also follow Apple SDK development guidelines, or use these new Folly drop-in replacements.
In any case regarding what you mentioned with `fcntl`, I can give it a try but am wondering whether we won't run into symbol definition clashes...
In the meantime, I'll get the other fixes in another dedicated PR and keep you posted.

rcari:
@Orvid after careful consideration, this would not work on Apple platforms because they define the symbols themselves whereas `fcntl` is not defined on WIN32 and thus defined entirely in Folly.
The only viable option here in my view would be to `#define clock_gettime folly_clock_gettime`, in the header and `#undef clock_gettime` in the implementation to actually lookup the right symbol from the SDK.

PR #1777, opened 2022-05-17
Use std::declval<A>() instead of A{} to workaround gcc-12 + CWG 903

### LINKED ISSUE #1772: compile error on gcc12
image
I am having this compile error on GCC12.1 and I don't understand why~，any advice will help，thanks~

alex-700:
- it is workaround CWG 903 in gcc-12, which is considering `int{}`  as null pointer constant, and so can't choose between `safe_assert_msg_types_one(uint64_t)` and `safe_assert_msg_types_one(char const *)`
- see https://godbolt.org/z/fov5xPjEe

closes #1772

### DIFF
--- a/folly/lang/SafeAssert.h
+++ b/folly/lang/SafeAssert.h
@@ -130,7 +130,7 @@ FOLLY_INLINE_VARIABLE constexpr safe_assert_msg_types_one_fn
     safe_assert_msg_types_one{}; // a function object to prevent extensions
 
 template <typename... A>
-safe_assert_msg_type_s<decltype(safe_assert_msg_types_one(A{}))::value...>
+safe_assert_msg_type_s<decltype(safe_assert_msg_types_one(std::declval<A>()))::value...>
 safe_assert_msg_types_seq_of(A...); // only used in unevaluated contexts
 
 template <typename>


folly/lang/SafeAssert.h
```diff
@@ -130,7 +130,7 @@ FOLLY_INLINE_VARIABLE constexpr safe_assert_msg_types_one_fn
     safe_assert_msg_types_one{}; // a function object to prevent extensions
 
 template <typename... A>
-safe_assert_msg_type_s<decltype(safe_assert_msg_types_one(A{}))::value...>
+safe_assert_msg_type_s<decltype(safe_assert_msg_types_one(std::declval<A>()))::value...>
```
yfeldblum, line 133: It looks like a shorter syntax is available.
```suggestion
safe_assert_msg_type_s<decltype(safe_assert_msg_types_one((A)A{}))::value...>
```

alex-700: @yfeldblum thanks for the idea!
But I prefer to use `std::declval` here, because it explicitly claims that this is unevaluated context.

yfeldblum: I assumed `declval` might be slower to compile, but I am not seeing that assumption holding in my tests.

yfeldblum: The use of `declval` triggered some bug or pathological behavior somewhere. I don't know what the problem is, but that's why this was reverted.

alex-700: @yfeldblum should I do something with it?

yfeldblum: I have merged this again but with `(A)`. Hopefully that will not trigger the same bug.

PR #1790, opened 2022-06-10
Add double conversion flags to folly json

chaoz2:
support `double-conversion` flags for folly json serialization

### DIFF
--- a/folly/Conv.h
+++ b/folly/Conv.h
@@ -646,9 +646,10 @@ typename std::enable_if<
 toAppend(
     Src value,
     Tgt* result,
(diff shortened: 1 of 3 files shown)


chaoz2:
@Orvid do you mind reviewing it since the change is simple and straightforward?

Orvid:
Because this is inserting a new argument to `toAppend` in the middle of the argument list, with no default value, this actually breaks every existing place that calls this function, of which there are quite a few internally. To be able to merge this, the new argument would need to be the final one, and have a default value.

PR #1812, opened 2022-07-06
[json] Re-enable compilation of JsonTest and add two perf tests

Adenilson:
It seems that compiling of the main Json parser unit test was disabled
due to a 'MSVC Preprocessor' bug.

It is safe to assume that g++/clang++ will be used on environments like
Linux/etc while MSVC would be used on MS-Windows.

This change will re-enable the unit test compilation while also adding
two new unit tests targeting stressing the performance of the Json parser.


Adenilson:
@aaronabramov, @yfeldblum and @ispeters any feedback?

The idea is to re-enable compilation of the main JSON unit test and add two new unit tests that could be used for a perf profile targeting identifying hot spots to guide optimization work.

I'm also re-posting some questions and remarks I've added on other merge request (#1805 ), quote: "I started by having a look on the JSON parsing class and noticed that the main unit test was commented out (apparently related to a MSVC bug?).

To get it building again was a one liner fix, I proceeded next to use as an input file a JSON file generated by retrieving the comments from my own Facebook account (I named the file 'input.json').

I wonder if there is a corpus of sample JSON files that are relevant for Folly client code?.

The reason I ask about a corpus is that I observed diverging ordering of hot spots depending on the input JSON file (e.g. twitter.json vs FB comments sample file mentioned above).

A perf profile showed that one of the top 8 hot spots (alongside parseValue, parseString, F14Table, dynamic::operator=) while parsing JSON was relate to hashing.

That lead me to have a look on which hashes and where they were deployed in the code base.

There may be good potential for performance gains in JSON parsing by leveraging SIMD optimizations.

I've done some previous work on vectorizing zlib for Chromium and it is quite surprising the performance gains achievable by working with vectors instead of a scalar approach.

The good news is that there is previous work done on the area (e.g. simdjson).

Would Folly be open to adopt SIMD specific optimizations in its JSON parser or maybe just migrate to simdjson (if it has all the required features)?"

Orvid:
Enabling the json test for open source builds is good, but for the perf tests, we do have a benchmark framework (https://github.com/facebook/folly/blob/main/folly/Benchmark.h - With https://github.com/facebook/folly/blob/main/folly/test/AsciiCaseInsensitiveBenchmark.cpp as an example). The issue is that I don't think we have these setup to be buildable from cmake currently.

Adenilson:
Sure can do!

Adenilson:
@Orvid first draft on turning the new tests into performance benchmarks.

folly/test/JsonBenchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <folly/init/Init.h>
+#include <folly/json.h>
+
+#include <fstream>
+#include <streambuf>
+
+using namespace folly;
+
+BENCHMARK(PerfJson2Obj, iters) {
+  std::ifstream fin("input.json");
```
Orvid, line 27: Is this a file that we can reasonably import into the repo?

folly/test/JsonTest.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include <folly/portability/GTest.h>
 
+FOLLY_GNU_DISABLE_WARNING("-Wmissing-field-initializers")
```
Orvid, line 25: Where are these showing up? We should generally be clear of this particular warning.

Adenilson:
It is a bit odd that comments are listed as 'pending' instead of simply being submitted. Perhaps a github quirk?

Adenilson:
Anything pending to have this landed?

Adenilson:
Friendly ping on the subject.

Can we proceed and land this patch as it should be low risk?

Orvid:
Sorry for the delay on this. The current status is that enabling the json tests is fine, but for the benchmarks the default needs to have at least something checked in as the default files to use. If folks want to pull in a third-party json file to benchmark more completely that is fine, but we need something to be there by default.

Adenilson:
@Orvid just added a sample file in the current patch and comments in the source code to where to find better/bigger JSON files.

Tested by building the new target (i.e. make json_benchmark) and next running the benchmark. Steps:
a) mkdir out; cd out;
b) cmake ../ -DCMAKE_BUILD_TYPE=Release -DBUILD_BENCHMARKS=ON
c) make json_benchmark json_test -j8
d) Followed next by executing the newly enabled/added test (https://gist.github.com/Adenilson/d36aa75f561c72e7377fa554a9e71be6).

I guess it should be good to go.
:-)

Adenilson:
I see that the build bots seem to fail.

A quick look on the linux bot showed the following compilation error: "/home/runner/work/folly/folly/folly/test/JsonTest.cpp:361:64: sorry, unimplemented: non-trivial designated initializers not supported"

I'm building with g++ 11.2.0 (standard in Ubuntu 22 LTS IIRC) with ToT (Top of the Tree) and no special configuration flags (i.e. only used what I posted in my previous message).

A quick look on the same bot seems to show that apparently gcc 7.5 is used (quote "The CXX compiler identification is GNU 7.5.0" in https://github.com/facebook/folly/runs/7888073890?check_suite_focus=true).

Any particular reason to use such older version of gcc/g++ by the build bots? It seems that gcc 7.2 was released in Aug 2017 (i.e. 5 years ago).

It might be the case that it could be workarounded using an extra compiler flag as needed (e.g. -std=c++14).

Since I don't have the privileges to add a patch to the compiler bots for a dry run, I would like to coordinate a fix to avoid multiple tries to address the issue at hand.

@Orvid : What are your thoughts?

Adenilson:
And gcc/g++ 7.5 apparently was released in November 2019 (https://gcc.gnu.org/releases.html), therefore almost 3 years old.

Adenilson:
The mac compilation error is at "ProducerConsumerQueueBenchmark.cpp:49:7: error: unknown type name 'cpu_set_t'" which is unrelated to the current change.

As a matter of fact, it seems that JsonTest.cpp builds fine with the compiler used by the mac bot (i.e. AppleClang 13.0.0.13000029).

Adenilson:
The Windows bot failure also seems unrelated with the current change.

The compilation error is "tests\Base64SpecialCasesTest.cpp(58): error C2131: expression did not evaluate to a constant".

Adenilson:
@Orvid just pushed an update with a patch on top trying to address the build issue on Linux, roughly based on the error message (i.e. 'unimplemented: non-trivial designated initializers not supported').

I'm trying to build gcc/g++ 7.5 locally from sources to reproduce the issue but that is kinda of a pain.

folly/test/JsonTest.cpp
```diff
@@ -357,44 +357,60 @@ TEST(Json, DuplicateKeys) {
...
-          obj,
-          {
-              .allow_non_string_keys = true,
-              .convert_int_keys = true,
-          }));
+      folly::json::serialize(obj, tmp));
+  resetOpts();
 
+  tmp.allow_non_string_keys = true;
+  tmp.convert_int_keys = true;
```
adenilsoncavalcanti, line 410: This can be removed since the state is the same as in the previous call to json::serialize().

Plus saves an extraneous call to resetOpts().

adenilsoncavalcanti:
@Orvid when you got some spare time, could you please add the current patchset to be run through the bots?

CMakeLists.txt
```diff
@@ -507,6 +507,10 @@ if (BUILD_TESTS OR BUILD_BENCHMARKS)
     set(HAVE_CMAKE_GTEST OFF)
   endif()
 
+  # Some benchmarks require an input sample file.
+  file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/folly/test/input.json
+     DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
Orvid, line 512: CONTENT_DIR exists and is what's intended to do this. `input.json` will need to be moved into it's own directory though.

Adenilson:
@Orvid thanks for adding the new updated patchset for a dry-run.

That is some progress, it seems that now the patch will pass in the linux bot.

The other failing bots seem to have issues unrelated with the current changes:
a) Mac: "test/ProducerConsumerQueueBenchmark.cpp:49:7: error: unknown type name 'cpu_set_t''
b) Windows: "tests\Base64SpecialCasesTest.cpp(58): error C2131: expression did not evaluate to a constant"

What is the procedure for this situation? I'm afraid is something I cannot fix (i.e. flakey bots on unrelated code).

Adenilson:
Ok, just done the requested change using CONTENT_DIR.

Adenilson:
Turns out that building benchmarks on OSX is broken, this patches will fix it:
https://github.com/facebook/folly/pull/1847

Issue #1817, opened 2022-07-13
openssl/opensslv.h file not found


Romick2005:
Got this `'openssl/opensslv.h' file not found` build error on case sensitive file system (macOS Monterey 12.4)

![image](https://user-images.githubusercontent.com/17779660/178680508-d730c343-3adb-4a3d-8f7d-16bdae3a10e8.png)

gnagaraja323:
@Romick2005  I am also getting same issue, have you solved this issue. Please help me on this.

Romick2005:
> 

So all you need to do is to create a symlink for openssl to OpenSSL. I have added this into ios/Podfile:

```
target '[Name]' do
  ...
  post_install do |installer|
    ...
    ## Fix openssl incompatibility on case-sensitive fs. The real directory is Pods. In case-sensitive file systems the file won't exists.
      unless File.exist? "pods"
        system("cd Pods/OpenSSL-Universal/Frameworks/OpenSSL.xcframework/ios-arm64_armv7 && ln -sfh OpenSSL.framework openssl.framework")
    end
    ...
  end
end
```

PS. There are multiple plaforms inside OpenSSL.xcframework folder, so you need to select one depending on your architecture.

faithk7:
My system information: MacOS 13.3.1 (22E261)

For me, since I installed openssl using brew, from line 27 to line 29 in CMakeLists.txt, I have to set the OPENSSL_ROOT_DIR, OPENSSL_LIBRARIES, OPENSSL_INCLUDE_DIR to the followings: 

  set(OPENSSL_ROOT_DIR "/opt/homebrew/opt/openssl@1.1" )
  set(OPENSSL_LIBRARIES "/opt/homebrew/opt/openssl@1.1/lib" )
  set(OPENSSL_INCLUDE_DIR "/opt/homebrew/opt/openssl@1.1/include" )
  
  I also discovered that sometimes CMAKE_SYSTEM_NAME failed to detect my system as "Darwin", so I also have to manually set the above variables in the ELSE() block of the CMakeLists.txt. 

So I think you should check the installation directory of your ssl and check whether they match with the paths set in the CMakeLists.txt.

PR #1818, opened 2022-07-13
#1817 Fix openssl include as OpenSSL for case sensitive file systems.

Romick2005:
This should solve the issue for case sensitive file systems.

### DIFF
--- a/folly/io/async/test/AsyncSSLSocketTest.cpp
+++ b/folly/io/async/test/AsyncSSLSocketTest.cpp
@@ -53,7 +53,7 @@
 #endif
 
 #if FOLLY_OPENSSL_IS_110
-#include <openssl/async.h>
+#include <OpenSSL/async.h>
 #endif
 
 using std::cerr;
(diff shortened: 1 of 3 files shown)


Orvid:
As far as I can tell, the official casing is lowercase: https://github.com/openssl/openssl/tree/master/include/openssl

What is providing this in mixed case?

Orvid:
The casing there seems to be the result of a non-standard casing being provided by those packages, so we can't go changing them here.

The couple of linux distros I checked, upstream openssl, and vcpkg all have the directory name as lowercase, so that's what we want to keep here.

The correct fix is to update the casing used in https://github.com/krzyzanowskim/OpenSSL to match with what is already used elsewhere.

Romick2005:
But does the change break anything on your side? I think that it actually fix. Can you please confirm that this pull request does not work for you. Thank you.

Orvid:
This change should break the build on linux, since the default file system is case sensitive, but I'll kick off the builds to verify.

Orvid:
This does appear to break the linux build.

Romick2005:
Great thank you! Can you please then merge my changes? In order to fix issues for case sensitive file systems. Thank you for the check!

PR #1822, opened 2022-07-18
Add clear to SparseByteSet.

dasfex:
Add O(1) non-amortized clear to SparseByteSet and test for it.

### DIFF
--- a/folly/container/SparseByteSet.h
+++ b/folly/container/SparseByteSet.h
@@ -83,7 +83,7 @@ class SparseByteSet {
    *
    *  O(1), non-amortized.
    */
-  inline bool clear() {
+  inline void clear() {
     size_ = 0;
   }
 
--- a/folly/container/SparseByteSet.h
+++ b/folly/container/SparseByteSet.h
@@ -24,12 +24,13 @@ namespace folly {
 /***
  *  SparseByteSet
  *
- *  A special-purpose data structure representing an insert-only set of bytes.
+ *  A special-purpose data structure representing a set of bytes.
  *  May have better performance than std::bitset<256>, depending on workload.
  *
  *  Operations:
  *  - add(byte)
  *  - contains(byte)
+ *  - clear()
  *
  *  Performance:
  *  - The entire capacity of the set is inline; the set never allocates.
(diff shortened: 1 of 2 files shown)


dasfex:
Looks like builds failed not because my code.

folly/container/SparseByteSet.h
```diff
@@ -77,6 +78,15 @@ class SparseByteSet {
     return sparse_[i] < size_ && dense_[sparse_[i]] == i;
   }
 
+  /***
+   *  clear()
+   *
+   *  O(1), non-amortized.
+   */
+  inline bool clear() {
+    size_ = 0;
```
Orvid, line 87: While this will make it always return false for all values, it would fail if you try to insert anything into it again. If clearing is needed, then this would need to use versioning for the sparse marker rather than just storing the `size_`.

Orvid (changes requested):

folly/container/test/SparseByteSetTest.cpp
```diff
@@ -66,3 +66,15 @@ TEST_F(SparseByteSetTest, each_random) {
     EXPECT_TRUE(s.contains(c));
   }
 }
+
+TEST_F(SparseByteSetTest, clear) {
+  for (auto c = lims::min(); c < lims::max(); ++c) {
+    EXPECT_TRUE(s.add(c));
+  }
+  s.clear();
+  for (auto c = lims::min(); c < lims::max(); ++c) {
```
Orvid, line 75: Swapping this to count down from max() to min() should expose the issue I'm talking about above.

folly/container/SparseByteSet.h
dasfex: Fix return type. 
But I don't understand what are you talking about. Can you explain a little bit more? Why my solution should not to work?

folly/container/test/SparseByteSetTest.cpp
dasfex: This test are passed. Quick example for you: https://godbolt.org/z/vdEvhGWE4

Orvid: Ah, I missed an important part of how this actually works. This should be fine then :)

Orvid:
Yep, it's a known error that we're working on fixing. The internal builds are still working though and do show the new test passing, so this should be good to go :)

dasfex:
@densto88, why are you revert this commit in your fork? :)

PR #1847, opened 2022-08-27
Fixes necessary to build benchmark tests on OSX Monterey

Adenilson:
ToT (Top of Tree) will fail to build if benchmarks are enabled (i.e.  cmake ../ -DBUILD_BENCHMARKS=ON) on Monterey with clang++ 13.1.6.

This patchset will fix the build.


Adenilson:
@Orvid can we merge this one?

PR #1858, opened 2022-09-14
[devel] Add support for Address Sanitizer (ASAN)

Adenilson:
ASAN allows to detect memory corruption issues like overflows/leaks and UaF (Use after Free) being supported by both clang, gcc and MSVC.

This patch adds a new option (i.e. -DASAN_BUILD) to toggle the instrumentation on debug builds.


Adenilson:
@Orvid when you got some spare time, could you please check this one?

Adenilson:
Nevermind, it is unnecessary:
https://github.com/facebook/folly/blob/main/CMake/folly-deps.cmake#L266

PR #1878, opened 2022-10-19
remove __vector_base

DenisYaroshevskiy:
I think this is the fix.

I couldn't get the tests to run but at least the test compiled.


DenisYaroshevskiy:
Recreated as internal pr

PR #1897, opened 2022-12-10
Swap positions of `ptr` and `extra` in PackedSyncPtr

goldsteinn:
Summary:
Change layout:
    [63:48][47:0] ->  [63:16][ 15:0]
    [extra][ ptr] ->  [  ptr][extra]

Previous PackedSyncPtr was organized s.t:
    [63:48][47:0]
    [extra][ ptr]

So access pointer required a 48-bit bitmask:
    `data_.getData() & (-1ull >> 16)`

This isn't great on x86 because imm operands max out at imm32 so this requires two instructions:
    1. movabs + and
    2. movl + andn

Either way its almost certainly cheaper to just just a shift which can be done by swap the placements of `extra` and `ptr` s.t:
    [63:16][ 15:0]
    [  ptr][extra]

As a slightly plus, `extra` can also be accessed essentially for free with `movzwl` (which can move-elim).

On aarch64 `get()` might be slightly more expensive now (and `extra()` correspondingly cheaper) because `lsl` is sometimes slightly slower than `and`.

### DIFF
--- a/folly/PackedSyncPtr.h
+++ b/folly/PackedSyncPtr.h
@@ -77,7 +77,7 @@ class PackedSyncPtr {
   void init(T* initialPtr = nullptr, uint16_t initialExtra = 0) {
     auto intPtr = reinterpret_cast<uintptr_t>(initialPtr);
     CHECK(!(intPtr >> 48));
-    data_.init(intPtr);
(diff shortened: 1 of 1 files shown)


goldsteinn:
ping

PR #1900, opened 2022-12-16
Fix cache_locality_test.Getcpu.VdsoGetcpu on aarch64

wash-amzn:
Summary:
```Getcpu::resolveVdsoFunc()``` will return null when the vdso does not contain the function ```__vdso_getcpu```, which it does only on x86_64. The test was assuming a non-null return, resulting in segfaults on other platforms such as aarch64.

Testing:
I tested this change on x86 and aarch64, both Ubuntu 22.

Example failure on aarch64 before the change:
```
7/9 Test  #322: cache_locality_test.Getcpu.VdsoGetcpu ......................***Exception: SegFault  0.11 sec
Note: Google Test filter = Getcpu.VdsoGetcpu
[----------] 1 test from Getcpu
[ RUN      ] Getcpu.VdsoGetcpu
*** Aborted at 1670948200 (Unix time, try 'date -d @1670948200') ***
*** Signal 11 (SIGSEGV) (0x0) received by PID 54696 (pthread TID 0xffffb5236020) (linux TID 54696) (code: address not mapped to object), stack trace: ***
(error retrieving stack trace)
```

### DIFF
--- a/folly/concurrency/test/CacheLocalityTest.cpp
+++ b/folly/concurrency/test/CacheLocalityTest.cpp
@@ -1034,10 +1034,13 @@ TEST(CacheLocality, BenchmarkSysfs) {
 
 #if defined(FOLLY_HAVE_LINUX_VDSO) && !defined(FOLLY_SANITIZE_MEMORY)
 TEST(Getcpu, VdsoGetcpu) {
-  unsigned cpu;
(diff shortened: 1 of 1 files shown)


wash-amzn:
My employer (Amazon) sent an email to cla@fb.com on Tuesday, Dec 13 to get me added.

AGSaidi:
any chance this will get committed soon?

Issue #1835, opened 2022-08-10
Multiple errors in folly/net/TcpInfo when building with gcc12 for ppc


barracuda156:
```
In file included from /opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfo.h:27,
                 from /opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfo.cpp:18:
/opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfoTypes.h:184:20: error: 'tcp_connection_info' in namespace '::' does not name a type
  184 | using tcp_info = ::tcp_connection_info;
      |                    ^~~~~~~~~~~~~~~~~~~
/opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfoTypes.h:185:31: error: 'TCP_CONNECTION_INFO' was not declared in this scope; did you mean 'TCP_CONNECTIONTIMEOUT'?
  185 | const int tcp_info_sock_opt = TCP_CONNECTION_INFO;
      |                               ^~~~~~~~~~~~~~~~~~~
      |                               TCP_CONNECTIONTIMEOUT
/opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfo.h:273:35: error: 'tcp_info' in namespace 'folly::detail' does not name a type
  273 |   using tcp_info = folly::detail::tcp_info;
      |                                   ^~~~~~~~
/opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfo.h:276:26: error: 'tcp_info' does not name a type; did you mean 'TcpInfo'?
  276 |   explicit TcpInfo(const tcp_info& tInfo)
      |                          ^~~~~~~~
      |                          TcpInfo
/opt/local/var/macports/build/_opt_PPCRosettaPorts_devel_folly/folly/work/folly-v2022.08.01.00/folly/net/TcpInfo.h:289:38: error: expected ',' or '...' before '::' token
  289 |   const T1* getFieldAsPtr(T1 tcp_info::*field) const {
[...]

PR #1907, opened 2023-01-03
Fixes for older macOS and atomic builtins detection

barracuda156:
With these fixes `folly` builds on older macOS, including 10.5.8 and 10.6 on PowerPC. I have put PPC-related ones into a separate commit.

One non-macOS-specific fix is related to atomic builtins detection. Current `folly` configure code does not work correctly: even when atomics requires linking to `libatomic`, the test still passes, as if it is not needed: `Performing Test FOLLY_CPP_ATOMIC_BUILTIN - Success`; sure enough, the build fails at linking stage.
I borrowed the check from here, where it has been tested on macOS and Linux 32-bit: https://github.com/oneapi-src/oneTBB/pull/987


barracuda156:
Those Fedora checks _failing in 0 sec_ cannot be due to the code changes, I guess.

barracuda156:
I will see why it fails now, allow me a bit of time. On Mac it should build in fact, since it builds on Macports buildbots.

barracuda156:
@Orvid Referring to your reply earlier: https://github.com/facebook/folly/issues/1835#issuecomment-1212555583
Could you please take a look? Will be greatly appreciated.

P. S. Non-Mac and non-PPC test result in checks should be irrelevant (I looked at some, and appears they are). I could not have broken anything on new MacOS too, I believe.

folly/File.cpp
```diff
@@ -120,8 +120,10 @@ File File::dupCloseOnExec() const {
     int fd;
 #ifdef _WIN32
     fd = ::dup(fd_);
-#else
+#elseif defined(F_DUPFD_CLOEXEC)
```
Orvid, line 123: `#elseif` doesn't exist, did you mean `#elif`?

folly/SocketAddress.cpp
```diff
@@ -125,6 +125,10 @@ struct GetAddrInfoError {
 
 namespace folly {
 
+#ifndef AI_NUMERICSERV		
+#define AI_NUMERICSERV 0
```
Orvid, line 129: Where would this normally be defined? Most of the compatibility defines for sockets live in [folly/net/NetOps.h](https://github.com/facebook/folly/blob/main/folly/net/NetOps.h).

folly/detail/base64_detail/Base64Api.cpp
```diff
@@ -31,10 +31,16 @@ Base64RuntimeImpl base64EncodeSelectImplementation() {
         base64URLDecodeSWAR};
   }
 #endif
+#if defined(__POWERPC__) // PowerPC BE
```
Orvid, line 34: Does this get defined for PPC64 LE, or only 32-bit PPC BE?

folly/io/async/AsyncUDPSocket.cpp
```diff
@@ -36,6 +36,14 @@
 #define SO_REUSEPORT 15
 #endif
 
+#ifndef IPV6_TCLASS
+#if defined(__GNU__)
+#define IPV6_TCLASS 61
```
Orvid, line 41: Where was this value gotten from?

folly/detail/base64_detail/Base64Api.cpp
Orvid: Also would prefer to check for FOLLY_PPC if this is only ppc 32-bit be.

Orvid (changes requested):

folly/File.cpp
barracuda156: Thank you for noticing, will fix this. (`elseif` crept in from Tcl.)

folly/detail/base64_detail/Base64Api.cpp
barracuda156: @Orvid `__POWERPC__` includes `__ppc__` and `__ppc64__` – only Big endian, and for practical purposes only Darwin. (I think it was also used on BeOS.)

I generally try to be specific, since I have no way to test on *BSD, AIX or Linux, which could also use `ppc32`. Perhaps `FOLLY_PPC_BE` may do, since we want to have macOS in one define, but do not include `ppc64le` (well, not in general case).

folly/io/async/AsyncUDPSocket.cpp
barracuda156: Linux headers: https://github.com/torvalds/linux/blob/master/include/uapi/linux/in6.h#L259
Should I drop it?

folly/SocketAddress.cpp
barracuda156: I will check and move it accordingly, thank you.

folly/detail/base64_detail/Base64Api.cpp
barracuda156: Added a define for non-Apple PPC Big endian. Hope it is correct.

folly/SocketAddress.cpp
barracuda156: @Orvid Moved to the respective header as you recommended.

folly/File.cpp
barracuda156: Fixed.

barracuda156:
@Orvid Please take a look. I updated the patched, taking into account your recommendations, and verified that `folly` latest release builds with them fine on ppc32.

PR #1910, opened 2023-01-07
suppress violations of gcc warning tautological-pointer-compare

yfeldblum:
This suppresses several gcc warning violations of the form:
  warning: comparison of function '{...}' not equal to a null pointer is always true [-Wtautological-pointer-compare]

It is documentsed that the warning may be suppressed by using `&` and comparing the address of the function to a null pointer, rather than comparing the function directly to a null pointer.


yfeldblum:
This won't work on platforms lacking weak symbols since they use function-pointer-typed global variables with nullptr values, but whose *addresses* are not nullptr.

PR #1922, opened 2023-01-22
Fix build with GCC 13 (add missing includes)

thesamesam:
GCC 13 (as usual for new compiler releases) shuffles around some internal includes and so <stdexcept> etc is no longer transitively included.

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/folly/system/AtFork.cpp
+++ b/folly/system/AtFork.cpp
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


thesamesam:
Thank you! There's one for the Watchman side too if feeling generous: https://github.com/facebook/watchman/pull/1089. Cheers again.

PR #1933, opened 2023-02-14
Use folly::IsRelocatable in folly::Function

Quuxplusone:
The `Op::MOVE` operation here is exactly "move-construct plus destroy," which is the same operation optimized into memcpy by `FBVector::make_window` and `FBVector::reserve`. So we can optimize it into memcpy here for the same set of types, namely, trivially relocatable types.

I haven't figured out how to run `FunctionTest.cpp` locally, so someone should look closely at whether the test change actually even works at all.

### DIFF
--- a/folly/Function.h
+++ b/folly/Function.h
@@ -747,7 +747,7 @@ class Function final : private detail::function::FunctionTraits<FunctionType> {
               noexcept(Fun(FOLLY_DECLVAL(Fun))))>
   /* implicit */ Function(Fun fun) noexcept(IsSmall) {
     using Dispatch = conditional_t<
-        IsSmall && is_trivially_copyable_v<Fun>,
+        IsSmall && folly::IsRelocatable<Fun>::value,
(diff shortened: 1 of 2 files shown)


Quuxplusone:
Waaait a minute, actually this is incorrect. We'd have to make a new `DispatchSmallRelocatable`, where `Op::MOVE` is a memcpy but `Op::NUKE` actually calls the destructor of the type.

GitHub doesn't let me "plan changes" on my own PR, so I'm going to abandon this for now and reopen it later with a proper patch. (And also figure out why my new test didn't fail! The Linux build is green, which seems wrong, if it actually ran my test.)

Quuxplusone:
Okay, I've pushed a better patch to Quuxplusone:reloc-in-function. But now GitHub doesn't let me reopen my own PR! Well, I don't think this patch is likely to be accepted anyway, given that it adds a fiddly codepath to `memcpy` in a place where the compiler can probably achieve the same effect just by inlining.

PR #1935, opened 2023-02-17
Add address sanitizer annotations when building with MS compiler

pps83:
I use UninitializedMemoryHacks.h, but it adds lots of false positives when running with address sanitizer. I grepped MS includes and based on the version that comes with VS2022 I added these annotations. In my case, this fixed all the issues. I did check vs 2019 headers and didn't see the same annotation code used there.


pps83:
check __sanitizer_annotate_contiguous_container in https://github.com/microsoft/STL/blob/main/stl/inc/string and https://github.com/microsoft/STL/blob/main/stl/inc/vector

pps83:
based on git history, seems that these are available only in vs 2022:
image

pps83:
FYI, it looks like in current version that ships with latest VS2022 asan annotations are commented out in `std::string`. I had some false positives because of that with asan reporting and had to comment out the string part in UninitializedMemoryHacks.h

pps83:
More fixes added (support for std::string annotations was added starting from _MSC_VER >= 1938)

BillyONeal:
Has folly considered asking for an _Ugly member that does what they need rather than trying to incapability declare standard library parts?

folly/memory/UninitializedMemoryHacks.h
```diff
@@ -254,13 +285,19 @@ struct MakeUnsafeStringSetLargerSize {
...
-      void (std::basic_string<TYPE>::*)(std::size_t),           \
-      &std::basic_string<TYPE>::_Eos>;                          \
+#if (_MSC_VER >= 1938)
+#define Eos_spec_ noexcept
+#else
+#define Eos_spec_
+#endif
+
+#define FOLLY_DECLARE_STRING_RESIZE_WITHOUT_INIT(TYPE)                \
+  template void std::basic_string<TYPE>::_Eos(std::size_t) Eos_spec_; \
```
yfeldblum, line 295: No need for a macro. Can just `noexcept(_MSC_VER >= 1938)`.

folly/memory/UninitializedMemoryHacks.h
```diff
@@ -110,6 +130,11 @@ inline void resizeWithoutInitialization(
       s.reserve(n);
     }
     detail::unsafeStringSetLargerSize(s, n);
+#if defined(_MSC_VER) && _MSC_VER >= 1938 && defined(_INSERT_STRING_ANNOTATION)
+    if (_Asan_string_should_annotate && s.capacity() > std::string().capacity())
```
yfeldblum, line 134: Where does the symbol `_Asan_string_should_annotate` come from?

folly/memory/UninitializedMemoryHacks.h
```diff
@@ -110,6 +130,11 @@ inline void resizeWithoutInitialization(
...
+      __sanitizer_annotate_contiguous_container(&s[0], &s[0] + s.capacity() + 1,
```
yfeldblum, line 135: We already wrap some of the sanitizer interface. We exclusively use the wrappers.

folly/memory/UninitializedMemoryHacks.h
```diff
@@ -21,11 +21,30 @@
 #include <type_traits>
 #include <vector>
 
-// On MSVC an incorrect <version> header get's picked up
+// On MSVC an incorrect <version> header gets picked up
 #if !defined(_MSC_VER) && __has_include(<version>)
 #include <version>
 #endif
 
+#if defined(_MSC_VER) && !defined(_M_CEE_PURE) && !defined(_DISABLE_STRING_ANNOTATION)
```
yfeldblum, line 29: Is `_DISABLE_STRING_ANNOTATION` some well-known symbol? If not and this is supposed to be specific to folly, then it must be namespaced to folly.

yfeldblum (changes requested):

folly/memory/UninitializedMemoryHacks.h
yfeldblum: Expect in specific cases. For example, the existing calls to `__sanitizer_annotate_contiguous_container` below are exactly as libc++ does, located in the per-library implementation for libc++.

folly/memory/UninitializedMemoryHacks.h
```diff
@@ -110,6 +130,11 @@ inline void resizeWithoutInitialization(
       s.reserve(n);
     }
     detail::unsafeStringSetLargerSize(s, n);
+#if defined(_MSC_VER) && _MSC_VER >= 1938 && defined(_INSERT_STRING_ANNOTATION)
+    if (_Asan_string_should_annotate && s.capacity() > std::string().capacity())
+      __sanitizer_annotate_contiguous_container(&s[0], &s[0] + s.capacity() + 1,
+          &s[0] + n0 + 1, &s[0] + n + 1);
+#endif
```
yfeldblum, line 137: This looks like library-specific implementation that must vary by library and library version. So it should be in the relevant per-library implementations, below. For clarity, see where the existing calls to `__sanitizer_annotate_contiguous_container` are located.

folly/memory/UninitializedMemoryHacks.h
pps83: I looked into the code of std::string how they do it and replicated code pretty much

pps83:
@yfeldblum thanks for the review. I'm traveling in remote locations and rarely connect, so, will take some time before I can address it. There was also a suggestion by @BillyONeal  and I think that would be a better approach.

BillyONeal:
> There was also a suggestion by @BillyONeal and I think that would be a better approach.

To be clear, while in the long term doing this with some mechanism that isn't a hack would have the benefit that it doesn't break every 3 seconds.

That doesn't change that VS2022 17.8 is already out there which means Windows customers who want to use Folly are broken

pps83:
`_INSERT_STRING_ANNOTATION` / `_INSERT_VECTOR_ANNOTATION` are defined by https://github.com/microsoft/STL/blob/0403d19f5461fd15983737c3f01ec34800ea9275/stl/inc/__msvc_sanitizer_annotate_container.hpp (this file is included by string/vector). string/vector undef these defines at the end. This PR replicates what the MS headers do.

`_Asan_string_should_annotate` also comes from __msvc_sanitizer_annotate_container.hpp

pps83:
@yfeldblum ping

Orvid:
There are merge conflicts in this that need to be resolved before the tooling will let me import this PR.

pps83:
> There are merge conflicts in this that need to be resolved before the tooling will let me import this PR. Would you be able to rebase this PR?

give me a few mins

pps83:
I rebased, but I didn't have any merge conflicts.

Orvid:
Thanks, the import worked that time. Not sure why it was having issues before if there weren't merge conflicts.

PR #1944, opened 2023-03-04
Update x86 Arch Detection

assignUser:
The old detection set `IS_X86_ARCH=TRUE` on Apple M1 causing compilation to fail due to the erroneous `-msse4.2` flag. Several reports of this issue can be found here: https://github.com/facebookincubator/velox/issues/3453

This PR uses `CMAKE_SYSTEM_PROCESSOR`  which contains the result of `uname -m` to replace the previous detection mechanism. Additionally it allows users to override the variable should there be any detection issues (e.g. running in rosetta).


assignUser:
The packit fails appear to be unrelated:
```
+ /usr/bin/patch --no-backup-if-mismatch -f -p1 --fuzz=0
patching file folly/executors/test/CodelTest.cpp
Hunk #1 FAILED at 41.
1 out of 1 hunk FAILED -- saving rejects to file folly/executors/test/CodelTest.cpp.rej
error: Bad exit status from /var/tmp/rpm-tmp.hHrf1f (%prep)
```

assignUser:
Will add `AMD64` for correct detection on windows. Though I don't know why the build failed on windows as https://github.com/facebook/folly/blob/fix_arch_detection/CMakeLists.txt#L254-L261 is visible in the logs.

The detection worked correctly on macos I am unsure if the test failure in "AsyncFileWriter.fork" is related: https://github.com/facebook/folly/actions/runs/4328417552/jobs/7597646813

PR #1952, opened 2023-03-09
[Minor] Fix typos in ConcurrentHashMap.h

Aiee:
As title.

### DIFF
--- a/folly/concurrency/ConcurrentHashMap.h
+++ b/folly/concurrency/ConcurrentHashMap.h
@@ -30,7 +30,7 @@ namespace folly {
  * support erase and update.
  *
  * Readers are always wait-free.
- * Writers are sharded, but take a lock.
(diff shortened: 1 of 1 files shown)


folly/concurrency/ConcurrentHashMap.h
```diff
@@ -30,7 +30,7 @@ namespace folly {
  * support erase and update.
  *
  * Readers are always wait-free.
- * Writers are sharded, but take a lock.
+ * Writers are shared, but take a lock.
```
Orvid, line 33: This was correct as it was. Writers take a lock, but only lock part of the map, so are sharded.

Orvid (changes requested):
The rest looks good, just need to drop the first change.

Orvid (approved):

PR #1992, opened 2023-04-28
update disable ubsan macro

urvishdesai:
no_sanitize attribute now supports undefined-behavior sanitizers including `no_sanitize("unsigned-integer-overflow")`, `no_sanitize("undefined")` etc. 

In the current code, running `__attribute__((no_sanitize("unsigned-integer-overflow")))` for ASan, MSan, or TSan is enabled leads to the warnings for attribute ignored. The logic here is hence incorrect. This kind of an attribute should be used only when UBSan is enabled.

### DIFF
--- a/folly/CPortability.h
+++ b/folly/CPortability.h
@@ -151,22 +151,34 @@
 #endif
 
 /**
+ * Define a convenience macro to test when undefined-behavior sanitizer is being used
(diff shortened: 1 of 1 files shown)


Orvid:
Yep, working on it. This PR got lost in my pile of things to do. It should hopefully be merged either later today or on monday.

PR #2005, opened 2023-05-26
Remove CXX_STD and inherit CMAKE_CXX_STANDARD from parent project

mhx:
This follows up on a discussion in #1949 and the comment by @ot that the C++ standard for folly should be (easily) settable by a parent project.

There are two commits in this PR:

- The first commit removes the redundant `CXX_STD`, which causes `-std=` to be passed to the compiler twice.
- The second commit introduces a check before setting `CMAKE_CXX_STANDARD`, so a value set by a parent project won't be overridden.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,8 +90,11 @@ set(FOLLY_SUPPORT_SHARED_LIBRARY "${BUILD_SHARED_LIBS}")
 include(FBBuildOptions)
 fb_activate_static_library_option()
 
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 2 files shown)


mhx:
ping

CMakeLists.txt
```diff
@@ -90,8 +90,11 @@ set(FOLLY_SUPPORT_SHARED_LIBRARY "${BUILD_SHARED_LIBS}")
 include(FBBuildOptions)
 fb_activate_static_library_option()
 
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
```
eugene536, line 94: it would be better to set `CMAKE_CXX_STANDARD` as a `CACHE` variable and remove the `if`
```
set(CMAKE_CXX_STANDARD "17" CACHE STRING  "folly C++ standard")
```

PR #2010, opened 2023-05-31
fix deadlock in RCU retire when rcu_reader active

nbronson:
Summary:
Every 3.2 seconds RCU retire tries to opportunistically drain the queue of pending work.  If this happens while the current thread is holding an rcu_reader and another thread is currently running synchronize(), a deadlock will occur.  This diff adds a unit test for the behavior and fixes the problem.

Test Plan:
* unit test that reproduces the problem
* unit test passes with the fix

### DIFF
--- a/folly/synchronization/Rcu.h
+++ b/folly/synchronization/Rcu.h
@@ -382,8 +382,9 @@ class rcu_domain {
         syncTime_.compare_exchange_strong(
             syncTime, time, std::memory_order_relaxed)) {
       list_head finished;
-      {
-        std::lock_guard<std::mutex> g(syncMutex_);
+      // synchronize() blocks while holding syncMutex_,
+      // so we must not wait for syncMutex_
+      if (std::unique_lock<std::mutex> g(syncMutex_, std::try_to_lock)) {
         half_sync(false, finished);
       }
       // callbacks are called outside of syncMutex_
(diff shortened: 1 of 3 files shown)


folly/synchronization/Rcu.h
```diff
@@ -383,8 +383,12 @@ class rcu_domain {
...
       {
-        std::lock_guard<std::mutex> g(syncMutex_);
-        half_sync(false, finished);
+        // synchronize() blocks while holding syncMutex_,
+        // so we must not wait for syncMutex_
+        std::unique_lock<std::mutex> g(syncMutex_, std::try_to_lock);
+        if (g) {
+          half_sync(false, finished);
+        }
       }
```
ot, line 392: We can avoid the double indent this way?

```
if (std::unique_lock<std::mutex> g(syncMutex_, std::try_to_lock)) {
  half_sync(false, finished);
}
```

folly/synchronization/Rcu.h
```diff
@@ -383,8 +383,12 @@ class rcu_domain {
             syncTime, time, std::memory_order_relaxed)) {
```
ot, line 383: If the lock acquire fails, should we roll back `syncTime_` somehow?

folly/synchronization/Rcu.h
nbronson: Sure, I'll change it

folly/synchronization/Rcu.h
nbronson: If the lock fails somebody else must be running synchronize, so the purpose of syncTime_ has been accomplished

nbronson:
@ot, is anybody at Meta working on a functioning implementation of `rcu_barrier()`? If not, would you be willing to take it as an external contribution? Rockset has a barrier implementation we have been using that uses a second RCU domain to track pending cleanup tasks. It doubles the memory and TLS footprint of the domain, but has essentially no performance impact (one RCU lock + unlock per cleanup list node).

PR #2022, opened 2023-06-20
Fix libfmt errors from not finding enum formatter

mhx:
Recent versions of libfmt have become more strict and require `enum` types to be formattable:

    static assertion failed due to requirement 'formattable': Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt

This is a quick fix to simply use the underlying type.

### DIFF
--- a/folly/futures/detail/Core.cpp
+++ b/folly/futures/detail/Core.cpp
@@ -19,6 +19,7 @@
 #include <new>
 
 #include <fmt/core.h>
+#include <folly/Utility.h>
 #include <folly/lang/Assume.h>
 
 namespace folly {
@@ -30,7 +31,7 @@ namespace {
 template <class Enum>
 void terminate_unexpected_state(fmt::string_view context, Enum state) {
   terminate_with<std::logic_error>(
-      fmt::format("{} unexpected state: {}", context, state));
+      fmt::format("{} unexpected state: {}", context, to_underlying(state)));
 }
 
 } // namespace


ot:
Thanks for the fix, I think that `fmt::underlying` would be a terser option, but it's not clear from the documentation whether it's available in `fmt/core.h` or we need `fmt/format.h`.

@vitaut Do you have recommendations?

mhx:
> Thanks for the fix, I think that `fmt::underlying` would be a terser option [...]

Agreed, I just wasn't sure if this is already in libfmt 8.x which folly currently lists in its build manifest. At least libfmt 10.x defines `underlying` in `fmt/format.h`.

vitaut:
As [documented](https://fmt.dev/latest/api.html#_CPPv4I0EN3fmt10underlyingE12underlying_tI4EnumE4Enum), `fmt::underlying` is defined in `fmt/format.h`. You could also use `std::to_underlying` instead.

Orvid:
Folly has [folly::to_underlying](https://github.com/facebook/folly/blob/4a892bbcf525367b910d3fa6e05704db68c8eeea/folly/Utility.h#L708) which does this without the need to worry about C++ language versions, or where `fmt::underlying` lives. I would recommend just using that.

mhx:
Updated PR using `folly::to_underlying`.

PR #2036, opened 2023-07-10
Support ASAN suppressions under MSVC

Orvid:
Summary: MSVC has support for asan, so enable the ASAN suppression macros where needed.

Differential Revision: D47345965


yfeldblum:
Merged in c56cdac808dda262dd0318207195c7e2be96d722.

PR #2040, opened 2023-07-17
Update Varint.h

duxing2007:
fix Varint.h ZigZag comment


yfeldblum:
Thanks. This PR was indeed opened first. However, this fix was done in #2072 for #2046 via b1988081a256cd33ce1794a162c2c41b98d6bf69.

PR #2052, opened 2023-08-06
F14MapTest.cpp: fix compilation fail

xnhp0320:
The 'dest1' has been deduced to two parameters: std:pair<int, double>
and std::initializer_list, not an initializer_list of std::pair.

add an extra braces to pass compilation. 
Not sure if this is a compiler bug. 

# gcc --version
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0


Orvid:
This is likely the result of a compiler bug, as the way it is currently written is intentional.

Excluding this test for the versions of GCC where it fails to compile would be a reasonable solution here.

PR #2084, opened 2023-10-16
fix windows build

ahornby:
fix windows build

Test plan:

build locally with VS 2019 Build Tools

before, broken

after, works


Orvid:
This will need to be rebased before the tooling will let me import it.

PR #2085, opened 2023-10-17
Fix tvOS build

douglowder:
Changes to fix compile errors on tvOS:

- Add `TARGET_OS_TV` to Apple platforms in `folly-config.h.cmake`
- Exclude TV from adding duplicate time and clock definitions in `Time.h`

Test strategy:

- Build and test RNTester for `react-native-tvos` fork 0.73 branch with the above Folly changes
- Apple TV should compile and run


folly/portability/Time.h
```diff
@@ -40,7 +40,7 @@
 
 // These aren't generic implementations, so we can only declare them on
 // platforms we support.
-#if !FOLLY_HAVE_CLOCK_GETTIME && (defined(__MACH__) || defined(_WIN32))
+#if !FOLLY_HAVE_CLOCK_GETTIME && (defined(__MACH__) || defined(_WIN32)) && !(defined(TARGET_OS_TV) && TARGET_OS_TV)
```
NickGerleman, line 43: FWIW this is broken on iOS with XCode as well. Needed to ad `FOLLY_HAVE_CLOCK_GETTIME` explicitly in RN ☹️.

douglowder: Could we not add this in `folly-config.h.cmake`?

NickGerleman: Which part? The iOS RN build doesn’t use CMake.

douglowder: Aha now I see the source of my confusion... I was testing Folly from main with RN from 0.73, the Podfile for that branch is different. I'll test with that, and make changes to this PR. Thanks!!!

douglowder: @NickGerleman thanks for the help with this -- I tested this today, and if the RCT-Folly podspec has `FOLLY_HAVE_CLOCK_GETTIME=1` added, this patch is not needed. Closing the PR.

douglowder:
Not needed (see inline comments). Closing.

PR #2089, opened 2023-10-20
fix path errors in commonts

liuqinfei:
Fix the errors in commonts.


yfeldblum:
Thanks! Done separately in 6fff2142be20c9329cfd9524a11f1d79683ecccc.

Issue #2097, opened 2023-11-13
folly::ConcurrentHashMap crashes under high contention


DanielLiamAnderson:
When under significant contention and with a specific key distribution, `folly::ConcurrentHashMap` fails checks inside `HazptrObjLinked.h`.  I've reproduced this on multiple machines.

**OS:** Ubuntu 20.04.6 LTS  (reproduced on Ubuntu 22.04)
**Compiler:** GCC 13.1.0  (reproduced on GCC 11.4.0)

The following minimal reproducible example consistently reproduces the bug.

```c++
#include <cmath>
#include <cstdint>

#include <latch>
#include <limits>
#include <random>
#include <thread>
#include <vector>

#include <folly/concurrency/ConcurrentHashMap.h>
#include <folly/init/Init.h>


struct KeyDist {

  KeyDist(uint64_t num_items, double zipfian_const) :
      items_(num_items), theta_(zipfian_const), zeta_n_(num_items) {
    zeta_2_ = Zeta(2, theta_);
    zeta_n_ = Zeta(num_items, theta_);
    alpha_ = 1.0 / (1.0 - theta_);
    eta_ = Eta();
  }

  uint64_t operator () (size_t i) const {
    uint64_t r = hash64(i);
    double u = static_cast<double>(r) / static_cast<double>(std::numeric_limits<uint64_t>::max()); // uniform between 0.0 and 1.0
    double uz = u * zeta_n_;
    if (uz < 1.0) return 0;
    if (uz < 1.0 + std::pow(0.5, theta_)) return 1;
    return std::lround((items_-1) * std::pow(eta_ * u - eta_ + 1, alpha_));
  }

  double Eta() {
    return (1 - std::pow(2.0 / items_, 1 - theta_)) / (1 - zeta_2_ / zeta_n_);
  }

  static uint64_t hash64(uint64_t u) {
    uint64_t v = u * 3935559000370003845ul + 2691343689449507681ul;
    v ^= v >> 21;
    v ^= v << 37;
    v ^= v >> 4;
    v *= 4768777513237032717ul;
    v ^= v << 20;
    v ^= v >> 41;
    v ^= v << 5;
    return v;
  }

  static double Zeta(uint64_t cur_num, double theta) {
    double result = 0.0;
    for (int i = 0; i < cur_num; i++) {
      result += 1.0/ std::pow(i+1, theta);
    }
    return result;
  }

  uint64_t items_;
[...]

PR #2107, opened 2023-11-30
Safeguard against link overflows in ConcurrentHashMap

instw:
Summary:
PROBLEM
Folly ConcurrentHashMaps use Hazard pointers to ensure map entries that were recently removed (using `erase`, `insert_or_assign`, etc) aren't cleaned up when there are readers for those objects. Instead, they are removed as part of a reclamation process which typically happens asynchronously. Moreover within ConcurrentHashMap, entries are linked to one another, and this linkage needs to be known within the hazard pointer logic to ensure we don't clean up an object that itself doesn't have any direct hazard pointers, but is referenced by another object that might have hazard pointers. That logic is within `HazptrObjLinked`.
Under high contention situations (see https://github.com/facebook/folly/issues/2097 ) , the link counting logic can overflow, because a single object has too many dangling links. For example, consider a map that has 2 entries with the same hash code- `(A,0)` and `(B,0)`. Let's assume that `A` is stored before `B` internally within the `ConcurrentHashMap`'s `BucketTable`. `B` stores initially that it has a 1 link (to `A`). Now, let's assume that we replace `(A,0)` with `(A,1)`. While `(A,0)` is erased out of the `ConcurrentHashMap`, its not immediately reclaimed/deleted. During this interim, `B` has a link count of 2 to the 2 entries of `A`. This link count is stored as a 16 bit unsigned integer. If the above operation happens very quickly, then we end up in a situation where `B`'s link count overflows past 65535, and wraps around.
This situation is caught in debug compilation (due to `DCHECK`), but in opt builds, it results in bad retirements. For eg, if `B`'s link count goes past 65535 to 65537 (i.e. `1`), then when 1 object of `A` is reclaimed, the `B`'s link count would decrement past `1` back to `0`, causing `B` to be incorrectly retired. Now if we actually end up removing all of `A`, the link count will overflow backwards, from `0` back to `65535` and then back to `0`, causing a double retirement - a sign to corruption.

SOLUTION
While the situation is rare, it can arise for skewed data with a lot of contention. There are 3 options to "solve" this:
1. Increase the link count data structure size from 16bit to something higher - Simple, but a work-around. Eventually high-enough contention would bugs to show up there as well.
2. Crash the process when there is very high contention - Maintains the current performance guarantees, and when ConcurrentHashMap cannot meet those guarantees, it causes a fatal error.
3. Slow ConcurrentHashMap erasures under high contention (this diff) - Very high contention would cause ConcurrentHashMap to slow down, and give reclamation time to act. Functionally `ConcurrentHashMap` remains the same, but does exhibit different perf characteristics.

In this change, the `HazptrObjLinked` code is changed is disallow for overflows since it leads to corruption, and the callers are responsible for handling cases where links cannot be created. For `ConcurrentHashMap`, we keep waiting, until we can acquire a link : which means erasures under high contention are lock-free but not wait-free.
For reclamation, there are buffers within the cohort to store both retired objects (aka `list`) and reclaimed objects (aka `safe list`). In cases where `ConcurrentHashMap` is unable to acquire a link, it's imperative it tries to initiate a reclamation cycle to make progress, and thus I added a `cleanup()` method within the cohort to flush any existing retired objects to the hazard pointer domain for retirement-evaluation, kick off a reclamation cycle, and also retire any retired objects pending within the cohort.

Differential Revision: D51647789

### DIFF
--- a/folly/concurrency/detail/ConcurrentHashMap-detail.h
+++ b/folly/concurrency/detail/ConcurrentHashMap-detail.h
@@ -224,7 +224,7 @@ class NodeT : public hazptr_obj_base_linked<
     this->set_deleter( // defined in hazptr_obj
         concurrenthashmap::HazptrDeleter<Allocator>());
     this->set_cohort_tag(cohort); // defined in hazptr_obj
-    this->acquire_link_safe(); // defined in hazptr_obj_base_linked
+    CHECK(this->acquire_link_safe()); // defined in hazptr_obj_base_linked
   }
 
   ValueHolder<KeyType, ValueType, Allocator, Atom> item_;
@@ -361,7 +361,11 @@ class alignas(64) BucketTable {
         }
       }
       // Set longest last run in new bucket, incrementing the refcount.
-      lastrun->acquire_link(); // defined in hazptr_obj_base_linked
+      while (!lastrun->acquire_link()) {
+        cohort->cleanup();
+        std::this_thread::yield();
+      }
+
       newbuckets->buckets_[lastidx]().store(lastrun, std::memory_order_relaxed);
       // Clone remaining nodes
       for (; node != lastrun;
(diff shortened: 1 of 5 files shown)


instw:
CC @magedm - Does this look like a reasonable fix?

DanielLiamAnderson:
I ran this against our benchmark suite in which we initially observed the bug ([ParlayHash](https://github.com/cmuparlay/parlayhash)) and it does indeed appear to be fixed!  Thanks for working on it.

PR #2109, opened 2023-12-19
folly::tape

DenisYaroshevskiy:
Summary:
A better version of `vector<vector>` for some scenarios.
Can also be an ok `vector<string>` if not everything is small vectors

Basic usecase: building vector<vector> by pushing one element at a time, no reserve.
```
PushBackVecInts_20_0_15                                     1.28us   782.19K
PushBackTapeInts_20_0_15                                  317.49ns     3.15M
```

Constructing a vector of strings (20 strings). When  everything is SSO, there is no point in tape. However, when we allocate, tape begins to win big.

```
ConstructVecSmallStrings_20_0_15                          122.31ns     8.18M
ConstructTapeSmallStrings_20_0_15                         129.04ns     7.75M
ConstructVecLargeStrings_20_16_32                         319.24ns     3.13M
ConstructTapeLargeStrings_20_16_32                        147.52ns     6.78M
```

Worst case scenario - 2 small strings. Overhead of tape makes this slower. We could tackle this but it requires very complicated code and so far, not the use case encountered.
```
ConstructVec2SmallStrings                                  20.95ns    47.74M
ConstructTape2SmallStrings                                 38.83ns    25.75M
```

Differential Revision: D52260714

### DIFF
--- a/folly/container/test/tape_bench.cpp
+++ b/folly/container/test/tape_bench.cpp
@@ -1,0 +1,134 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


folly/container/test/tape_bench.cpp
```diff
@@ -0,0 +1,134 @@
...
+using vv_int = std::vector<std::vector<int>>;
+using tv_int = folly::tape<std::vector<int>>;
+
+template <typename Cont>
+struct ContGenerator {
+  ContGenerator(std::size_t minLen, std::size_t maxLen)
+      : g(minLen + maxLen), len(minLen, maxLen) {}
+
+  const Cont& operator()() {
+    buf.resize(len(g));
```
tekawade, line 38: @DenisYaroshevskiy shouldn't this be `buf.reserve(len(g));` instead?

`resize` + `emplace_back()` will increase the size. While `reserve` will reserve the space and array won't grow dynamically as often.

PR #2112, opened 2023-12-20
symbolizer: add support for looking up CUs using the .gdb_index if present

oremanj:
This provides similar information as `.debug_aranges` but is better supported on DWARFv4 and with clang. Since it is generated at link time, it has the additional benefit that it is complete, meaning it can return negative results (for symbols such as `_start` that may be implemented in assembly and thus have no associated debug info) without walking the full debuginfo list.


Orvid:
How was this tested? Would it be possible to add to the existing test suite?

PR #2119, opened 2024-01-04
Add delete for SparseByteSet.

dasfex:
Added the `delete` method for SparseByteSet class. 

Notes on how it works. Assume we have next operations done:
```
add(0)
add(1)
add(2)
add(3)
add(4)
```
So now we have next structure:
```
dense:  0 1 2 3 4 |
sparse: 0 1 2 3 4
```
And `size_` equals to 5.

We have two options when deleting. 
1. Deleted element are in the end of our `dense` array. Here we need to decrease `size_`. F.e. let's try to delete `4`:
```
dense:  0 1 2 3 | 4
sparse: 0 1 2 3 4
```
2. Deleted element are not in the end of `dense`. Here we need to swap it with the last element and decrease `size_`. Let's try to delete `1`. After swap:
```
dense:  0 3 2 1 | 4
sparse: 0 3 2 1 4
```
After `size_` decrease:
```
dense:  0 3 2 | 1 4
sparse: 0 3 2 1 4
```

That's all : )

### DIFF
--- a/folly/container/SparseByteSet.h
+++ b/folly/container/SparseByteSet.h
@@ -31,6 +31,7 @@ namespace folly {
  *
  *  Operations:
  *  - add(byte)
+ *  - remove(byte)
  *  - contains(byte)
  *  - clear()
  *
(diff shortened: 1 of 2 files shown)


dasfex:
@Orvid, hey! Can u check it? :)

PR #2126, opened 2024-01-15
Fix C++ Atomic support check

luhenry:
Some platforms need libatomic linked for sub-word operations while others need it for super-word operations. Let's make sure we test for all.

Relates to https://github.com/facebook/folly/pull/1907


barracuda156:
@luhenry Thank you for addressing this!

barracuda156 (commented):
Maybe also modify a comment at line 184:
```
# Using clang with libstdc++ requires explicitly linking against libatomic
```
To something like:
```
# Using clang with libstdc++ requires explicitly linking against libatomic.
# gcc requires explicit linking against libatomic on some platforms.
```

PR #2132, opened 2024-01-24
Add a couple of throws for bad states

r-barnes:
Note that both of these fix a compilation warning.

### DIFF
--- a/folly/experimental/io/EpollBackend.cpp
+++ b/folly/experimental/io/EpollBackend.cpp
@@ -94,7 +94,9 @@ void SignalRegistry::notify(int sig) {
   int fd = notifyFd_.load();
   if (fd >= 0) {
     uint8_t sigNum = static_cast<uint8_t>(sig);
-    ::write(fd, &sigNum, 1);
+    if (::write(fd, &sigNum, 1) != 1) {
+      throw std::runtime_error("Failed to write all the byes.");
+    }
   }
 }
 
--- a/folly/experimental/settings/Types.cpp
+++ b/folly/experimental/settings/Types.cpp
@@ -16,6 +16,8 @@
 
 #include <folly/experimental/settings/Types.h>
 
+#include <string_view>
+
 namespace folly {
 namespace settings {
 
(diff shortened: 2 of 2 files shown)


folly/experimental/io/EpollBackend.cpp
```diff
@@ -94,7 +94,9 @@ void SignalRegistry::notify(int sig) {
   int fd = notifyFd_.load();
   if (fd >= 0) {
     uint8_t sigNum = static_cast<uint8_t>(sig);
-    ::write(fd, &sigNum, 1);
+    if (::write(fd, &sigNum, 1) != 1) {
+      throw std::runtime_error("Failed to write all the byes.");
+    }
```
yfeldblum, line 99: This looks like the type of `write` that is infallible, i.e., cannot fail (as long as the inputs are correct), and total, i.e. cannot be a partial write.

And if this `write` wre to fail, there is nothing that would know how to clean this up.

Failure would be an invariant violation. If we are looking for invariant violations, such as we might do in debug builds, we would use something in the family of `assert` (eg `DCHECK`, `PCHECK`, `FOLLY_SAFE_DCHECK`, etc).

Partial writes may occur if the backend has insufficient resources to accept the full write, and if the write is non-blocking or is interrupted by a signal while blocking. But this is not the sort of write where the backend could have insufficient resources.

folly/experimental/settings/Types.cpp
```diff
@@ -27,6 +29,8 @@ std::string_view toString(SetErrorCode code) {
       return "rejected";
     case SetErrorCode::FrozenImmutable:
       return "frozen immutable";
+    default:
+      throw std::invalid_argument("Code '" + std::to_string(static_cast<int>(code)) + "' had no string representation!");
```
yfeldblum, line 33: IMO this should not be an error - it's for diagnostics.

We should like to return `"(unknown)"` in case of an unrecognized input. But we should also like the compiler to warn us when the `case`s in the `switch` are non-exhaustive.

yfeldblum (changes requested):

PR #2145, opened 2024-02-15
Allow multiple flags to be passed into toAppend for double conversion

danielin917:
Summary:
Currently the function argument accepts an enum meaning only one flag can be passed in. We should allow multiple flags to be passed in and this is specifically required in order to use EMIT_TRAILING_ZERO_AFTER_POINT.


yfeldblum:
An [`enum`](https://github.com/google/double-conversion/blob/v3.3.0/double-conversion/double-to-string.h#L75-L84) still allows multiple flags to be passed in.

PR #2147, opened 2024-02-20
Fix 3 bugs related to libc++

yaoxinliu:
1. `std::terminate()` is declared in `<exception>`; if not `#include <exception>`, the compiler cannot recognize `std::terminate()`.

2. return a local variable to an l-value reference is an error.

3. `std::unique_ptr<T>` requires T must not be an incomplete type.

### DIFF
--- a/folly/experimental/coro/Traits.h
+++ b/folly/experimental/coro/Traits.h
@@ -152,8 +152,8 @@ template <
             folly::Negation<detail::_has_member_operator_co_await<Awaitable>>>::
             value,
         int> = 0>
-Awaitable&& get_awaiter(Awaitable&& awaitable) {
-  return std::forward<Awaitable>(awaitable);
+Awaitable& get_awaiter(Awaitable&& awaitable) {
+  return static_cast<Awaitable&>(awaitable);
 }
 
 template <
--- a/folly/experimental/coro/Traits.h
+++ b/folly/experimental/coro/Traits.h
@@ -152,8 +152,8 @@ template <
             folly::Negation<detail::_has_member_operator_co_await<Awaitable>>>::
             value,
         int> = 0>
-Awaitable& get_awaiter(Awaitable&& awaitable) {
-  return awaitable;
+Awaitable&& get_awaiter(Awaitable&& awaitable) {
+  return std::forward<Awaitable>(awaitable);
 }
 
 template <
--- a/folly/experimental/coro/Traits.h
+++ b/folly/experimental/coro/Traits.h
@@ -147,6 +147,7 @@ template <
     typename Awaitable,
     std::enable_if_t<
         folly::Conjunction<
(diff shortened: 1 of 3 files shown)


folly/experimental/coro/Traits.h
```diff
@@ -152,8 +152,8 @@ template <
             folly::Negation<detail::_has_member_operator_co_await<Awaitable>>>::
             value,
         int> = 0>
-Awaitable& get_awaiter(Awaitable&& awaitable) {
-  return awaitable;
+Awaitable&& get_awaiter(Awaitable&& awaitable) {
+  return std::forward<Awaitable>(awaitable);
 }
```
Quuxplusone, line 157: I don't know anything about how this code is used, but it's a little scary that line 191 of this file has
```
template <typename Awaitable>
struct awaiter_type<Awaitable, std::enable_if_t<is_awaitable_v<Awaitable>>> {
  using type = decltype(get_awaiter(std::declval<Awaitable>()));
};
```
So IIUC, you're changing the meaning of `awaiter_type<A>::type` from an lvalue reference to an rvalue reference. Is that part of your intent?

If your goal is simply to fix the compiler warning and make this code C++23-compatible ("Simpler Implicit Move") without changing anything about its current behavior at all, then the correct fix is:
```
Awaitable& get_awaiter(Awaitable&& awaitable) {
  return static_cast<Awaitable&>(awaitable);
}
```

folly/tracing/AsyncStack.cpp
```diff
@@ -19,6 +19,7 @@
 #include <atomic>
 #include <cassert>
 #include <mutex>
+#include <exception>
```
Quuxplusone, line 22: Nit: this should probably be alphabetized after "cassert" and before "mutex"

Orvid: This will be handled automatically internally - we have clang-format with include sorting enabled there, though I'm not sure if we have the configs in the folly repo itself.

folly/experimental/coro/Traits.h
yaoxinliu: I agree with you.

yaoxinliu:
```
template <
    typename Awaitable,
    std::enable_if_t<
        folly::Conjunction<
            !std::is_rvalue_reference<Awaitable&&>::value,
            is_awaiter<Awaitable>,
            folly::Negation<detail::_has_free_operator_co_await<Awaitable>>,
            folly::Negation<detail::_has_member_operator_co_await<Awaitable>>>::
            value,
        int> = 0>
Awaitable& get_awaiter(Awaitable&& awaitable) {
  return static_cast<Awaitable&>(awaitable);
}
```

I added a type check on the template function to prevent it from taking an rvalue-reference object.

PR #2148, opened 2024-02-21
Fix compilation with nvcc

luhenry:


Orvid:
Which versions of NVCC has this been tested with?

luhenry:
@Orvid this has been tested with CUDA `V12.2.140`. This is the patchset that I've needed to be able to compile Velox on nvidia: https://github.com/rivosinc/velox/commit/05a892a21dbb3dd6335c9ab6b07626bac8e0eaca#diff-cd873badec87da0106ffc48d83c62dc7607bbd219ee8f31487a695aa304c6e46

cc @pedroerp

PR #2154, opened 2024-03-01
fix quoting in succeed.bat

ahornby:
fix quoting in succeed.bat

Summary:

Windows builds are failing due to a some missing quoting in the batch file generation from 66f05b0b861a7278693f3c9ee345a543159bc9d6

Add the necessary quotes.

Test Plan:

Run a build that loads vcvarsall:
```
python build/fbcode_builder/getdeps.py build double-conversion
```

Before, fails with
```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

After, works


ahornby:
Cc @Orvid

ahornby:
I see @jdelliot has done similar change in the meantime

Orvid:
And I see this right as I go to rebase it... Oh well, at least everything is working now :D

PR #2159, opened 2024-03-11
[DRAFT] Make IsRelocatable use the Clang builtin __is_trivially_relocatable(T)

Quuxplusone:
Clang 18's `__is_trivially_relocatable(T)` isn't yet compatible with Folly's use of the term; false negatives are fine, but Clang dangerously gives false positives. However, after https://github.com/llvm/llvm-project/pull/84621, the builtin will no longer give any (known) false positives, which means it will be safe for Folly to start using the builtin as of Clang 19+.

I'm posting this here hoping to kickstart a Folly discussion — "is Folly in favor of the Clang patch? would Folly indeed start using the Clang builtin, if it were long-term supported with the Folly(/Abseil/BSL/Qt/P1144/etc) semantics?"

***Please do not merge*** this patch until https://github.com/llvm/llvm-project/pull/84621 is successfully merged upstream; that patch is a prerequisite for this one.


Quuxplusone:
(attn @ot, @Gownta)

Quuxplusone:
Suggest closing this one in favor of #2216.

PR #2197, opened 2024-05-12
fix(cmake): prefer `folly-config.h` from build directory

mhx:
After 1866c47b0 and 45f3a6cc80, (some) cmake builds break because they'll pick up the newly committed `folly-config.h` from the source directory instead of the one created by cmake in the build directory.

This change fixes this by prioritizing the build directory include path.


Orvid:
We're going to end up fixing this in a different way.

PR #2199, opened 2024-05-13
Remove generated folly-config.h

vitaut:
This file is generated from CMake and the currently committed version breaks CI.


vitaut:
cc @yfeldblum

mhx:
@vitaut, see also #2197, which links to two commits, one of which has (some) context on why this was added.

Personally, I'd also prefer to see this file go away, but I don't know the full story behind why it was added.

vitaut:
The removal will be done by other means since this PR cannot be imported or merged directly.

PR #2203, opened 2024-05-15
fix: gcc uninitialized warnings triggered by unsafe_default_initialized

mhx:
Despite all the `FOLLY_*_DISABLE_WARNING`, the implementation of `unsafe_default_initialized_cv` triggers hundreds of `-Wuninitialized` and `-Wmaybe-uninitialized` warnings when building with GCC 12, 13, 14. As much as I've tried, it seems impossible to silence the warnings purely by diagnostic pragmas.

This change, however, eliminates all warnings, and the new code should otherwise have the exact same effect as before. At least, I can see that clang / gcc both produce identical code before and after this change when using `unsafe_default_initialized` to initialize variables.

### DIFF
--- a/folly/Utility.h
+++ b/folly/Utility.h
@@ -493,8 +493,8 @@ struct unsafe_default_initialized_cv {
 #if __cpp_lib_is_constant_evaluated >= 201811L
 #if !defined(__MSVC_RUNTIME_CHECKS)
     if (!std::is_constant_evaluated()) {
-      T uninit;
-      return uninit;
+      aligned_storage_for_t<T> uninit;
+      return reinterpret_cast<T&>(uninit);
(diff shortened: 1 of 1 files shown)


mhx:
ping

mhx:
Updated to a version that mostly resembles the existing `bit_cast` implementation in `folly/lang/Bits.h`.

yfeldblum:
e11e2c257cb0e47273495291fb411f81c01f34ce suppresses this path under gcc.

The approach in this PR could also work, but would have to be restricted to trivial types.

mhx:
Yes, you're correct about the restriction.

I don't mind which way this is ultimately fixed, so thanks for your fix!

Issue #2027, opened 2023-06-28
Checksum performance is slow on Arm64


kevinzs2048:
The checksum performance in folly is not optimized on Arm64 with Neon, which induce that the performance is quite slow.

./folly/hash/detail/ChecksumDetail.h

Cachelib heavily rely on Folly to realize the checksum.

From the perf top, in the cachelib with hyprid cache configuration, the checksum is consuming a lot of CPU time, which has been a bottleneck.

Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz
$ checksum_benchmark --bm_min_usec=10000
============================================================================
folly/hash/test/ChecksumBenchmark.cpp           relative  time/iter  iters/s
============================================================================
crc32_512                                                   55.73ns   17.94M
crc32_1024                                                  85.15ns   11.74M
crc32_2048                                                 116.29ns    8.60M
crc32_4096                                                 191.03ns    5.23M
crc32_8192                                                 341.44ns    2.93M
crc32_16384                                                627.76ns    1.59M
crc32_32768                                                  1.21us  827.16K
============================================================================
Comparison:

============================================================================
[...]folly/hash/test/ChecksumBenchmark.cpp     relative  time/iter   iters/s
============================================================================
crc32_512                                                   1.80us   554.82K
crc32_1024                                                  3.58us   279.35K
crc32_2048                                                  7.14us   140.13K
crc32_4096                                                 14.25us    70.18K
crc32_8192                                                 28.47us    35.12K
crc32_16384                                                56.93us    17.57K
crc32_32768                                               113.83us     8.79K

Orvid:
If checksum is the bottleneck, the first thing I'd recommend doing is shifting away from using crc32, which, even fully optimized on x86_64 is less than 1/4th the speed of hash algorithms designed for speed like XXH3. XXH3 in particular should be well optimized for AArch64.

It does appear that there are equivalent hardware instructions to do the CRC32 hashing on ARM, we just haven't implemented it yet since we haven't needed it.

PR #2204, opened 2024-05-16
Optimize crc32 & crc32c on NVIDIA Grace

krenzland:
This pull request adds hardware accelerated routines for CRC32 and CRC32C for Arm AARCH64 CPUs. The changes here have been tested on NVIDIA Grace.
In detail, it contains routines for:

- Computing CRC32 and CRC32C hashes on dataset using the CRC intrinsics. On Grace/Neoverse V2, this can process 8 bytes/cycle.
- A vectorized implementation of the `gf_multiply_crc32c_hw` and `gf_multiply_crc32_hw` functions used in routines to merge partial CRC checksums. These functions are more or less a 1:1 translation of the x86 vectorized routines. 
- I've introduced feature flags for AES, and SHA extensions for Arm CPUs. The feature checks for the vectorized functions are a bit more messy than on x86 because CPUs can implement a subset of these extensions.

This should resolve issue #2027.


folly/hash/detail/Crc32CombineDetail.cpp
```diff
@@ -105,7 +111,35 @@ static uint32_t gf_multiply_crc32_hw(uint64_t crc1, uint64_t crc2, uint32_t) {
   return _mm_cvtsi128_si32(_mm_srli_si128(_mm_xor_si128(res3, res1), 4));
 }
 
-#else
+#elif FOLLY_NEON && FOLLY_ARM_FEATURE_CRC32 && FOLLY_ARM_FEATURE_AES && FOLLY_ARM_FEATURE_SHA2
```
meteorfox, line 114: This line is too long.  
Can you resubmit after a clang-format?

Sorry, we should have this automated as part of a git hook or something, not why don't have it.

krenzland: Okay. Can I suggest to add clang-format to https://github.com/facebook/folly/blob/main/CONTRIBUTING.md?

krenzland:
Thanks for the review! I forgot to add that this should be compiled with the flags
` python3 build/fbcode_builder/getdeps.py --allow-system-packages build --extra-cmake-defines '{"CMAKE_CXX_FLAGS": "-march=armv8.5-a+crc+crypto"}'
`
or similar (+crypto could be replaced by +aes+sha2?) to enable all required features.

meteorfox:
@krenzland Hey after internal discussions, we would like to request to move your contributions under `folly/external/nvidia-crc32` to have a more defined copyright lines, you can still define them under `folly` namespace.

AFAIU, CMake should automatically pick them up, as we have auto_source with recurse.

Thanks in advanced

folly/hash/Checksum.cpp
```diff
@@ -86,7 +96,48 @@ bool crc32_hw_supported() {
   return id.sse42();
 }
 
-#else
+#elif FOLLY_ARM_FEATURE_CRC32
+uint32_t crc32_hw(const uint8_t* buf, size_t len, uint32_t crc) {
+  auto* buf_64 = reinterpret_cast<const uint64_t*>(buf);
```
yfeldblum, line 101: This looks like unaligned access.

Note that while an architecture may support unaligned access, the language generally deems unaligned access of this form to be undefined behavior. The kind of undefined behavior that may end up being subject to miscompilations.

To avoid this unaligned access, do we need to start off with an optional round of each of `__crc32b`, `__crc32h`, `__crc32w`, and `__crc32d`? Alternatively, perhaps we can `memcpy` from `buf` instead of `reinterpret_cast` - modern compilers recognize this idiom and lower to `mov` or `ldr` instructions without emitting calls to `memcpy`.

Here and `crc32c_hw` below.

krenzland: Thanks, you're totally right. Unaligned access isn't directly forbidden in C++ but my implementation is incorrect by the strict aliasing rule. While this is *typically fine* when operating on bytes my code is still incorrect. 

The memcpy version is a good idea. It's obviously correct and at a first glance seems to lead to (nearly) the same code for GCC/clang. I'll test it and update the PR.

PR #2211, opened 2024-05-22
Changed BOOST_PP_IIF to BOOST_PP_IF

aaditya2200:


Orvid:
This is fine as-is, we were just missing an `#include`, which has now been added.

PR #2237, opened 2024-06-15
Update build.sh

SnehaKumari2005:
removed the extra spaces in the comments and the main code and changed a typo i.e. "fi" to "if".


Orvid:
The spacing here is intentional, and the `fi` is correct for bash, and indicates the end of the `if` statement.

Issue #2167, opened 2024-03-22
off64_t undefined on alpine


benstadin:
https://github.com/facebook/folly/blob/faedbb468779075a022a9b97c70979a64bf13369/folly/portability/Unistd.h#L23

The missing macro definition results in a build failure on alpine (3.19) with the following error:

```
folly/system/MemoryMapping.h:102:5: error: 'off64_t' does not name a type; did you mean 'off_t'?
  102 |     off64_t pageSize = 0;
      |     ^~~~~~~
      |     off_t
```

Defining that macro fixes the issue. Though I do not have a generic solution as a proposal.

Orvid:
Is this on a 32 or 64-bit platform? It's possible we've just failed to include `<sys/types.h>`.

benstadin:
That's alpine 3.19 (docker) on arm64.

Alpine support is basically broken anyways on that platform. I get over a hundred of failed tests (mainly due to segfaults), even though stack size is set generously. From these tests probably the aligned malloc and thread tests are worth taking a look:

# The following tests FAILED:
# 	336 - atomic_shared_ptr_test.AtomicSharedPtr.DeterministicTest (SEGFAULT)
# 	354 - core_cached_shared_ptr_test.CoreCachedSharedPtr.AtomicCoreCachedSharedPtr (SEGFAULT)
# 	403 - dynamic_bounded_queue_test.DynamicBoundedQueue.enqDeq (SEGFAULT)
# 	420 - unbounded_queue_test.UnboundedQueue.enqDeq (SEGFAULT)
# 	474 - executor_test.ManualExecutor.getViaDoesNotDeadlock (SEGFAULT)
# 	484 - fiber_io_executor_test.FiberIOExecutorTest.event_base (SEGFAULT)
# 	487 - global_executor_test.GlobalExecutorTest.GlobalImmutableIOExecutor (SEGFAULT)
# 	489 - global_executor_test.GlobalExecutorTest.GlobalIOExecutor (SEGFAULT)
# 	501 - threaded_executor_test.ThreadedExecutorTest.exception (SEGFAULT)
# 	514 - priority_unbounded_blocking_queue_test.PriorityUnboundedBlockingQueueTest.concurrent_push_pop (SEGFAULT)
# 	520 - unbounded_blocking_queue_test.UnboundedBlockingQueue.concurrentPushPop (SEGFAULT)
[...]

PR #2248, opened 2024-07-02
Fix `using` directive in `Unistd.h` to fix react native compilation error

SagarSDagdu:
- When using React Native 0.74.x, RCT-Folly is installed as a `Pod` dependency.
- When compiling the app for iOS using `xcodebuild`, the file `Unistd.h` gives compilation error "Unexpected type name 'off_t': expected expression".

image

- Changing it to typedef makes this error go away, and the app compiles and runs correctly.

This might also fix: https://github.com/facebook/folly/issues/2167#issuecomment-2196249001

Note: CLA is signed ✅

### DIFF
--- a/folly/portability/Unistd.h
+++ b/folly/portability/Unistd.h
@@ -21,7 +21,7 @@
 #include <unistd.h>
 
 #if defined(__APPLE__) || defined(__EMSCRIPTEN__)
-using off64_t = off_t;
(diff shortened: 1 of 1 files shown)


Orvid:
Thanks for this!

SagarSDagdu:
@Orvid Any plans to merge this?

PR #2252, opened 2024-07-13
fix: compatible with lower clang version

YouSenRong:
* At least the Clang 11.0.0 can support "asm volatile inline", the lower version (e.g. 9.0.1) can only support "asm volatile".

### DIFF
--- a/folly/chrono/Hardware.h
+++ b/folly/chrono/Hardware.h
@@ -91,7 +91,11 @@ inline std::uint64_t hardware_timestamp_measurement_start() noexcept {
   return ret;
 #elif defined(__GNUC__) && FOLLY_X64
   uint64_t ret = 0;
-  asm volatile inline(
+  #if __clang_major__ >= 11
+    asm volatile inline(
+  #else 
+    asm volatile(
+  #endif
       "lfence\n"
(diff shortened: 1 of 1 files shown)


YouSenRong:
When I compile the compile the master of folly with Clang compiler (9.0.1), I met a problem as bellow. 
```
"/folly/folly/chrono/Hardware.h:94:16: error: expected '(' after 'asm'
  asm volatile inline("
```

And the problem can be reproduce on Compiler Explorer.
When compile with Clang 9.0.1, the problem can be reproduced.
image
After I removed the "inline" keyword, the problem was solved.
image

Besides, I found that, the problem can also be solved when compile with at least Clang 11.0.0, 
image

And I added the condition of __clang__major__ to make the code work.
```
  #if __clang_major__ >= 11
    asm volatile inline(
  #else 
    asm volatile(
  #endif
```

YouSenRong:
Hi, will this pr be merged, or is there any thing else should be done before being merged?

Orvid:
I think this will probably be able to be merged, though I've had to make some minor changes internally to make sure that GCC still has the right behavior. The tests are running on the new version now.

YouSenRong:
Ok, Thanks! @Orvid

PR #2257, opened 2024-07-15
Add missing unistd.h header for the close() function

yurivict:


Orvid:
This is already in progress in https://github.com/facebook/folly/pull/2245

It's taken a bit to actually get tested and working internally due to this needing to go through the portability header instead (which I didn't think it needed to, so it took multiple round trips to actually get in the right state). Hopefully I've finally got it in a good state, which should mean it can land either later tonight, or in the next couple of days.

Orvid:
https://github.com/facebook/folly/commit/af86faad1fe31866d749aa3e1c64a97662f44ae2 has landed to fix this.

PR #2265, opened 2024-07-24
fix: un-qualify Range::value_type for compatibility with fmt v11

mhx:
`folly::StringPiece` no longer works with fmt v11 due to `value_type` being `const char` instead of the `char` expected by fmt.

Looking at `std::span`, which is probably the closest equivalent to `folly::Range`, `value_type` is defined as `std::remove_cv_t<T>`, so I think it would make sense for `folly::Range` to define `value_type` similarly, i.e. without cv-qualification.

This change removes cv-qualifiers from `value_type` and replaces the the use of `value_type&` by `reference` and `const value_type&` by the newly introduced `const_reference`.

See https://github.com/fmtlib/fmt/pull/4086 for additional context.

### DIFF
--- a/folly/Range.h
+++ b/folly/Range.h
@@ -104,8 +104,7 @@ inline size_t qfind(
  */
 template <class Iter>
 size_t qfind(
-    const Range<Iter>& haystack,
-    const typename Range<Iter>::value_type& needle);
+    const Range<Iter>& haystack, typename Range<Iter>::const_reference needle);
 
 /**
  * Finds the last occurrence of needle in haystack. The result is the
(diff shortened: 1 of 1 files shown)


folly/Range.h
```diff
@@ -247,10 +245,11 @@ class Range {
   using size_type = std::size_t;
   using iterator = Iter;
   using const_iterator = Iter;
-  using value_type = typename std::remove_reference<
-      typename std::iterator_traits<Iter>::reference>::type;
-  using difference_type = typename std::iterator_traits<Iter>::difference_type;
   using reference = typename std::iterator_traits<Iter>::reference;
+  using const_reference = typename std::add_const<reference>::type;
```
yfeldblum, line 249: For new changes let’s use the C++17 style traits.

mhx: More than happy to change this, I was just sticking to current style of the code.

michel-slm:
Thanks! I'll add this to the Fedora build while you work with @yfeldblum to get this merged upstream

vitaut:
Superseded by https://github.com/facebook/folly/commit/21e8dcd464ee46b2144a1e4d4c0e452355ae15f0 but thanks for working on this!

PR #2268, opened 2024-07-31
Add fbcode_builder manifests for mcrouter and ragel

mszabo-wikia:
https://github.com/facebook/mcrouter currently uses a legacy autotools-based build system, which forces third parties to rely on a collection of scripts hosted in the repository to build it with all required dependencies. This is brittle and has lead to many issue reports about build problems.

In https://github.com/facebook/mcrouter/pull/449, I've prepared a working CMake-based build system for mcrouter that could replace the legacy autotools setup. This patch adds the necessary manifests for mcrouter and its ragel dependency so that fbcode_builder itself will be setup for the repo without having to do it in the PR.

### DIFF
--- a/build/fbcode_builder/manifests/mcrouter
+++ b/build/fbcode_builder/manifests/mcrouter
@@ -1,0 +1,23 @@
+[manifest]
+name = mcrouter
+
+[git]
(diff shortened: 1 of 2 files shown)


mszabo-wikia:
Thanks!

PR #2291, opened 2024-09-11
Use sysfs if parsing /proc/cpuinfo doesn't work

krenzland:
On Arm CPUs, such as NVIDIA Grace, /proc/cpuinfo may not contain the physical or core id, which is used to approximate the cache hierarchy. This PR implements a simple work-around: If parsing /proc/cpuinfo does not work, parse /sys/devices instead.


krenzland:
If necessary, an optimization of that could be to ifdef  this and always directly parse the sysfs on Arm systems - there, /proc/cpuinfo will essentially always fail.

Issue #2250, opened 2024-07-08
GenPkgConfig.cmake: file(GENERATE) failed with $<COMPILE_LANGUAGE> in fmt 11.0.0 MSVC


moyoumiu:
fmt [11.0.0](https://github.com/fmtlib/fmt/releases/tag/11.0.0) added the CMake generator expression `<$<COMPILE_LANGUAGE:CXX>:/utf-8>` for `MSVC` in <https://github.com/fmtlib/fmt/pull/3995>, which will cause folly to fail when generating pkgconfig file in [CMakeLists.txt#L497-L502](https://github.com/facebook/folly/blob/v2024.07.01.00/CMakeLists.txt#L497-L502).

### Problem

For fmt, it added `$<$<COMPILE_LANGUAGE:CXX>:/utf-8>` in [11.0.0/CMakeLists.txt#L358-L362](https://github.com/fmtlib/fmt/blob/11.0.0/CMakeLists.txt#L358-L362) for `MSVC AND FMT_UNICODE`.

If folly upgrade [dependency fmt](https://github.com/facebook/folly/blob/v2024.07.01.00/build/fbcode_builder/manifests/fmt) to 11.0.0, will generate `libfolly.pc.gen` with `Cflags: -I${includedir} $<$<COMPILE_LANGUAGE:CXX>:/utf-8>` in [CMakeLists.txt#L497-L502](https://github.com/facebook/folly/blob/v2024.07.01.00/CMakeLists.txt#L497-L502).

But `file(GENERATE)` context does not support `$<COMPILE_LANGUAGE:...>`, will failed with `Evaluation file to be written multiple times with different content`. See [CMake #26110](https://gitlab.kitware.com/cmake/cmake/-/issues/26110#:~:text=Since%20file(GENERATE)%20isn%27t%20compilation%2C%20that%20context%20doesn%27t%20support%20%24%3CCOMPILE_LANGUAGE%3A...%3E.).

Maybe it could fixed by add conditional processing like <https://github.com/facebook/folly/pull/1433>.

PR #2293, opened 2024-09-16
fix: fmt workaround for pkg-config generation

aminya:
Workaround for #2250.

Ideally, the code should support the generator expression, but this workaround fixes the problem for fmt.

Also, see https://github.com/fmtlib/fmt/pull/4159 and https://github.com/microsoft/vcpkg/pull/40944


aminya:
CLA signed @facebook-github-bot

PR #2310, opened 2024-10-10
build: make dependency on fast_float optional in CMake build

mhx:
The fast_float library isn't widely available, especially for older distributions. Furthermore, `FOLLY_CONV_ATOD_MODE` isn't even set by the CMake build, so conversion doesn't actually use fast_float.


mhx:
Ping

mhx:
I can update the PR if needed, just let me know.

PR #2324, opened 2024-10-22
Fix incorrect include

johnweldon:
Was breaking after upgrading compiler

Discovered in the brew recipe. The error was:

```
.../folly/concurrency/SingletonRelaxedCounter.h:71:12: error: no template named 'unordered_map' in namespace 'std'; did you mean 'unordered_set'?
      std::unordered_map<LocalLifetime*, CounterAndCache*> lifetimes;
      ~~~~~^~~~~~~~~~~~~
           unordered_set
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/unordered_set:411:28: note: 'unordered_set' declared here
class _LIBCPP_TEMPLATE_VIS unordered_set
                           ^
1 error generated.
make[2]: *** [CMakeFiles/folly_base.dir/folly/settings/Settings.cpp.o] Error 1
make[1]: *** [CMakeFiles/folly_base.dir/all] Error 2
make: *** [all] Error 2
```

PR #2331, opened 2024-10-29
Added unordered_map header to file to resolve bug

greauxd:
Added header to resolve build failure issue


Orvid:
This has the same effect as https://github.com/facebook/folly/pull/2324, which is already in the process of being merged.

PR #2338, opened 2024-11-12
Undefine clearenv() on FreeBSD

yurivict:


yurivict:
@Orvid 

Hi Orvid,

There is another problem is that ```#include <stdlib.h>``` is missing in system/EnvUtil.cpp for ```clearenv()```.

[log](https://pkg-status.freebsd.org/beefy16/data/133amd64-default/3c22ee0df4c5/logs/folly-2024.11.11.00.log)

Could you please add it?

Orvid:
`system/EnvUtil.cpp` already [includes](https://github.com/facebook/folly/blob/main/folly/system/EnvUtil.cpp#L22) `folly/portability/Stdlib.h`, which [includes](https://github.com/facebook/folly/blob/main/folly/portability/Stdlib.h#L19) `cstdlib`. Our portability headers that are named after standard library headers are explicitly defined to always include the standard library header it's named after, if it's available on the current platform.

Does FreeBSD not expose `clearenv` via `cstdlib` for some reason?

PR #2394, opened 2025-03-06
Add inlining hints to concurrent hash map

krenzland:
This PR adds inlining hints to several functions of the ConcurrentHashMap. With clang17 and NVIDIA Grace, this speeds up the benchmarks for begin() by ~35-40% and find() by ~15-20%. The iterate benchmarks show inconclusive results and may have a slowdown or speedup depending on input size and number of threads. Overall the effect is positive.

This PR depends on #2393, otherwise the additional inlining hints will allow the compiler to remove parts of the benchmark.

### DIFF
--- a/folly/concurrency/ConcurrentHashMap.h
+++ b/folly/concurrency/ConcurrentHashMap.h
@@ -288,10 +288,12 @@ class ConcurrentHashMap {
     return true;
   }
 
-  ConstIterator find(const KeyType& k) const { return findImpl(k); }
+  FOLLY_ALWAYS_INLINE ConstIterator find(const KeyType& k) const {
+    return findImpl(k);
+  }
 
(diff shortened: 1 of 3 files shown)


YifanYuan3:
hey @krenzland, we have some internal discussion about this. Force inlining makes benchmarks faster, but makes binaries bigger and therefore slower in the wild. Some of the force inlining, for functions that are simple forwarding functions, makes sense, but we imagine that those are already inlined by the compiler. 

We will have to find some real prod services to test and evaluate it, where concurrent hash map pop up as a hot function/bottleneck. Let's keep this PR pending and we will let you know when we figure out the next step. Thanks!

krenzland:
Makes sense. The functions that are force inlined here should be pretty minimal. I would guess that you'd see some real-world performance gains mostly from the find function.

PR #2405, opened 2025-03-23
Subprocess: fix constructor for macOS

barracuda156:
See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119435

P. S. May also be relevant for Solaris: https://github.com/OpenMathLib/OpenBLAS/issues/700
But I have no way to check that.


yfeldblum:
What versions of macOS is this failing on, and what compilers?

PR #2414, opened 2025-04-14
Fix va_list reuse and null dereference in getenv/setenv on Windows

INitu2002:
This PR fixes two small but important issues I found while reading and testing the code in `folly`. I'm a student learning more about programming and contributing to open-source projects, and I noticed two places where the code could cause real issues due to undefined or unsafe behavior.

1. Fix: Reusing `va_list` without `va_copy` in `dprintf()`
In `Stdio.cpp`, the function `dprintf` calls `vsnprintf` twice using the same `va_list` (`args`). According to the C++ standard, once `va_list` is passed to a function like `vsnprintf`, it is considered consumed and cannot be reused safely.
This could cause unpredictable behavior, incorrect formatting, or even a crash.
I fixed this by adding a `va_copy` before the first usage, and properly calling `va_end` on the copy.

2. Fix: Potential null pointer dereference when using `getenv`
In `Stdlib.cpp`, there was a line that was directly dereferencing the result of `getenv(name)`: *getenv(name) = '\0';
This assumes that getenv(name) always returns a valid pointer. However, if the environment variable does not exist, getenv returns nullptr, and this line could cause a segmentation fault.
I added a check to ensure the return value of getenv(name) is not null before attempting to dereference it. I also applied the same fix to the _wgetenv section that handles wide environment variables.

Please let me know if these fixes are correct or if there’s anything that needs to be improved.

### DIFF
--- a/folly/portability/Stdlib.cpp
+++ b/folly/portability/Stdlib.cpp
@@ -102,12 +102,19 @@ int setenv(const char* name, const char* value, int overwrite) {
 
   // Here lies the documentation we blatently ignore to make
   // this work >_>...
-  *getenv(name) = '\0';
+
+  char* env = getenv(name);
+  if (!env) {
+    errno = EINVAL;
+    return -1;
+  }
+  *env = '\0';
+  *(env + 1) = '=';
+
   // This would result in a double null termination, which
   // normally signifies the end of the environment variable
   // list, so we stick a completely empty environment variable
   // into the list instead.
-  *(getenv(name) + 1) = '=';
 
   // If _wenviron is null, the wide environment has not been initialized
   // yet, and we don't need to try to update it.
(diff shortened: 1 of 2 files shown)


folly/portability/Stdlib.cpp
```diff
@@ -102,12 +102,19 @@ int setenv(const char* name, const char* value, int overwrite) {
...
   // this work >_>...
-  *getenv(name) = '\0';
+
+  char* env = getenv(name);
+  if (!env) {
+    errno = EINVAL;
+    return -1;
+  }
+  *env = '\0';
+  *(env + 1) = '=';
```
Orvid, line 112: This needs to be back below this next comment on line 114.

INitu2002:
Thanks a lot for reviewing and importing! Really happy to contribute!

folly/portability/Stdlib.cpp
Orvid: I'm also not entirely convinced the change here is required, since we already call `_putenv_s` for this environment variable up on line 98, so the variable should be guaranteed to exist.

Orvid:
As a heads up, internally I've dropped the change to `getenv/setenv` and will just land the change to stdlib, since the changes to `getenv/setenv` doesn't actually check anything, since we set that specific environment variable immediately prior in that function. The commit that lands will have the commit summary and title updated to reflect that.

If you think I'm missing something about the `getenv/setenv`, I can bring those changes back in a separate commit.

INitu2002:
You are right, since the environment variable is set right before the getenv() call, it should always be valid. My check was added out of caution, but I understand that in this context it's not needed. Totally fine with dropping that part. I appreciate the explanation and the merge of the other fix!

PR #2443, opened 2025-05-27
Switch to `sb` instruction at runtime for ARM v8.5 onward

salvatoredipietro:
For portability of the library, it is beneficial to implement the switch between `isb` and `sb` at runtime rather than at build time. Here an implementation of it which uses `HWCAP_SB` to identify the capability and switch to the appropriate instruction at runtime.

### DIFF
--- a/folly/portability/Asm.h
+++ b/folly/portability/Asm.h
@@ -24,6 +24,14 @@
 #include <intrin.h>
 #endif
 
+#if defined(__linux__) && FOLLY_AARCH64
+#include <sys/auxv.h>
+#ifndef HWCAP_SB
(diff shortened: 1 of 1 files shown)


salvatoredipietro:
Can anyone take a look to this and provide some initial feedback, please?

salvatoredipietro:
any feedback on this?

Orvid:
We really don't want this to be dispatched at runtime, it just doesn't make sense for a single instruction like this. If there's a noticeable benefit from using `sb` rather than `isb`, then we would want this dispatch between instructions to be done at a higher level than directly within `asm_volatile_pause`, since `asm_volatile_pause` is almost always called in a loop.

PR #2449, opened 2025-06-06
Remove use of unsupported std::shared_ptr::unique()

mszabo-wikia:
libc++ 18 and newer gate this behind the
`_LIBCPP_ENABLE_CXX20_REMOVED_SHARED_PTR_UNIQUE` macro, since the feature was deprecated in C++17 and removed in C++20.

It's probably better to replace it with use_count() rather than using the macro.


mszabo-wikia:
Thanks!

PR #2484, opened 2025-08-06
feat: enforce little-endianness for `folly::Bits<>` storage

mhx:
The inline docs for `folly::Bits` already mention that storage is expected to be little-endian. This change enforces little-endianness by using `folly::Endian` for load/store operations. This should be a no-op on little-endian systems.

The main motivation for this change is that it immediately enables fbthrift Frozen2 files to be exchanged between little- and big-endian systems.


mhx:
Updated to simplify the change and leave the `BitsTraits` untouched. Also fixed the unit tests on big-endian systems, which were previously broken.

PR #2499, opened 2025-09-10
fix(folly): Replace __type_pack_element in ForeachTest.cpp for GCC 14 Compatibility

ihb2032:
This pull request addresses a build failure in folly/container/test/ForeachTest.cpp when compiling with GCC 14. The error is due to the use of the compiler builtin __type_pack_element in function signatures, which is no longer permitted.
The fix replaces the problematic use of type_pack_element_t with the standard library equivalent, std::tuple_element_t. This change ensures that the test suite remains compatible with the latest compiler versions without altering the logic of the tests.


yfeldblum:
Thanks! Separately fixed by: 9d729ef9a6b06dfe3b77e55262f22db1d1abd476.

PR #2503, opened 2025-09-17
fix(oss): explicitly enable coroutine support on g++

ben--:
g++ doesn't enable coroutines by default and some upcoming changes require them. This change adds the necessary compiler flag on the open source build. The code here is brought over from fbthrift:

https://github.com/facebook/fbthrift/blob/main/CMakeLists.txt#L71-L75

Tested by including folly in a sapling build with PYTHON_EXTENSIONS enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,12 @@ if(NOT CMAKE_CXX_STANDARD)
   message(STATUS "setting C++ standard to C++${CMAKE_CXX_STANDARD}")
 endif()
 
+# Explicitly enable coroutine support, since GCC does not enable it
+# by default when targeting C++17/20.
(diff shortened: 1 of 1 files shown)


mszabo-wikia:
o/ I dug up the conversation where I landed this change in fbthrift (https://github.com/facebook/fbthrift/pull/615) and it seems this flag might not be needed on GCC >=11 if c++20 is required (which is now the folly default). Sapling still seems to default to c++17, I wonder if that may be the issue?

ben--:
@mszabo-wikia is correct -- this change is not required if building with `-std=c++20` on GCC version 11 or above.

However, if this change is dropped, someone may want to update README.md which currently reads:

> folly supports gcc (5.1+), clang, or MSVC

I'm not arguing one way or another here, but folly needs to decide what its backwards-compatibility strategy ought to be.

PR #2509, opened 2025-09-30
Allow overwriting module paths for HHVM OSS

mszabo-wikia:
HHVM OSS[1] mirrors its Meta first-party library dependencies via shipit and consumes them via `ExternalProject` so that it can be built as a single CMake project. It also vendors a number of third-party library dependencies such as gflags and glog that are also dependencies of these first-party library dependencies. It currently tries to ensure that the glog include directories and link libraries used when compiling and linking folly point to the vendored glog by setting `GLOG_INCLUDEDIR` / `GLOG_LIBRARY` which are forwarded to `find_library` and co. by FindGlog.cmake, but this doesn't work because `find_library` and co. only consider these hint paths after system default paths have been searched, so the system glog (if any) would end up being preferred.

So, explicitly opt into CMP0074[2] to allow HHVM to simply set `Glog_ROOT` to override the glog search paths. Also ensure that folly-deps.cmake puts gflags after glog on the link line: while this isn't an issue in the common case that both are consumed as shared libraries, the current order will fail if they are consumed as static libraries as glog depends on gflags and must therefore come before gflags on the link line.

---
[1] https://github.com/facebook/hhvm
[2] https://cmake.org/cmake/help/latest/policy/CMP0074.html


mszabo-wikia:
Closing this since we can work around the issue via CMAKE_PREFIX_PATH manipulation instead.

PR #2512, opened 2025-10-06
Neon xoshiro256pp uplift

kielfriedt:
arm changes in folly that provide a uplift in xoshiro256. Optimized NEON version code using VecResCount = 4 to retain all data in registers, eliminating spills and reducing mov instructions to avoid the high register pressure
Achieved 15.76% uplift on xoshiro256_64 case and 8.64% uplift on xoshiro256 case

| Grace-Clang19 | Original Code | Optimized code (Neon+VecResCount=4) | Uplift |
| ------------- | ------------- | ----------------------------------- | ------ |
| xoshiro256    | 684.66M       | 743.83M                             |  8.64% |
| xoshiro256_64 | 529.31M       | 612.75M                             | 15.76% |

**NOTE, when build with clang, we need manually tell it to use eor3 by specifying "--extra-cmake-defines '{"CMAKE_C_FLAGS": "-O3 -march=native+sha3", "CMAKE_CXX_FLAGS": "-O3 -march=native+sha3"}'" to getdeps.py.**


folly/random/xoshiro256pp.h
```diff
@@ -138,7 +138,109 @@ class xoshiro256pp {
...
+
+      curState3[0] = s3_0;
+      curState3[1] = s3_1;
+      curState3[2] = s3_2;
+      curState3[3] = s3_3;
+
+    }
+    cur = 0;
+#else
+   for (uint64_t i = 0; i < VecResCount; i++) {
```
mcfi, line 243: What code gen do you get if you add #pragma clang loop unroll_count(4) to unroll this loop 4 times? Is it still worse than the manually unrolled loop above?

mcfi: Would it make sense to set the loop unroll count to the number of SIMD units?

folly/random/xoshiro256pp.h
```diff
@@ -138,7 +138,109 @@ class xoshiro256pp {
...
+      uint64x2_t s2_2 = curState2[2];
+      uint64x2_t s2_3 = curState2[3];
+
+      uint64x2_t s3_0 = curState3[0];
+      uint64x2_t s3_1 = curState3[1];
+      uint64x2_t s3_2 = curState3[2];
+      uint64x2_t s3_3 = curState3[3];
+
+      uint64x2_t sum0 = vaddq_u64(s0_0, s0_3);
+      uint64x2_t rot0 = vorrq_u64(vshlq_n_u64(sum0, 23),vshrq_n_u64(sum0, 64 - 23));
```
Nicoshev, line 172: @kielfriedt  What about using instruction XAR for the rotate. It would be a XAR with a vdupq_n_u64(0) and rotating by 41

kielfriedt:
@Nicoshev are we waiting on anymore feedback or can this be closed?

PR #2519, opened 2025-10-10
Fix Subprocess.cpp with clang + _FORTIFY_SOURCE on glibc >= 2.40

mszabo-wikia:
On glibc >= 2.40, _FORTIFY_SOURCE overloads openat() when building with clang, causing the `decltype` call in Subprocess.cpp to fail since the target becomes ambiguous. So, explicitly specify our overload of choice in this case.

Reproduce/test by setting `CMAKE_CXX_FLAGS=-D_FORTIFY_SOURCE=3 -DCMAKE_CXX_COMPILER=clang++`.

### DIFF
--- a/folly/Subprocess.cpp
+++ b/folly/Subprocess.cpp
@@ -145,6 +145,11 @@ static Ret subprocess_libc_load(
 #define FOLLY_DETAIL_SUBPROCESS_LIBC_X_OPEN(X) \
   X(open, __open_real)                         \
   X(openat, __openat_real)
+#elif defined(_FORTIFY_SOURCE) && defined(__clang__) && (__GLIBC__ > 2 || __GLIBC__ == 2 && __GLIBC_MINOR__ >= 40)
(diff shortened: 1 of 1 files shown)


Wilfred:
FWIW we're seeing build failures here:

```
error: no member named '__openat_2' in the global namespace
```

and openat_2 has existed since at least glibc 2.7: https://github.com/bminor/glibc/blob/glibc-2.7/include/fcntl.h#L29

Are you sure this is correct?

Wilfred (changes requested):
See my comment on the PR.

mszabo-wikia:
So this was aiming to workaround the glibc 2.40+ changes introduced in https://github.com/bminor/glibc/commit/86889e22db329abac618c6a41f86c84657a15324 that only manifest with `FORTIFY_SOURCE` defined.  Not sure what might be causing those failures, I'll look into it!

chenx97:
try `__openat_alias`

PR #2524, opened 2025-10-22
Fix asan check in UninitializedMemoryHacks.h

pps83:
libcxx removed `_LIBCPP_HAS_ASAN`/`_LIBCPP_HAS_NO_ASAN` macro in favor simple feature test for sanitizer. See: https://github.com/llvm/llvm-project/commit/040e9e02ccbf69bebd293a4c389948d7ecbc7bd9


Spikhalskiy:
This problem seems to be addressed in https://github.com/facebook/folly/commit/fdde9bc360d525a1b2889b9ba89d671c3a13e72e

pps83:
> This problem seems to be addressed in [fdde9bc](https://github.com/facebook/folly/commit/fdde9bc360d525a1b2889b9ba89d671c3a13e72e)

Yep, it's a better fix, backward compatible with older libcxx.
There are still [two more fixes](https://github.com/facebook/folly/pulls/pps83) pending. msvc related is most likely still relevant

PR #2547, opened 2025-12-25
[Build] Fix ABI ODR violation for C++20 consumers linking against C++…

### LINKED ISSUE #2477: Double free error when mixing c++20 project with folly built using c++17
If folly is built with c++17, but the project using folly is built using c++20, when jemalloc is enabled, `folly::usingJEMalloc()` has double free error at [folly/memory/Malloc.h#L192](https://github.com/facebook/folly/blob/470fa8d7c9772af2833d25f4c2bafa0726f5278a/folly/memory/Malloc.h#L192).

Compiler is clang 18.

Minimal code to reproduce:

```
int main(int argc, char **argv) {
    const folly::Init init(&argc, &argv);
    std::cout << "use jemalloc: " << folly::usingJEMalloc() << std::endl;
    return 0;
}
```
`folly::usingJEMalloc()` should return true but actually it returns false.

This is because template class `FastStaticBool<Initializer>` is initialized twice using different definitions under c++17 and c++20 (this violates ODR and causes UB). As a result, `Initializer::operator()()` is called twice. For reasons I cannot understand, `ptr` is declared as static at [folly/memory/Malloc.h#L186](https://github.com/facebook/folly/blob/470fa8d7c9772af2833d25f4c2bafa0726f5278a/folly/memory/Malloc.h#L186). It's allocated once but freed twice. 

This section of code is excluded if `FOLLY_SANITIZE` defined, which makes it escaped from sanitizer checks.

Apart from that strange `static` declaration on `ptr`, I think we should avoid the template being defined differently. A possible fix is to move `FOLLY_CPLUSPLUS` macro from `Portability.h` to `folly-config.h`, and set its value to the C++ version used when compiling folly.

divitsinghall:
# Fix ABI ODR violation when linking C++20 consumer against C++17 Folly

## Summary

Fixes #2477 - Memory corruption / double-free crashes when a C++20 application links against a Folly library compiled with C++17.

## Root Cause

The `[[no_unique_address]]` C++20 attribute changes struct memory layout by allowing empty class members to occupy zero bytes. Folly uses this attribute via the `FOLLY_ATTR_NO_UNIQUE_ADDRESS` macro in several internal structures.

**The Problem:** When Folly is compiled as C++17, the attribute is not applied, and empty members occupy ≥1 byte. When the same headers are included from a C++20 application, the attribute IS applied, creating a smaller layout. This ODR (One Definition Rule) violation causes memory corruption at runtime.

### Affected Structures

| File | Class | Member |
|------|-------|--------|
| `folly/container/span.h:126` | `fallback_span::span` | `extent_` |
| `folly/detail/tuple.h:126` | `lite_tuple::entry` | `entry_value` |
| `folly/executors/SerialExecutor-inl.h:265` | `SerialExecutorMPSCQueue` | `mutex_` |
| `folly/lang/Exception.cpp:316` | `scope_guard_` | `func_` |
| `folly/coro/AutoCleanup.h:105` | `AutoCleanup` | `scheduled_` |
| `folly/python/ProactorExecutor.h:61` | (internal) | `overlapped_` |

### Layout Difference Example

```cpp
struct Empty {};

struct WithAttr {
    char* ptr;
    [[no_unique_address]] Empty e;  // C++20: 0 bytes
};

struct WithoutAttr {
    char* ptr;
    Empty e;  // C++17: 1+ byte (padding to 8)
};

// C++17: sizeof(WithoutAttr) == 16 (8 + 1 + 7 padding)
// C++20: sizeof(WithAttr) == 8 (Empty optimized away)
```

### Compiler Flags

```bash
# Build Folly library with C++17
cmake -DCMAKE_CXX_STANDARD=17 ..
make folly

# Build consumer application with C++20
clang++ -std=c++20 app.cpp -lfolly  # CRASH: double-free or corruption
```

### Minimal Reproduction

```cpp
// app.cpp - compiled with -std=c++20
#include <folly/futures/Future.h>

int main() {
    // Internal structures have mismatched layouts between
    // the C++17 library and C++20 header interpretation
    folly::makeFuture(42).thenValue([](int x) {
        std::cout << x << std::endl;
    });
    return 0;  // CRASH on destruction
}
```

### 1. Record Library ABI at Build Time

**`CMake/FollyConfigChecks.cmake`** - Detect if `[[no_unique_address]]` works:

```cmake
check_cxx_source_compiles("
  struct Empty {};
  struct Test {
    [[no_unique_address]] Empty e;
    int x;
  };
  static_assert(sizeof(Test) == sizeof(int), \"\");
  int main() { return 0; }"
  FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS
)
```

**`CMake/folly-config.h.cmake`** - Export to installed headers:

```c
#define FOLLY_LIBRARY_CXX_STANDARD @CMAKE_CXX_STANDARD@
#cmakedefine01 FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS
```

### 2. Force ABI Compatibility in Headers

**`folly/CppAttributes.h`** - Conditional attribute application:

```cpp
#if defined(FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS)
  #if !FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS && FOLLY_CURRENT_HAS_NO_UNIQUE_ADDRESS
    // Library=C++17, App=C++20: DISABLE attribute to match library layout
    #pragma message("Warning: Folly compiled without [[no_unique_address]]...")
    #define FOLLY_ATTR_NO_UNIQUE_ADDRESS /* disabled for ABI compatibility */
  #elif FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS && !FOLLY_CURRENT_HAS_NO_UNIQUE_ADDRESS
    // Library=C++20, App=C++17: ERROR - impossible to fix
    #error "Folly requires C++20 or later to match library ABI"
  #else
    // Matching versions: use appropriate setting
    #define FOLLY_ATTR_NO_UNIQUE_ADDRESS ...
  #endif
#endif
```

### Before Fix (ODR Violation)

```
$ clang++ -std=c++17 -c lib.cpp     # sizeof(span<int>) = 16
$ clang++ -std=c++20 -c app.cpp     # sizeof(span<int>) = 8  ← MISMATCH
$ clang++ lib.o app.o -o app && ./app
*** Error: double free detected ***
```

### After Fix (ABI Compatible)

```
$ clang++ -std=c++17 -c lib.cpp     # sizeof(span<int>) = 16
                                    # FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS = 0
$ clang++ -std=c++20 -c app.cpp     # sizeof(span<int>) = 16  ← MATCHES
                                    # Attribute disabled due to library config
In file included from app.cpp:
folly/CppAttributes.h:171:9: warning: Folly library was compiled without
[[no_unique_address]] (likely C++17)... [-W#pragma-messages]

$ clang++ lib.o app.o -o app && ./app
Value: 42
If you see this without a crash, ABI is compatible!
```

### Verification Test

```cpp
#include <folly/container/span.h>
#include <iostream>

int main() {
    std::cout << "sizeof(folly::span<int>) = " 
              << sizeof(folly::span<int>) << std::endl;
    
    // With fix: This will be consistent regardless of -std= flag
    // Without fix: 16 bytes with C++17, 8 bytes with C++20
    
    #if defined(FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS)
    std::cout << "FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS = "
              << FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS << std::endl;
    #endif
    
    return 0;
}
```

## Behavioral Changes

| Library | Application | Before Fix | After Fix |
|---------|-------------|------------|-----------|
| C++17 | C++20 | 💥 Crash | ⚠️ Warning + works |
| C++20 | C++17 | 💥 Crash | ❌ Compile error |
| C++17 | C++17 | ✅ Works | ✅ Works |
| C++20 | C++20 | ✅ Works | ✅ Works |

## New Macros

| Macro | Purpose |
|-------|---------|
| `FOLLY_LIBRARY_CXX_STANDARD` | C++ standard version used to build the library |
| `FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS` | Whether `[[no_unique_address]]` was active at library build |
| `FOLLY_SKIP_ABI_CHECK` | Define to suppress ABI mismatch warnings |

## Testing

- [x] Verified layout consistency with `sizeof()` checks
- [x] Tested C++17 library + C++20 app (warning issued, no crash)
- [x] Tested C++20 library + C++20 app (attribute applied, optimal)
- [x] Tested C++17 library + C++17 app (attribute disabled, consistent)
- [x] Added reproduction script: `repro_issue_2477/reproduce_abi_issue.sh`

## Files Changed

- `CMake/folly-config.h.cmake` - Export library ABI info
- `CMake/FollyConfigChecks.cmake` - Detect `[[no_unique_address]]` support
- `folly/CppAttributes.h` - ABI-aware attribute macro
- `folly/Portability.h` - C++ version mismatch warning
- `repro_issue_2477/repro.cpp` - Documented reproduction case
- `repro_issue_2477/reproduce_abi_issue.sh` - Automated reproduction script

## Related Issues

- Fixes #2477

## Notes for Reviewers

1. This fix prioritizes **ABI stability over performance** when there's a version mismatch. The `[[no_unique_address]]` optimization is disabled to match the pre-compiled library.

2. Users who want optimal performance should compile Folly with the same C++ standard as their application.

3. The warning can be suppressed with `-DFOLLY_SKIP_ABI_CHECK` for users who understand the risks.

### DIFF
--- a/CMake/FollyConfigChecks.cmake
+++ b/CMake/FollyConfigChecks.cmake
@@ -164,6 +164,22 @@ check_cxx_source_compiles("
   FOLLY_HAVE_EXTRANDOM_SFMT19937
 )
 
+# Check if [[no_unique_address]] attribute is available at library compile time.
+# This is critical for ABI stability: the attribute changes struct layout,
+# and if headers are compiled with a different C++ standard than the library,
+# ODR violations and memory corruption can occur.
+# See: https://github.com/facebook/folly/issues/2477
+check_cxx_source_compiles("
+  struct Empty {};
+  struct Test {
+    [[no_unique_address]] Empty e;
+    int x;
+  };
+  static_assert(sizeof(Test) == sizeof(int), \"\");
+  int main() { return 0; }"
+  FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS
+)
+
 check_cxx_source_runs("
   #include <stdarg.h>
   #include <stdio.h>
--- a/CMake/folly-config.h.cmake
+++ b/CMake/folly-config.h.cmake
@@ -85,3 +85,14 @@
 #cmakedefine01 FOLLY_HAVE_LIBRT
 
 #cmakedefine01 FOLLY_HAVE_VSOCK
+
+// Record the C++ standard version used to compile the Folly library.
+// This enables ABI compatibility checks when headers are included by
+// applications compiled with a different C++ standard version.
+// See: https://github.com/facebook/folly/issues/2477
+#define FOLLY_LIBRARY_CXX_STANDARD @CMAKE_CXX_STANDARD@
+
+// Record whether [[no_unique_address]] was active when building the library.
+// This attribute changes struct layout and can cause ODR violations if
+// headers are compiled with a different C++ standard than the library.
+#cmakedefine01 FOLLY_LIBRARY_HAS_NO_UNIQUE_ADDRESS
--- a/folly/CppAttributes.h
+++ b/folly/CppAttributes.h
@@ -16,11 +16,22 @@
 
 /**
  * GCC compatible wrappers around clang attributes.
+ *
+ * ABI STABILITY NOTE:
+ * Some attributes (notably [[no_unique_address]]) affect struct layout.
+ * When the Folly library is compiled with one C++ standard and headers are
+ * included from an application compiled with a different standard, ODR
+ * violations can occur, leading to memory corruption and crashes.
+ *
+ * This header includes logic to detect such mismatches and force the headers
+ * to use the same layout as the compiled library.
+ * See: https://github.com/facebook/folly/issues/2477
  */
 
 #pragma once
 
 #include <folly/Portability.h>
+#include <folly/portability/Config.h>
 
 #ifndef __has_attribute
 #define FOLLY_HAS_ATTRIBUTE(x) 0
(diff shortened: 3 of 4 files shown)


divitsinghall:
I have signed the CLA an hour ago

divitsinghall:
cc @yfeldblum — I’m taking a pass at fixing the C++20 ABI regression/ODR violation reported in #2477.

The issue stems from standard-dependent layout changes in structures using [[no_unique_address]]. I've implemented a detection layer in folly/CppAttributes.h and Portability.h that ensures header-binary consistency when consumers use a different C++ standard than the pre-compiled library. Full reproduction details and the behavioral matrix are included in the description

yfeldblum:
@divitsinghall Interesting!

Two key thing to note:
* We are migrating from C++17 to C++20. The default for folly was updated in: [Try bumping OSS builds to C++20](https://github.com/facebook/folly/commit/97f90a3c2488132dbba27e6d2287dc8c39612325). Some parts of folly still build with C++17 due to internal holdouts, but these are being addressed. Meanwhile, various parts of folly are already unconditionally using C++20 syntax and library facilities.
* In general, building and using a library with different sets of configuration defines or other settings will not work and can be assumed to lead to ODR violations. Although I suppose this more directly leads to ODR violation since `no_unique_address` controls layouts.

Issue #2538, opened 2025-12-03
`SimpleAllocator` segfaults when the size of the objects is less than 8


SaiCharanMarrivada:
I noticed that the [`SimpleAllocator`](https://github.com/facebook/folly/blob/main/folly/concurrency/CacheLocality.cpp#L580) segfaults if `sz` is less than 8. Is it implicit that the allocator is used for objects greater than or equal to 8 bytes?

yodakagekira:
Yes, that assumption is implicit

`SimpleAllocator` relies on each allocation being at least `sizeof(void*)` (8 bytes on 64-bit systems). Freed blocks store the freelist next pointer in their first 8 bytes, and each slab also begins with an 8-byte owner header. If `sz < 8`, these writes overrun or overlap internal metadata, leading to corruption and segfaults.

I think he public `Allocator` enforces this by mapping all `size <= 8` requests to the `sz_ = 8` size class, so `SimpleAllocator` is never used with smaller sizes unless constructed directly.

wangqin0:
validation added in #2558

SaiCharanMarrivada:
Thanks. Closing the issue.

PR #2558, opened 2026-01-14
Enhance SimpleAllocator with size validation

wangqin0:
## Summary

- Added size validation to `SimpleAllocator` constructor to ensure minimum allocation size of `sizeof(void*)` 
- Updated documentation in `CacheLocality.h` to clarify that allocations smaller than `sizeof(void*)` are rounded up
- Added test case `CoreAllocator.MinimumAllocationSize` to verify correct behavior for small allocations

## Test plan

- [x] Added new test `CoreAllocator.MinimumAllocationSize` that verifies 1-byte, 4-byte, and 8-byte allocations work correctly
- [ ] Run existing `CacheLocalityTest` suite to ensure no regressions

### DIFF
--- a/folly/concurrency/CacheLocality.cpp
+++ b/folly/concurrency/CacheLocality.cpp
@@ -582,7 +582,18 @@ class SimpleAllocator {
   // To support array aggregate initialization without an implicit constructor.
   struct Ctor {};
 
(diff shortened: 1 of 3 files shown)


wangqin0:
closes https://github.com/facebook/folly/issues/2538

PR #2562, opened 2026-01-20
Fix aarch64 duplicate symbol errors in libfolly

ptarjan:
The -use variants of memcpy_aarch64 and memset_aarch64 were being included in the monolithic libfolly.so, causing duplicate symbol errors because they compile the same assembly files as the non-use variants.

Add EXCLUDE_FROM_MONOLITH to memcpy_aarch64-use and memset_aarch64-use to match the pattern used for memcpy-use and memset-use in the main folly/CMakeLists.txt. The -use variants are meant for explicit opt-in to override libc memcpy/memset, not for inclusion in the main library.

### DIFF
--- a/folly/external/aor/CMakeLists.txt
+++ b/folly/external/aor/CMakeLists.txt
@@ -34,6 +34,7 @@ folly_add_library(
 
 folly_add_library(
   NAME memcpy_aarch64-use
+  EXCLUDE_FROM_MONOLITH
   SRCS
     memcpy-advsimd.S
     memcpy-armv8.S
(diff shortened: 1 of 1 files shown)


ptarjan:
Same result as https://github.com/facebook/folly/pull/2561 different implementation

cho-m:
I tried patching this into latest v2026.03.02.00 for Homebrew's arm64 Linux build and noticed that it doesn't work when building shared libraries:
```
  CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
  Missing variable is:
  CMAKE_ASM_CREATE_SHARED_LIBRARY
```

I worked around this by adding `set(CMAKE_ASM_CREATE_SHARED_LIBRARY ${CMAKE_C_CREATE_SHARED_LIBRARY})`

PR #2582, opened 2026-02-06
Fix: Add cxxabi fallback to buffer-based demangle function

### LINKED ISSUE #2546: Symbolizer does not demangle function names if binutils-dev is not installed before compiling folly
We have 2 versions of demangle() in Demangle.h:
```
fbstring demangle(const char* name);
size_t demangle(const char* name, char* out, size_t outSize);
```
The only difference is supposed to be that one allocates new string while the other uses your provided buffer. (Potentially reused or memory pooled and therefore more efficient.) But that's not what actually happens.

https://github.com/facebook/folly/blob/1b1832193fce19e9fff5283890efe13ac396a94e/folly/Demangle.cpp#L149
In the 1st version, we try libiberty, spelled as liberty in the code. And if not found, try cxxabi.

https://github.com/facebook/folly/blob/1b1832193fce19e9fff5283890efe13ac396a94e/folly/Demangle.cpp#L221
We don't do that in the 2nd version.

Note that the function actually does allow using provided buffer. We just choose to set it to nullptr to get new buffer. https://gcc.gnu.org/onlinedocs/libstdc++/latest-doxygen/a00026.html#aaf2180d3f67420d4e937e85b281b94a0

In Symbolizer, we call the 1st version. https://github.com/facebook/folly/blob/a6cd0997a7e7fa155dceae2c1c58906cf4294048/folly/debugging/symbolizer/SymbolizePrinter.cpp#L156

Other Facebook projects such as Velox rely on this code to print stack trace. So all test logs are unreadable unless you install binutils-dev and recompile folly first.

SHH123-crypto:
Fixes #2546

The buffer-based demangle() function now falls back to cxxabi when libiberty is not available, matching the behavior of the string-based version. This ensures stack traces are readable even without binutils-dev.

### DIFF
--- a/folly/Demangle.cpp
+++ b/folly/Demangle.cpp
@@ -256,6 +256,19 @@ size_t demangle(const char* name, char* out, size_t outSize) {
     }
   }
 
+  // fallback to cxxabi if available
+  if (cxxabi_demangle) {
+    int status;
+    size_t len = outSize;
+    // Use the provided buffer instead of allocating
+    char* demangled = cxxabi_demangle(name, out, &len, &status);
+    if (status == 0 && demangled == out) {
+      // Successfully demangled into the provided buffer
+      return strlen(out);
+    }
+    // If demangling failed or allocated (shouldn't happen), fall through
+  }
+
   // fallback - just return original
   return folly::strlcpy(out, name, outSize);
 }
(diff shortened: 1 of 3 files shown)


yfeldblum:
* Every line shows as changed in the `Files changed` view.
* The `demangle` overload which writes to the argument out-buffer is called in signal handlers. It must not allocate.

yfeldblum:
`__cxa_demangle` may and typically does allocate internally, even when it is provided with an output buffer. This would be forbidden when invoked within a signal handler.

The overload of `demangle` which returns `fbstring` already allocates for the `fbstring` and is therefore already forbidden in signal handlers. So that overload has a fallback to `__cxa_demangle`.

PR #2587, opened 2026-02-20
Static lib handling for glog

afrind:
Summary: Move BUILD_SHARED_LIBS=ON from the unconditional cmake.defines section to a shared_libs=on conditional section in the glog manifest, so that static builds don't force shared library generation.

Reviewed By: bigfootjon

Differential Revision: D93644456


bigfootjon (approved):
Review automatically exported from Phabricator review in Meta.

PR #2590, opened 2026-02-23
Make folly_detail_perf_scoped depend on folly_subprocess

mszabo-wikia:
On Linux platforms PerfScoped.cpp pulls in Subprocess.h, so linking in folly_detail_perf_scoped either on its own or via the catch-all target fails unless folly_subprocess is explicitly passed in as well. Make the dependency explicit instead.

### DIFF
--- a/folly/detail/CMakeLists.txt
+++ b/folly/detail/CMakeLists.txt
@@ -169,6 +169,11 @@ folly_add_library(
     folly_mpmc_queue
 )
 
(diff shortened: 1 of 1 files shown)


afrind:
This file is now generated based on our BUCK file.  Let me take a look

afrind:
I see what's wrong with our generator and will try to update it soon.

afrind:
See https://github.com/facebook/folly/pull/2595

PR #2631, opened 2026-04-10
OID truncation in getExtOid due to insufficient buffer size

metsw24-max:
getExtOid resizes the buffer to length when OBJ_obj2txt requires more space, but the returned length does not include the null terminator. 
This makes the buffer one byte too small and causes the last character of long OIDs to be truncated. Update the condition to length >= buf_size and resize to length + 1 to return the full OID correctly


metsw24-max:
any update?

PR #2634, opened 2026-04-16
Fix OOB read in .gnu_debuglink parsing due to unbounded strlen

metsw24-max:
The ELF debuglink parsing logic used strlen() on raw section data without verifying the presence of a null terminator within section bounds.

This results in an out-of-bounds read when processing malformed ELF files where the .gnu_debuglink section is not null-terminated.

-Replaces strlen() with bounded memchr()-based parsing
-Validates that a null terminator exists within section bounds
-Avoids copying beyond validated length and explicitly terminate output buffer

### DIFF
--- a/folly/debugging/symbolizer/Elf.cpp
+++ b/folly/debugging/symbolizer/Elf.cpp
@@ -51,6 +51,26 @@
 namespace folly {
 namespace symbolizer {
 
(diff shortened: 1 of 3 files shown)


metsw24-max:
gentle ping

PR #2661, opened 2026-06-16
Fix compatibility with fmt 12.2.0

xiaozhuai:
## Problem

fmtlib/fmt deprecated `<fmt/core.h>` in commit [0007426](https://github.com/fmtlib/fmt/commit/0007426c2d8d22df6e56d3b4d109415242e683e0). Starting from fmt 12.2.0, using `<fmt/core.h>` causes compilation errors because the deprecated include is now opt-in only.

## Fix

Replace all `#include <fmt/core.h>` with `#include <fmt/format.h>`, which is the standard header in fmt 12.2.0+.

### DIFF
--- a/CMake/FindFmt.cmake
+++ b/CMake/FindFmt.cmake
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-find_path(LIBFMT_INCLUDE_DIR fmt/core.h)
+find_path(LIBFMT_INCLUDE_DIR fmt/format.h)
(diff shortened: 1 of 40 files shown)


zhjwpku:
Since folly is now targeting c++20, why not just replace fmt with std::format?

xiaozhuai:
> Since folly is now targeting c++20, why not just replace fmt with std::format?

That's another matter.

mavam:
Needing this one urgently, now that Homebrew is on 12.2.

xiaozhuai:
Since there has been no response for a long time, this PR has been closed.

PR #2688, opened 2026-08-30
Use overflow-safe arithmetic in folly::Histogram

alwaysprince05:
## Summary

Replace raw unsigned addition/subtraction in `Histogram::addValue`, `addRepeatedValue`, `removeValue`, and `removeRepeatedValue` with the existing overflow-safe Bucket helpers (`addHelper`/`subtractHelper`) that use `constexpr_add_overflow_clamped` / `constexpr_sub_overflow_clamped`.

This eliminates undefined behavior from signed integer overflow when the bucket sum exceeds the representable range of `ValueType`, and replaces the previous wrapping behavior with deterministic clamping to numeric limits.

## Problem

The prior code performed `static_cast<ValueType>(to_unsigned(sum) + addend)`, which is implementation-defined for signed types when the result exceeds the representable range. `removeRepeatedValue` also computed `value * nSamples` before subtracting, which could overflow the intermediate product.

These issues had existing TODO comments from maintainers requesting this exact fix.

## Solution

Wire `Histogram` mutation methods into the safe arithmetic infrastructure already present in `detail::Bucket.h`:

- `addValue()` / `removeValue()` → delegate to `addToAccum()` / `subFromAccum()` which use `constexpr_add_overflow_clamped` / `constexpr_sub_overflow_clamped`
- `addRepeatedValue()` / `removeRepeatedValue()` → constant-time saturating multiply-add (see the follow-up commit), which avoids the overflow in the `value * nSamples` intermediate product and can no longer loop `nSamples` times

## Tests

- Updated 3 existing overflow tests to verify the new clamped behavior
- Added tests for clamping on overflow/underflow, repeated-value saturation (including a huge-`nSamples` case that must terminate immediately), unsigned histograms, and estimator edge cases

All `Histogram` and `TimeseriesHistogram` tests pass, along with the full `folly/stats` test suite.

[This description was updated after the follow-up commit, which replaced the temporary per-element loops with a constant-time saturating multiply-add.]

### DIFF
--- a/folly/stats/Histogram.h
+++ b/folly/stats/Histogram.h
@@ -239,22 +239,19 @@ class Histogram {
   /* Add a data point to the histogram */
   void addValue(ValueType value) {
     Bucket& bucket = buckets_.getByValue(value);
-    // NOTE: Overflow is handled elsewhere and tests check this
-    // behavior (see HistogramTest.cpp TestOverflow* tests).
-    // TODO: It would be nice to handle overflow here and redesign this class.
-    auto const addend = to_unsigned(value);
-    bucket.sum = static_cast<ValueType>(to_unsigned(bucket.sum) + addend);
+    addToAccum(bucket.sum, value);
     bucket.count += 1;
   }
 
   /* Add multiple same data points to the histogram */
   void addRepeatedValue(ValueType value, uint64_t nSamples) {
     Bucket& bucket = buckets_.getByValue(value);
-    // NOTE: Overflow is handled elsewhere and tests check this
-    // behavior (see HistogramTest.cpp TestOverflow* tests).
-    // TODO: It would be nice to handle overflow here and redesign this class.
-    auto const addend = to_unsigned(value) * nSamples;
-    bucket.sum = static_cast<ValueType>(to_unsigned(bucket.sum) + addend);
+    // Use a loop with overflow-safe addition to avoid overflow in the
+    // product (value * nSamples) that a single clamped multiply-and-add
+    // would require.
+    for (uint64_t i = 0; i < nSamples; ++i) {
+      addToAccum(bucket.sum, value);
+    }
     bucket.count += nSamples;
   }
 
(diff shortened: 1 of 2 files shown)


alwaysprince05:
Hey! I pushed a follow-up commit (`5cad237c7`) that tightens this up after some more testing locally.

The main thing: the per-element loop in `addRepeatedValue`/`removeRepeatedValue` effectively hangs if a caller passes a huge `nSamples` (count is `uint64_t`, so nothing stops that from happening), so I replaced it with a constant-time saturating multiply-add. It produces the same result the loop would, it just can't spin for ages, and it keeps the sum clamped at the type limits exactly like the single-value path does. I validated the arithmetic with a standalone fuzz harness (~7.5M randomized cases across int8→uint64, cross-checked against exact 128-bit ground truth, clean under UBSan) before touching folly.

While testing that I ran into a second spot where the new clamping surfaces an old overflow: the percentile estimator extrapolates with expressions like `high - (2 * (high - avg))`, and when a bucket average is now `INT64_MIN` or `INT64_MAX` that arithmetic overflows on its own. I moved those two extrapolations into unsigned arithmetic so the existing "adjust if it wrapped" logic still works instead of being UB.

Also expanded the tests a bit: cases for the huge-count behavior (must terminate immediately), unsigned histograms, and estimator edge cases, and tightened the overflow assertions so they can't pass vacuously (strict sign and range checks instead of the previous loose ones).

Everything passes locally on macOS (Apple clang): the `Histogram` and `TimeseriesHistogram` suites are green, and the full `folly/stats` ctest suite (79 tests, including `BucketedTimeSeries`, quantile estimator/histogram, and buffered stats, which all include these headers) passes.

Happy to rework any of this if you'd rather keep it simpler — the helper lives in `detail/Bucket.h` next to the existing add/subtract helpers.

alwaysprince05:
Hey! Rebased this onto latest main — that picked up d1a569ec1 (the constructor/percentile argument validation), which is where the merge conflicts were coming from. Nothing tricky in the overlap, both sides are kept.

While re-verifying after the rebase I found a real bug in my own change: `Histogram<double>` stopped compiling. The O(1) repeated-value helper was unconditionally casting `numeric_limits<ValueType>::max()` to `uint64_t`, which breaks for floating types in a constexpr context — and doubles used to work fine, so that's on me. Fixed by dispatching on `is_integral`: floating types now go through the existing add/subtract helpers with `value * nSamples`, same arithmetic as before. Added a `FloatingPointRepeatedValue` test so it doesn't come back quietly.

Re-ran the fuzz harness against the committed helpers (~7.5M cases, int8_t through uint64_t, cross-checked against __int128) and the Histogram tests on macOS — 25/25 green, including the new upstream validation tests.

